PR #37, opened 2021-10-28
Add basic CMake integration

wpumacay:
## Description
Hi all!, thanks for making MuJoCo open-source :sweat_smile: . This PR just adds a proof of concept of some basic integration with `CMake` (so far for linux). There are some targets exposed under the `mujoco::` alias, namely `mujoco::mujoco` (for `libmujoco210.so`, the one that has the rendering features enabled), `mujoco::glfw` and `mujoco::glew` (for the provided `glfw` and `glew` libraries that come with the pre-built binaries).

You can test that the build works by doing the following (it will build the `simulate` application from the samples :
```bash
cmake -B build
cmake --build build
```

To add to your CMake project, just add this as a subproject; all the setup (downloading the pre-built library, etc.) is done by `CMake` during the configuration phase.
```cmake
# Somewhere in your project
add_subdirectory(PATH_TO_MUJOCO_REPO)
```

Perhaps it might be useful for integrating it with your current workflows. If required I can clean the code a bit more, setup the appropriate cases (e.g. `no-GL`, `GLX`, `EGL`, `OSMESA`,  etc.) and setup some basic CI to make sure it works in all platforms (currently I've just tested it on linux).


nimrod-gileadi:
Thanks for this. At the moment we're focusing on setting up MuJoCo for open sourcing and our writing cmake files as well, but I think it's a good idea to have CMake support while we only have a binary release, as you did here.

We'll look into how to incorporate this change, with support for other platforms. Thanks for your patience.

nimrod-gileadi:
Hey, sorry for the long wait. Hopefully the latest CMake changes do what you want.

You can check out https://github.com/deepmind/mujoco/blob/main/sample/CMakeLists.txt

PR #141, opened 2022-02-14
Allow MjGeoms positions to be anywhere when converted to free objects

makeplayhappy:
If you create a MJGeom move it into a position (other than the origin) and then convert it to a free object, it breaks the simulation. The body (parent) and free joint are created at the origin, while the geom is at a differnet location, on simulation jittery unusual behavior and oddness ensues. In the fix below I move the newly created GameObjects to the geoms position before parenting happens.


saran-t:
Are you still interested in getting this change merged? If you do, you'll need to [sign the Google CLA](https://github.com/deepmind/mujoco/pull/141/checks?check_run_id=5184453774) before I can proceed. Also you'll also need to look into the merge conflict detected above.

erez-tom:
Actually, I believe I implemented something similar and fixed it.

PR #192, opened 2022-03-16
Make python build work with older cmake versions

aftersomemath:
The ability to create a interface library target with source files was introduced in cmake 3.19.

Ubuntu 20.04 ships with cmake 3.16.3. This PR changes the CMake scripts to use the older syntax, which allows the Python bindings to build on Ubuntu 20.04 systems.

https://cmake.org/cmake/help/latest/command/add_library.html#interface-libraries


saran-t:
Thanks! Have you managed to successfully built the Python bindings on your machine with this change?

aftersomemath:
Yes and I was able to run the minimal example from the docs and the render_test without errors.

Thanks for the quick response with the cmake files and mutable attribute fixes!

saran-t:
That's good to hear!

If you would like me to merge this PR you'll need to sign the CLA per the message from google-cla above.

aftersomemath:
Thanks for the reminder, I just signed the CLA.

saran-t (approved):

PR #210, opened 2022-03-27
Add `__version__` to Python module.

kevinzakka:
Add the ability to print the current version of MuJoCo.

```python
>>> import mujoco
>>> mujoco.__version__
'2.1.3'
```

### DIFF
--- a/python/mujoco/__init__.py
+++ b/python/mujoco/__init__.py
@@ -79,3 +79,14 @@ if _MUJOCO_GL not in _MUJOCO_GL_DISABLE:
       from mujoco.egl import GLContext
     else:
       from mujoco.glfw import GLContext
+
+
+def _get_version() -> str:
+  with open(os.path.join(HEADERS_DIR, 'mujoco.h'), 'r') as f:
+    for line in f:
+      if line.startswith('#define mjVERSION_HEADER'):
+        version = line.split()[2]
+        break
+  return '.'.join([d for d in str(version)])
+
+__version__ = _get_version()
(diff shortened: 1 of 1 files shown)


python/mujoco/__init__.py
```diff
@@ -79,3 +79,14 @@
...
+
+def _get_version() -> str:
+  with open(os.path.join(HEADERS_DIR, 'mujoco.h'), 'r') as f:
+    for line in f:
+      if line.startswith('#define mjVERSION_HEADER'):
+        version = line.split()[2]
+        break
+  return '.'.join([d for d in str(version)])
+
+__version__ = _get_version()
```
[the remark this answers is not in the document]
kevinzakka: Oops, I only discovered this last night while pouring over the docs :)

python/mujoco/__init__.py
```diff
@@ -79,3 +79,8 @@
       from mujoco.egl import GLContext
     else:
       from mujoco.glfw import GLContext
+
+
+from mujoco import mj_versionString
```
saran-t, line 84: You don't need this line: the `from mujoco._functions import *` line should've taken care of it already.

saran-t (changes requested):

python/mujoco/__init__.py
kevinzakka: Done.

python/mujoco/__init__.py
```diff
@@ -79,3 +79,6 @@
       from mujoco.egl import GLContext
     else:
       from mujoco.glfw import GLContext
+
+
```
saran-t, line 83: nit: remove one extraneous empty line

saran-t (changes requested):

saran-t (approved):

saran-t:
It might take me a little while to get this merged on our side. We can't use GitHub merge since our source of truth is Google's internal monorepo, and right now it looks like our Copybara workflow isn't correctly configured to accept PR (hence the error you see above).

PR #275, opened 2022-05-13
Fixing sensor reading in Unity, expose Advanced Settings property in Editor and add Priority tag to geometry

jespersmith:
This PR makes some small fixes to the Unity plugin to make it work with our controller

If desired, I could split this PR up but that would be a bit of work.

- Fixed offsets for the sensor readings. The sensor offset was multiplied by 3 twice. (MjSiteVectorSensor.cs)
- Added advanced settings to MjGeomEditor to be able to tune contact parameters (MjGeomEditor.cs)
- Added a raw configuration field to the hinge joint with double precision, to allow reading the angle of a continous joint without normalization
- Added the priority field to geometry to allow setting geometry priority for collisions


kevinzakka:
@jespersmith I'm super duper curious to see what working with Unity looks like before I try it myself. Was wondering if you have a video showing this off? (understandable if not possible).

saran-t (approved):
Approving the PR to trigger the migration.

PR #288, opened 2022-05-21
Add annotations for length of a given array in mj* structs

liuliu:
This PR added necessary annotations, following what we did for mjData
and mjModel for length of the given array.

This is useful because my parser used this info to generate Swift interface:
https://github.com/liuliu/swift-mujoco/blob/main/Sources/codegen/main.swift#L297
such that the access to underlying array is bound-checked.

Please let me know if this makes sense. After this change, I can
successfully generate all mj* struct interfaces in Swift without problem
(with notable exception of mjui): https://github.com/liuliu/swift-mujoco/tree/main/Sources

### DIFF
--- a/include/mujoco/mjrender.h
+++ b/include/mujoco/mjrender.h
@@ -108,7 +108,7 @@ struct mjrContext_ {              // custom OpenGL context
 
   // texture objects and info
   int ntexture;                   // number of allocated textures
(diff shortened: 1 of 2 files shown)


saran-t (approved):
Approving to trigger migration to internal repo

PR #302, opened 2022-05-27
Stricter parsing of XML array attributes

quagla:
- Added three new xml reading tests
- Using fail() instead of bad() to be more strict in accepting an element
- Improve clarity by using a while loop instead of a for
- Unified all ReadAttr in a single function using templates


quagla:
Solved by https://github.com/deepmind/mujoco/commit/f3453cf87f09773ea25d322df7c5a276d71820d7

PR #253, opened 2022-04-25
Unity Plugin: compiler directive to enable builds and synchronising C# scripts with MuJoCo

Balint-H:
# Description

Two small additions:

1.  The plugin's Editor scripts were not ignored during build, leading to unresolved references preventing standalone builds. The scripts that prevented the build were wrapped with the following directive:
```
#if UNITY_EDITOR
//Component Editor Script
#endif
```
Behaviour is not impacted in the editor, but builds are now successful. Fixes #245.

2.  An event was added to `MjScene` that allows C# scripts, such as custom controllers and agents to subscribe handlers that are invoked between `mj_step1` and `mj_step2`. This can improve the stability of certain actuator controllers, and can guarantee order of execution instead of relying on separate `FixedUpdate()`s. If no such handler is subscribed, then `mj_step` is executed as usual. Pointers to the simulation data for event handlers are included in a custom `EventArgs`. Related to #233.

Issue #245, opened 2022-04-15
Build issue with Unity Plugin with Editor scripts


Balint-H:
Hello, on Windows 10, Unity version 2021.2.5 running a build with MjComponents within fails due to unresolved references / type and namespace issues in existing MjComponent's Editor scripts. There are no issues due to these scripts inside the Editor, and wrapping all mujoco editor scripts (e.g. `MjGeomEditor.cs`) with the directive

```
#if UNITY_EDITOR
//Component Editor Script
#endif
```
resulted in no issues in compilation and build. Are there any specific steps that can be taken to build with the plugin without these directives? Or could this be an issue stemming from an incorrect import of the plugin on my side?

PR #303, opened 2022-05-27
Change asmdef to only compile scripts in unity/Editor in editor mode

jespersmith:
This PR is an alternative to https://github.com/deepmind/mujoco/pull/253, but in my opinion cleaner because it does not require #ifdef directives in every file.

Unity packages are based on .asmdef assembly definitions, which describe which targets scripts in the folder structure underneath the asmdef require. Hence, by changing the asmdef to target only the editor all editor scripts get ignored during build step.

I moved MujocoBinaryRetriever to the Editor assembly, as it is used only in Editor mode.


Balint-H:
I was not aware of this solution to #245, and I actually prefer it compared to how I handled it in #253.

saran-t (approved):

PR #304, opened 2022-05-27
Remove useless TINYOBJLOADER_IMPLEMENTATION definition when building MuJoCo with CMake

traversaro:
As far as I understand, `TINYOBJLOADER_IMPLEMENTATION` macro is used to be defined in libraries or executables that consume `tinyobjloader` only including its header, without linking the corresponding library. However mujoco (at least with the CMake build system provided in this repo) is already linking the tinyobjloader library in which all the symbols are defined, so there is no need to re-defined them.

### DIFF
--- a/src/user/CMakeLists.txt
+++ b/src/user/CMakeLists.txt
@@ -25,3 +25,4 @@ set(MUJOCO_USER_SRCS
 )
 
 target_sources(mujoco PRIVATE ${MUJOCO_USER_SRCS})
+target_compile_definitions(mujoco PRIVATE MUJOCO_DO_NOT_DEFINE_TINYOBJLOADER_IMPLEMENTATION)
--- a/src/user/user_mesh.cc
+++ b/src/user/user_mesh.cc
@@ -25,7 +25,11 @@
 #include <string>
 #include <vector>
 
+// This macro is defined when MuJoCo is built with CMake,
+// not defined when MuJoCo is built with other build systems
+#ifndef MUJOCO_DO_NOT_DEFINE_TINYOBJLOADER_IMPLEMENTATION
 #define TINYOBJLOADER_IMPLEMENTATION
+#endif
 
 #include <mujoco/mjmodel.h>
 #include "cc/array_safety.h"


saran-t (approved):

saran-t:
OK so this is required for our internal build process that we used to generate our official binaries.

The prebuilt library that we ship contains all of MuJoCo's dependencies statically linked in, but with symbols hidden. We still believe that this is the cleanest and safest way for us to ship prebuilt libraries (we also target `GLIBC_2.17` aka `manylinux2014` aka CentOS 7, and no external C++ ABI surface). This means that in our build we _want_ that file to provide `tinyobjloader` symbols.

The correct "fix" for this would be to define a CMake variable (`MUJOCO_STATIC_TINYOBJLOADER` or something like that) that translates to a macro that controls whether `TINYOBJLOADER_IMPLEMENTATION` is defined.

traversaro:
> The correct "fix" for this would be to define a CMake variable (`MUJOCO_STATIC_TINYOBJLOADER` or something like that) that translates to a macro that controls whether `TINYOBJLOADER_IMPLEMENTATION` is defined.

Sorry, I had missed that. I will look into this, thanks!

traversaro:
@saran-t I tried to implement in the simplest way possible your suggestions. 

What happens with this PR is the following:
* If MuJoCo is built with CMake, `TINYOBJLOADER_IMPLEMENTATION` is not defined
* When MuJoCo is built with some other build system not mantained in this repo such as your internal build process, `TINYOBJLOADER_IMPLEMENTATION` 

Is this ok for you? I kind of made the assumption that your internal build process is not based at all on CMake (I guess is blaze/bazel-based), but pleas tell me if this assumption is wrong.

traversaro:
> @saran-t I tried to implement in the simplest way possible your suggestions.
> 
> What happens with this PR is the following:
> 
>     * If MuJoCo is built with CMake, `TINYOBJLOADER_IMPLEMENTATION` is not defined
> 
>     * When MuJoCo is built with some other build system not mantained in this repo such as your internal build process, `TINYOBJLOADER_IMPLEMENTATION`
> 
> 
> Is this ok for you? I kind of made the assumption that your internal build process is not based at all on CMake (I guess is blaze/bazel-based), but please tell me if this assumption is wrong.

Hello @saran-t, I was doing a cleanup of PRs that I had open, are you still interested in this change? If yes I can keep the PR open, otherwise perhaps I would close it to cleanup my PRs list. Thanks a lot in advance!

saran-t (approved):

traversaro:
Thanks @saran-t !

PR #308, opened 2022-05-29
Annotate int type with actual mjt* type in mujoco.h

liuliu:
This PR annoated int type with the actual mjt* type in comment as
"parameter_name is mjtTYPE". This helps external parsers and document
generators to understand what the type MuJoCo refers to as in C, all
these are int.

The usefulness can be seen in swift-mujoco:

Before this PR:

https://github.com/liuliu/swift-mujoco/blob/0df02e2bc1eb07d74cb97f03fe1e24b7cd944428/Sources/MjrContext%2BFunctions.swift#L50

After this PR, this interface is generated:

https://github.com/liuliu/swift-mujoco/blob/main/Sources/MjrContext%2BFunctions.swift#L54

Next:

I need to figure out how to tag nullability for parameters. Would:
"parameter_name is nullable" acceptable one in comment?

### DIFF
--- a/include/mujoco/mujoco.h
+++ b/include/mujoco/mujoco.h
@@ -556,11 +556,13 @@ MJAPI void mjv_defaultOption(mjvOption* opt);
 MJAPI void mjv_defaultFigure(mjvFigure* fig);
 
 // Initialize given geom fields when not NULL, set the rest to their default values.
+// type is mjtGeom.
 MJAPI void mjv_initGeom(mjvGeom* geom, int type, const mjtNum size[3],
                         const mjtNum pos[3], const mjtNum mat[9], const float rgba[4]);
 
 // Set (type, size, pos, mat) for connector-type geom between given points.
 // Assume that mjv_initGeom was already called to set all other properties.
+// type is mjtGeom.
 MJAPI void mjv_makeConnector(mjvGeom* geom, int type, mjtNum width,
                              mjtNum a0, mjtNum a1, mjtNum a2,
                              mjtNum b0, mjtNum b1, mjtNum b2);
@@ -574,11 +576,11 @@ MJAPI void mjv_makeScene(const mjModel* m, mjvScene* scn, int maxgeom);
 // Free abstract scene.
 MJAPI void mjv_freeScene(mjvScene* scn);
 
-// Update entire scene given model state.
+// Update entire scene given model state. catmask is mjtCatBit.
 MJAPI void mjv_updateScene(const mjModel* m, mjData* d, const mjvOption* opt,
                            const mjvPerturb* pert, mjvCamera* cam, int catmask, mjvScene* scn);
 
-// Add geoms from selected categories.
+// Add geoms from selected categories. catmask is mjtCatBit.
 MJAPI void mjv_addGeoms(const mjModel* m, mjData* d, const mjvOption* opt,
                         const mjvPerturb* pert, int catmask, mjvScene* scn);
 
@@ -600,7 +602,7 @@ MJAPI void mjr_defaultContext(mjrContext* con);
 // Allocate resources in custom OpenGL context; fontscale is mjtFontScale.
 MJAPI void mjr_makeContext(const mjModel* m, mjrContext* con, int fontscale);
 
-// Change font of existing context.
+// Change font of existing context; fontscale is mjtFontScale.
 MJAPI void mjr_changeFont(int fontscale, mjrContext* con);
 
 // Add Aux buffer with given index to context; free previous Aux buffer.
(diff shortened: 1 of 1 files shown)


include/mujoco/mujoco.h
```diff
@@ -556,11 +556,13 @@ MJAPI void mjv_defaultOption(mjvOption* opt);
 MJAPI void mjv_defaultFigure(mjvFigure* fig);
 
 // Initialize given geom fields when not NULL, set the rest to their default values.
+// type is mjtGeom.
 MJAPI void mjv_initGeom(mjvGeom* geom, int type, const mjtNum size[3],
                         const mjtNum pos[3], const mjtNum mat[9], const float rgba[4]);
 
 // Set (type, size, pos, mat) for connector-type geom between given points.
 // Assume that mjv_initGeom was already called to set all other properties.
+// type is mjtGeom.
```
yuvaltassa, line 565: MuJoCo's line length is 100 chars, I don't think a new line is required?

yuvaltassa:
I feel like there are two very separate aspects here:

1. Documentation for human readers. I feel like this would be much better served simply by better documentation. Indeed we intend to regenerate and improve the [API reference](https://mujoco.readthedocs.io/en/latest/APIreference.html) page, and will easily be able to improve this aspect.
2. Annotation for automated parsers. This purpose makes more sense to me, but in this case would we not be better served by replacing all the ints in the API which are actually enums by the enum type? 

The latter is of course a bigger question. There is a reason the enums are ints in MuJoCo, and that is that the C ABI does not specify the size that the compiler should assign to enum types. So perhaps there is a compiler out there which would create incompatible data-structures using the same source code if we were to use the enums types. We actually have [a runtime test](https://github.com/deepmind/mujoco/blob/main/test/header_test.cc) that checks that enum types are in fact ints and would love to know if this test fails for someone using a compiler we have not tried.

Let us for the moment assume that indeed there are no compilers that we care about that would assign an enum type a size that is not sizeof(int). What would you think of a change that replaced all the int variables which are actually enums with the appropriate type?

Note that while in some senses this change would be trivial, it would also be a huge breaking change in terms of the API itself.

liuliu:
Agreed on the assessment. TBH, changing the int type to the appropriate enum type would be ideal. If we plan to do API breaking change, I would suggest also adopt "_Nullable / _Nonnull" annotation: https://clang.llvm.org/docs/AttributeReference.html#nullable which helps both the compiler and automated interface generator.

liuliu:
Also, this means we need to change types in all public structs, which currently relied on annotation for enum types, such as: https://github.com/deepmind/mujoco/blob/main/include/mujoco/mjmodel.h#L701

However, this is not exactly consistent (or the rules is more refined), for example, in mjData, this is directly the enum type: https://github.com/deepmind/mujoco/blob/main/include/mujoco/mjdata.h#L137

yuvaltassa:
It is consistent, `mjWarningStat` [is a struct](https://github.com/deepmind/mujoco/blob/main/include/mujoco/mjdata.h#L95).

You are correct that this would be a huge change in terms of the API, which is what we are hesitant. Perhaps we will try it internally and if it seems doable, we'll bring it up for discussion here.

One of the reason I am hesitant to approve this PR is precisely because I believe it is rather partial. Are there not many other functions that take an `int` which is actually an `enum`?
`skipstage` [here](https://github.com/deepmind/mujoco/blob/main/include/mujoco/mujoco.h#L142)
`type` [here](https://github.com/deepmind/mujoco/blob/main/include/mujoco/mujoco.h#L400) and [here](https://github.com/deepmind/mujoco/blob/main/include/mujoco/mujoco.h#L403)
`objtype` [here](https://github.com/deepmind/mujoco/blob/main/include/mujoco/mujoco.h#L423) and [here](https://github.com/deepmind/mujoco/blob/main/include/mujoco/mujoco.h#L427)

Do we really want this information in code comments? 

I propose that, if it's okay with you, have some custom code for the functions you really care about typing properly, or just make do with `int`s. In the meantime we'll look more carefully into making this API change (I reckon this would imply a major version number increment, btw). If we decide against doing it, then we'll revisit this PR.

Does that sound reasonable?

liuliu:
> It is consistent, `mjWarningStat` [is a struct](https://github.com/deepmind/mujoco/blob/main/include/mujoco/mjdata.h#L95).

Oh, right! Sorry I misread mjtWarning and mjWarningStat.

> Does that sound reasonable?

Yeah, definitely. Wouldn't block my end. One thing I would bring up for the discussion besides the enum is the nullability annotation for pointers. I probably would make do with a few selected APIs on my end. But I do want to do a full code scan (now with the source code) to make sure that I understand for which parameters null are OK and which are not. Would like to know if this is aligned with your goal as which parameter is nullable seems both under-documented and annotated.

yuvaltassa:
`_Nullable` and `_Nonnull` are actually super useful annotations that I was not aware of until now. 
Thanks so much for the education!

This seems like a very worthwhile thing to have, let us discuss and test internally and get back to you. Feel free to open a tracking issue for this question specifically so it's not relegated to a PR discussion :)

PR #323, opened 2022-06-03
Update XmlElementExtensions.cs

TachikakaMin:
I have some xml files that can be loaded by Mujoco 2.1.1 and only use one value to represent Vec3. So I think we can also add this support to unity plugin.


erez-tom (approved):

erez-tom (approved):

SebasNorenaR:
Hi @TachikakaMin, The proposed behavior (broadcasting a single scalar to a 3D vector and ignoring missing default classes) diverges from MuJoCo's strict MJCF schema rules, we are closing this pull request.

PR #329, opened 2022-06-07
Fix two asan detected memory bugs

liuliu:
I am running through [simulate.swift](https://github.com/liuliu/swift-mujoco/blob/main/Examples/simulate/main.swift) (my port of simulate.cc) with
address sanitizer to discover related bugs.

Besides ones in my port, there are two in MuJoCo:

1. in maketext, the logic to find . is not protected against j is less
   than 0 (due to the decreasing logic above), creating out of bound
   access.

2. in mj_printFormattedData, qfrc_applied should use length nv not nq,
   otherwise out of bound access could be triggered.

Test Plan:

Run through the simulate.cc with asan. Before this fix, when presenting
profiler view, it will trigger bug #1. When print data, it will trigger
bug #2. Both are using model/humanoid/22_humanoids.xml.


yuvaltassa:
Thanks!

yuvaltassa (approved):

saran-t (approved):

PR #356, opened 2022-06-23
Add ResetScene function to MjScene

TachikakaMin:
Add ResetScene function to Scene which will make it more convenient to interact with mlagent.


erez-tom (approved):
Approved for internal review.

PR #376, opened 2022-07-09
Simple toggleable dark mode for docs

rohit-kumar-j:
**Preview:**

dark_mode_mujoco

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -27,6 +27,7 @@ sys.path.append(os.path.abspath('ext'))
 
 import sphinxcontrib.katex as katex  # pylint: disable=g-import-not-at-top
 import sphinxcontrib.youtube as youtube  # pylint: disable=g-import-not-at-top
+import sphinx_rtd_dark_mode
 
 # -- Project information -----------------------------------------------------
 
@@ -45,6 +46,7 @@ extensions = [
     'sphinxcontrib.katex',
     'sphinxcontrib.youtube',
     'sphinx_reredirects',
+    "sphinx_rtd_dark_mode",
 ]
 
 # Add any paths that contain templates here, relative to this directory.
@@ -80,6 +82,7 @@ autodoc_default_options = {
 # a list of builtin themes.
 #
 html_theme = 'sphinx_rtd_theme'
+sphinx_rtd_dark_mode.default_dark_mode = False
 
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -3,6 +3,7 @@ sphinx-rtd-theme==1.0.0
 sphinxcontrib-katex==0.8.6
 sphinxcontrib-youtube==1.1.0
 sphinx-reredirects==0.0.1
+sphinx-rtd-dark-mode==1.2.4
 nbsphinx==0.8.0
 pandoc==1.0.2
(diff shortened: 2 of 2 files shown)


yuvaltassa:
This is pretty cool!

Unfortunately we have a breaking commit just ahead of you, let us fix that first... 

In the meantime, any thoughts on how font colours can be set specifically for dark mode?

yuvaltassa:
Our fix was submitted, please re-sync so tests pass, thanks.

rohit-kumar-j:
done

yuvaltassa (approved):

rohit-kumar-j:
Apologize for the late response, JS, CSS, and HTML were new to me and I was figuring out how to implement the dark mode without the extension since the sphinx docs for multiple `html_css_files` options were not easy to figure out.

@yuvaltassa, the locked theme problem can now be solved by changing the `href` in the layout.html instead of the conf.py

```html
<!-- Dark Mode-->
<link id="pagestyle" rel="stylesheet" type="text/css" href="_static/theme_overrides_dark.css">

<!-- Light Mode-->
<link id="pagestyle" rel="stylesheet" type="text/css" href="_static/theme_overrides.css">
```

I instead made a [`theme_overrides_dark.css`](https://github.com/rohit-kumar-j/mujoco/blob/main/doc/css/theme_overrides_dark.css) in addition to the [`theme_overrides.css`](https://github.com/deepmind/mujoco/blob/main/doc/css/theme_overrides.css) and modified [`layout.html`](https://github.com/rohit-kumar-j/mujoco/blob/main/doc/templates/layout.html) to respect the colors. Although I did not edit any of the colors. 

The `html_css_files` option is removed from the conf.py and the theme switching is handled by the JS added on each page by sphinx in the layout.html. It even remembers the theme when we switch websites and reload.

Here is a preview of messing around with the switching:

https://user-images.githubusercontent.com/37873142/179370990-dac422c4-f86b-449a-9ec6-6d632518ff66.mp4

yuvaltassa:
Hi @rohit-kumar-j !

Would you mind taking a look at @kevinzakka's #421 ?

The big win in #421's proposed use of `furo` is the right-hand sub-TOC which is incredibly useful. You have to try it to see what I mean.

Unfortunately it currently has a few glaring issues. Perhaps we can do some sort of best-of-both-worlds merge of the two proposals? You seem to be pretty good at this, it would be awesome if you took a stab.

Here are the issues in #421 as I see them:
1. General:
   1. The link symbol should not be '#' but the one we have now
   2. I'd prefer only light-dark mode toggle like in your PR, no "auto" mode. The asking-the-browser thing should only happen when you first load the page. This is more of a niggle than a deal breaker.
   3. The bold text seems "less bold", to the point where it's barely different from regular. Compare the current rendering of attribute headers in XML Reference (in red) with the rendering in 421.
2. Light mode:
   1. The table shadows are generally cute but don't work for the table-within-a-table in the schema at the top of XML Reference. If I had to choose I'd just go without shadows.
3. Dark mode
   1. The dark blue and red are too dark (of course making the dark blue lighter means making the regular blue lighter too so they wont be identical).
   2. The table borders become invisible, again look at the top of XML Reference.
  
I think that just fixing the colours (which you seem to be quite proficient at) and table issues would make a new-and-improved furo-based theme a slam dunk for documentation readability.

yuvaltassa:
Fixed by 571b703f1ccce0e768554adac82a1da95065bcad

PR #385, opened 2022-07-16
Fixes #382 (compile with unittests disabled)

### LINKED ISSUE #382: CMake configuration breaks if building with MUJOCO_BUILD_TESTS=OFF
Hi there.

I'm running into this issue when running the `CMake` configuration step, while using `-DMUJOCO_BUILD_TESTS=OFF`. It seems that tests are still requested for a python util via `MUJOCO_TEST_PYTHON_UTIL`, which enables testing, but it doesn't include `googletest`, therefore throwing an error during configuration, as shown in the screenshot below:

```bash
# CMake configuration with tests disabled
cmake -B build -DMUJOCO_BUILD_TESTS=OFF
```

img_mujoco_issue_build_with_no_tests

Notice that [here][0] `GoogleTest` is being included, but it's not included [here][1] for that python util (yet, `gtest_add_test` is being called without having `GoogleTest` included)

A fix would be to check for `MUJOCO_BUILD_TESTS` also on the following line on the root [`CMakeLists.txt`][2]:

img_mujoco_issue_build_with_no_tests_snippet

This works, but I'm not 100% sure what is the expected behaviour for this case. If that's the expected behaviour, I could write a small PR with this fix.

- **OS:** `Arch Linux`
- **MuJoCo version:** `2.2.0`

[0]: <https://github.com/deepmind/mujoco/blob/af12322e9d6fe1b5a7b59fe6eb4dc3f790f35580/test/CMakeLists.txt#L18> (tests-cmakefile)
[1]: <https://github.com/deepmind/mujoco/blob/af12322e9d6fe1b5a7b59fe6eb4dc3f790f35580/python/mujoco/util/CMakeLists.txt#L61> (tests-python-util-cmakefile)
[2]: <https://github.com/deepmind/mujoco/blob/af12322e9d6fe1b5a7b59fe6eb4dc3f790f35580/CMakeLists.txt#L164> (root-cmakefile)

wpumacay:
Small change to the `CMake` rules to fix #382


saran-t (approved):

wpumacay:
Oh, it's failing :grimacing: . Will check what went wrong :thinking:

wpumacay:
Sorry, made a typo `utils` instead of `util`; will update the PR

wpumacay:
Ok, sorry @saran-t , I've updated the PR. Tested locally this time and it works.

saran-t (approved):

PR #405, opened 2022-07-29
Updated Python doc to explain how __copy__ replaces mj_copyData

rpapallas:
Related issue: https://github.com/deepmind/mujoco/issues/404

### DIFF
--- a/doc/python.rst
+++ b/doc/python.rst
@@ -94,6 +94,9 @@ Functions
   arguments other than array sizes in the same order as they appear in :ref:`mujoco.h<inHeader>`, or use keyword
   arguments. For example, :ref:`mj_jac` should be called as ``mujoco.mj_jac(m, d, jacp, jacr, point, body)`` in Python.
 
+  Certain functions in the C API have been implemented differently in Python.  For example, the `mj_copyData` function 
+  is implemented as the `__copy__` dunder method. Therefore, `mj_copyData` is replaced with `copy.copy(data)` in Python.
+
   The bindings **releases the Python Global Interpreter Lock (GIL)** before calling the underlying MuJoCo function.
   This allows for some thread-based parallelism, however users should bear in mind that the GIL is only released for the
   duration of the MuJoCo C function itself, and not during the execution of any other Python code.


doc/python.rst
```diff
@@ -94,6 +94,9 @@ Functions
   arguments other than array sizes in the same order as they appear in :ref:`mujoco.h<inHeader>`, or use keyword
   arguments. For example, :ref:`mj_jac` should be called as ``mujoco.mj_jac(m, d, jacp, jacr, point, body)`` in Python.
 
+  Certain functions in the C API have been implemented differently in Python.  For example, the `mj_copyData` function 
+  is implemented as the `__copy__` dunder method. Therefore, `mj_copyData` is replaced with `copy.copy(data)` in Python.
```
saran-t, line 98: What do you think of something like the following? (I wrote the code, so will defer to you as an external reader to decide whether the explanation makes sense to you or not.)

Functions that operate on pointers to uninitialized objects or free resources are implemented in safer, more Pythonic manner rather than as direct analogs to the C API. This applies to initializers (e.g. `mj_makeData`, `mjv_makeScene`), which are implemented as constructors (e.g. `mujoco.MjData(m)`, `mujoco.MjvScene(m, maxgeom)`), and copiers (e.g. `mj_copyData`), which are implemented as `__copy__` (so that one can call `copy.copy(d)`). 

Feel free to reword to make it more succinct!

rpapallas: Hi @saran-t that sounds good and makes more sense, thank you. However, I realised that `mju_copy` is implemented in Python, probably for a reason, and the above statement might hint that the `mju_copy` is also replaced. The `mju_copy` isn't mainly copying numpy arrays? If so, numpy arrays also support __copy__, so it might be that the function is redundant as well?

SebasNorenaR:
Since this PR was opened, `mujoco.mj_copyData(dest, model, src)` has been directly exposed in the Python bindings (for copying into an existing `MjData` instance in-place, complementing `copy.copy(data) ` which allocates a new instance). Because `mj_copyData` is no longer replaced by `copy.copy`, we are closing this PR as obsolete.

PR #425, opened 2022-08-04
Add `mjRND_CULL_FACE` option to control `GL_CULL_FACE` and disable backward face culling

aftersomemath:
Disabling culling is useful when using meshes where only forward faces are available (as can happen with certain types of 3D scans). While disabling backface culling can cause aliased shadows it is not a problem when shadows are off, or when doing segmentation and depth rendering.

Also implicitly added a way to define "no shortcut" for a UI option since it seemed all the available keys were used up (and shift+something keys don't seem to work).


yuvaltassa (approved):

yuvaltassa:
Please sync to fix the builds.
Cheers

saran-t (approved):

PR #428, opened 2022-08-05
Fix off by one in segmentation with IDCOLOR documentation

aftersomemath:
Figured this out while using the feature however it wasn't hard to find the `+1` [in the rendering code](https://github.com/deepmind/mujoco/blob/7a0e0ed80cf65aa4f5142b1ee197c1f0c0a44955/src/render/render_gl3.c#L265)
 as well.

The rendered color needs to be `segid+1` so that the value "0" can represent "nothing".

### DIFF
--- a/doc/APIreference.rst
+++ b/doc/APIreference.rst
@@ -873,7 +873,7 @@ mjtRndFlag
        mjRND_FOG,                      // fog
        mjRND_HAZE,                     // haze
        mjRND_SEGMENT,                  // segmentation with random color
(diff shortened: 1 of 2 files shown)


yuvaltassa:
Could you also update `APIreference.rst` please?

Hopefully we'll autogenerate it soon but until then...

aftersomemath:
I think I did, is there another `APIreference.rst`?

yuvaltassa:
Sorry ignore me 🤪

yuvaltassa (approved):

Issue #486, opened 2022-09-16
Unity plugin, Weld offset field leads to instability


Balint-H:
Hello,

I'm using Unity 2021.2.19, Windows 10, MuJoCo 2.2.2, and latest version of the plugin. I've been using "puppeteering" from reference animations using mocap bodies and weld constraints. It used to be (back with 2.2.0) that if no offset was provided to the `MjWeld` component, it would use the transformation between the two constrained bodies at startup as the offset. If this was not the desired behaviour, we could supply a dummy, identity transform in the serialized field, so no offset was applied.

 Now if an offset is supplied, even if it is a transform with no `localPosition` or `localRotation` effects, the constraint leads to erratic movements from the constrained body, and very quickly an exception due to instability.
 
 8ca5887c20d47f7534c9c22b3e363b64e4e8e36b added additional data stored in `model->eq_data`, so is this caused by `MjTransformAtEntry` in the C# `MjWeld` component returning addresses not appropriate for `eq_data`? I tried simply increasing the address by 3 in `MjWeld`'s `OnBindToRuntime` , which resolved instability, but added an undesired offset in the vertical axis.

Balint-H:
Wanted to add, that if no Weld Offset transform is supplied then no weird behaviour happens perhaps because then edits to `eq_data` are skipped. However this disables the option to have no offset between constrained objects, but start at different states, which can be useful particularly for puppeteering (where the mujoco body definition might be more convenient in a non T-pose for e.g. joint limits, but have the reference animation skeleton in a T-pose).

erez-tom:
Your analysis is correct.  I believe you can fix things by replacing  `MjEngineTool.MjTransformAtEntry(model->eq_data, MujocoId)` with `model->eq_data + MujocoId * 11 + 3` (11 being the new value of `mjNEQDATA` following #8ca5887).  Let me know if that works out for you.

kevinzakka:
Fixed by #491.

PR #491, opened 2022-09-17
Fix MjWeld using wrong address in Unity plugin

Balint-H:
Fix according to advice in #486. In case someone wants to test it, it should be noted that this version of the plugin does not work with the prebuilt 2.2.2 dll, I think due to recent changes to the bindings in 1e2a9a53bcca2e2bf6758b350db32346d132098d (is a new release coming?). I tested the changes with a 2.2.2 friendly version at [this fork](/Balint-H/mujoco).

### DIFF
--- a/unity/Runtime/Tools/MjEngineTool.cs
+++ b/unity/Runtime/Tools/MjEngineTool.cs
@@ -106,7 +106,7 @@ public static class MjEngineTool {
   public static unsafe double* MjMatrixAtEntry(double* mjTarget, int offsetEntry) {
     return mjTarget + offsetEntry * _mjMat.Length;
   }
-  
+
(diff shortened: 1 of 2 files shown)


saran-t (approved):

unity/Runtime/Tools/MjEngineTool.cs
```diff
@@ -105,6 +106,11 @@ public static class MjEngineTool {
   public static unsafe double* MjMatrixAtEntry(double* mjTarget, int offsetEntry) {
     return mjTarget + offsetEntry * _mjMat.Length;
   }
+
```
saran-t, line 109: Our internal review tool is showing a stray space on this line. Could you please remove it?

Balint-H: Removed it, thanks for spotting it!

saran-t (approved):

PR #560, opened 2022-11-03
elastoplastic extension

mathl79:
implementation of the plasticity extension idea to the elasicity plugin. 
see:  discussion #549

- new xml config parameter yield stress ("yield"  [Pa]). Yield separates the elastic from the plastic domain
- generation of userdata output: von mieses equivalent stress, curvature, strain(change of curvature)
- skin/composit stress coloring
- skin per vertex color added to the datamodel
- CMAKE config of elasticity.dll changed to destination output folder \plugin\ 
- new test model xml file

googletest is not implemented yet --> need help to find sufficient test criterias


mathl79:
Hi! 
I never started a pull request before. 
So in order to fulfill the contribution requirements and adopt best practices, i am hoping for some beginner friendly remarks and some guidance.

quagla:
Hi @mathl79 I think you have changed something in the default behavior of the cable? It looks like the test for the exact solution of the cantilever beam fails.

Great that you are visualizing the stresses, this is something I always wanted to do!

mathl79:
thank you for taking a look.
The string loaded elasticity test model does not have the required  "yield" Parameter as it is a elastic  test only.
I need to set a infinite yield value in case the "yield" Parameter in the XML is missing at the cable constructor.
In that way, the test model should be loaded and run as expected.

quagla:
Do you mean that you updated `CantileverIntoCircle` in `elasticity_test`? I don't see that file in your commit.

mathl79:
The default elastic behavior should no be changed, if the new parameter "yield" is infinite. I missed to implement the default setting and remove the parameter check in this pull request, so the elasticity_test fails.

In the meantime i reworked my version to have a material model with strain hardening. 
image

Should i close this Pull draft and start a new branch in my fork with the advanced model ?

quagla:
Great! I think it's fine if you keep pushing new commits to this branch, the PR should be updated automatically.

quagla:
Thanks for the revised version! My main comments are:

- Why did you add so much damping to the original cable examples?
- Did you try the original examples and particularly to run `elasticity_test`? The default (purely elastic) behaviour has changed and it seems incorrect.
- More generally, can we break the PR into a visualization and a plasticity PR or would that be impossible?

mathl79:
I tried to implement a plasticity deformation proportional damping by reducing the torque depending on the change of omega an the Volume. 
 Now i see, that this additional damping is overlapping with the elastic joint damping. This needs to be separated.

I did not run the tests - but better should have...

So I will change back to the standard plugin models, add extra plastic test models and separate the viscous damping by only applying it, if elements are in plastic domain.

I try to separate the coloring feature from the plasticity as you suggested.
For the coloring, i assume it should be configurable by XML options, right ?
-definition of range (min,  max,  auto)
-constant RGB color below lower limit 
-constant RGB color above upper limit 
-color palette
-enable/disable

What else needs to be user controllable ?

Should any scalar properties other than equivalent stress be selectable in future ? 

Besides the separation from the plasticity, zhe stress needs to be calculated for the elasticity model anyway, in order to be able to show it.

Btw - what is a good place in the source for the color conversion and palette generation tools, that I have put inside the plugin code from now?

Is this something, that could be used generally or should it stay plugin specific?

mathl79:
I tried to implement a plasticity deformation proportional damping by reducing the torque depending on the change of omega an the Volume. 
 Now i see, that this additional damping is overlapping with the elastic joint damping. This needs to be separated.

I did not run the tests - but better should have...

So I will change back to the standard plugin models, add extra plastic test models and separate the viscous damping by only applying it, if elements are in plastic domain.

I try to separate the coloring feature from the plasticity as you suggested.
For the coloring, i assume it should be configurable by XML options, right ?
-definition of range (min,  max,  auto)
-constant RGB color below lower limit 
-constant RGB color above upper limit 
-color palette
-enable/disable

What else needs to be user controllable ?

Should any scalar properties other than equivalent stress be selectable in future ? 

Besides the separation from the plasticity, zhe stress needs to be calculated for the elasticity model anyway, in order to be able to show it.

Btw - what is a good place in the source for the color conversion and palette generation tools, that I have put inside the plugin code from now?

Is this something, that could be used generally or should it stay plugin specific?

mathl79:
Hi,  
i did run the tests on my windows machine and all configured tests do pass. 
But the `PluginTest.CantileverIntoCircle` is not even executed. Also the github build check showes, that the windows-2022 build does not run the elasticity tests, but fails at installing the file elasticity.dll into the plugin folder.

I don't know how to correct the config tests.

When building /debugging elasticity.dll, the file is placed into the /bin folder and needs to be copied into the /bin/plugin/ subfolder manually to be recognized as plugin. 

I think, that is also linked to the problem above.

quagla:
OK that's interesting, there is something broken with Windows then because on the main branch the dll gets copied. @saran-t what are your thoughts?

As mentioned in my previous comment, could you please open a new PR for the visualisation part? You can keep this one open as well and we will get back to it once we complete the visualisation. How does that sound to you? Maybe in the meantime we can fix the Windows issues as well.

mathl79:
https://github.com/deepmind/mujoco/blob/7b7fc09d1fc3d9a970b2123bc41c4d1fbcff35d6/test/CMakeLists.txt#L69-L71

This code excludes the plugin-test from the test config for Windows builds. what is the reason for this?

quagla:
Good point, we definitely need to fix Windows. Now I remember we disabled the test because we had some issues loading the DLL. I wonder if your change to CMakeList fixes that? Perhaps you want to open a third PR with only that change and also enabling again the plugin tests for Windows by deleting the lines you quoted in your post?

quagla:
My suggestion is that we can start from the visualization PR since that doesn't requires the windows tests.

quagla:
Hi @mathl79 I'm pleased to let you know that now you can run plugin tests on Windows as well, see https://github.com/deepmind/mujoco/commit/a9781d9beafd5b078538d41548065bebda6f99aa

mathl79:
Perfect, thanks!
 I was also at the point to implement the runtime dll loading, as I was not able to link the plugin into the elasticity-test executable at compile time.

mathl79:
btw. - you can test the branch for the stress skin visuals without the plasticity in my fork.

quagla:
Looks like we have to rollback the commit unfortunately, hopefully we'll be able to fix it and activate the windows tests in the next days.

quagla:
Good news, Windows is fixed if you rebase from https://github.com/deepmind/mujoco/commit/043ed273a7e90d4f70cc8fcfad8d5f21ffe81eca you'll be able to run Windows tests. 

For the visualisation, are you going to create a separate PR from your branch or would you prefer if we have a look and apply the changes ourselves?

mathl79:
Perfect ! Thanks for fixing again!

For me it absolutely ok, if you take my approach as basis to implement the coloring by our own. 
I do have the feeling, that my implementation is not generic enough to fit into a public release.

quagla:
Hi @mathl79 we are having a look the colouring internally (see https://github.com/deepmind/mujoco/pull/599).

I wanted to ask you what is the rationale for the Von Mises stress computed directly from the curvature stresses without membrane ones? Does this plastic model fit your needs?

mathl79:
Hi @quagla, 
for me the intention for the coloring was just to have an visual indicator for the effects of plasticity. 
The average bending stress is for my case enough information and the projected surface stress calculation of the skin elements was/is to difficult for me. 
Also the coloring on bone level allows non-skin compounds to be colorized.

Regarding the plasticity model:
In the meantime i changed my plasticity model in the way, that omega0 remains unchanged, but the internal bending stress / remaining strain is stored. In that way, the hysteresis behavior is working correctly now, which i did not achive with my first attempt.

I hope, that i will have some spare time the next days to finalize my testing and commit that change into my branch.

br,

Mathias

quagla:
Great to hear! Do you think you can write a unit test in `elasticity_test` comparing the reference vs computed hysteresis so any future change will make sure to not break your intended behaviour?

For the skin colouring, after an internal discussion it was decided to do it in a different way so we will have to implement it ourselves. I agree however that colouring geometries (bones) is probably already giving plenty of information so when you are committing your final changes you could perhaps leave the geom colouring but remove the skin part if you can?

By the way, I added a Jet color palette in my branch, maybe you like it https://github.com/deepmind/mujoco/pull/599/files#diff-f8ed3e4597989d57bd3588f824c10fe89b140ca2665a05dd3b8878f2f6f6c23b

quagla:
Thanks for the update!

Tip: I think you have to set the parameters of CantileverIntoCircle such that perfect elasticity is obtained without triggering a warning because warnings cause test failures.

[ RUN      ] PluginTest.CantileverIntoCircle
D:\a\mujoco\mujoco\test\fixture.cc(43): error: Failed
mju_user_warning: YieldStress is notset: perfect elasticity is assumed !

plugin/elasticity/cable.cc
```diff
@@ -105,7 +190,53 @@ Cable::Cable(const mjModel* m, mjData* d, int instance) {
...
+  if (!UtStress || !UtStrain) {
+      //switch from strain hardening to constant stress model for plastic domain
+      UtStress = YieldStress;
+      UtStrain = 1.0;
+      mju_warning("ultimateStress or ultimateStrain paramters not set: perfect plasticity is assumed (max Stress = Yield Stress)");
+  }
+  else {
+      dYield_UtY = UtStress - YieldStress; //strain hardening stress delta
+      k_deltaStrain_G = 5 / (UtStrain - YieldStrain_G); //exponential constant for torsional strain hardening
+      k_deltaStrain_E = 5 / (UtStrain - YieldStrain_E); //exponential constant for bending strain hardening
```
quagla, line 212: Are these 3 variables unused?

plugin/elasticity/cable.cc
```diff
@@ -26,6 +26,29 @@
 namespace mujoco::plugin::elasticity {
 namespace {
 
+
+//create color scale palette 
+void scalar2rgba(float rgba[4], mjtNum Scalar, mjtNum min, mjtNum max) {
```
quagla, line 31: As mentioned in the main discussion, it would be great to remove the coloring part from this branch.

plugin/elasticity/cable.cc
```diff
@@ -50,28 +73,91 @@ void QuatDiff(mjtNum* quat, const mjtNum body_quat[4],
...
-  mju_quat2Vel(omega, quat, scl);
-
-  // subtract omega0 in reference configuration
-  mjtNum tmp[] = {
-      - stiffness[0]*(omega[0] - omega0[0]) / stiffness[3],
-      - stiffness[1]*(omega[1] - omega0[1]) / stiffness[3],
-      - stiffness[2]*(omega[2] - omega0[2]) / stiffness[3],
-  };
-
+void LocalForce(mjtNum qfrc[3], mujoco::plugin::elasticity::Cable::stiffness_ Sel, mujoco::plugin::elasticity::Cable::stiffness_consts_ Sconst,
```
quagla, line 76: Is the full namespace needed?

plugin/elasticity/cable.cc
```diff
@@ -50,28 +73,91 @@ void QuatDiff(mjtNum* quat, const mjtNum body_quat[4],
...
+    mju_quat2Vel(omega, quat, scl);
+
+    // change of curvature from reference configuration
+    mjtNum d_omega[] = {
+        omega[0] - omega0[0],
+        omega[1] - omega0[1],
+        omega[2] - omega0[2]
+    };
+   
+    //actual strain
```
quagla, line 91: could you always add a space between // and the comment?

plugin/elasticity/cable.cc
```diff
@@ -50,28 +73,91 @@ void QuatDiff(mjtNum* quat, const mjtNum body_quat[4],
...
+    mjtNum strain[] = {
+         d_omega[0] / (Sel.L_Dyz + omega0[0]),
+         d_omega[1] / (Sel.L_Dy + omega0[1]),
+         d_omega[2] / (Sel.L_Dz + omega0[2]),
+    };
+    
+    mjtNum Yield_stress_top[] = {
```
quagla, line 98: Could you document this plastic model with e.g. a paper or a link and/or a more detailed explanation in the code?

plugin/elasticity/cable.cc
```diff
@@ -105,7 +190,53 @@ Cable::Cable(const mjModel* m, mjData* d, int instance) {
...
+      k_deltaStrain_E = 5 / (UtStrain - YieldStrain_E); //exponential constant for bending strain hardening
+      
+     //check if paramters are reasonable. Yieldstrain must not be larger than ultimate strain
+      if (YieldStress / (E > G ? G : E) > UtStrain || UtStress < YieldStress) {
+          mju_error("Error: elasto-plastic material parameters are not reasonable !\nPlease make sure, that Yield Stress < ultimate Stress and Yield Strain (=Yield Stress / Stiffness)  < ultimate Strain.");
+      }
+  }
+  }
+  else {
+      mju_warning("YieldStress is notset: perfect elasticity is assumed !");
```
quagla, line 221: Please update elasticity_test such that this warning is not triggered.

model/plugin/belt.xml
```diff
@@ -57,20 +57,24 @@
...
+        <config key="twist" value="1e10"/>
+        <config key="bend" value="2.5e10"/>
+	      <config key="yieldstress" value="10e7"/>
+	      <config key="ultimatestress" value="2e8"/>
+	      <config key="ultimatestrain" value="0.12"/>
       </plugin>
-      <joint kind="main" damping="0.1"/>
-      <geom type="box" size="0.03535533906 0.02 0.0015" rgba=".8 .2 .1 1" group="3"/>
-      <skin subgrid="3"/>
+      <joint kind="main" damping="10"/>
```
quagla, line 66: If I load this model, it immediately explodes.

model/plugin/coil.xml
```diff
@@ -68,20 +68,26 @@
...
+      <joint kind="main" damping="1" armature="0.01"/>
       <geom type="capsule" size=".005" rgba=".8 .2 .1 1"/>
     </composite>
     <composite prefix="actuated" type="cable" curve="cos(s) sin(s) s" count="41 1 1"
       size="0.251327412 .1 4" offset="0.25 0 .05" initial="fixed">
       <plugin plugin="mujoco.elasticity.cable">
         <!--Units are in Pa (SI)-->
-        <config key="twist" value="5e8"/>
-        <config key="bend" value="15e8"/>
+        <config key="twist" value="1e10"/>
```
quagla, line 84: When I pull the second spring, at some point it looks like the simulation blows up.

quagla (changes requested):
I left some comments. First of all, great work! However some things must be address before continuing:

- The original tests and the original examples must be working.
- A new test for plasticity should be added.
- It would be great to remove the visualization part from the branch, unless absolutely necessary (we should use the test to validate the results instead).

quagla:
We have added coloring of the geometries using a callback, please see https://github.com/deepmind/mujoco/commit/0d52feaa948b92f4ff0e2f844cb51c1d38df1b79

quagla:
Hi @mathl79 are you still interested in submitting this?

yuvaltassa:
closing due to inactivity

PR #567, opened 2022-11-07
Fix mjData comment on pluginstate

liuliu:
It seems there is no npluginstate, just nplugin, I checked the code, seems to be what it means.


saran-t:
This PR is not correct. `npluginstate` is defined [here](https://github.com/deepmind/mujoco/blob/main/include/mujoco/mjmodel.h#L594)

liuliu:
Thanks for the info! Will look into what I can do on my end.

Issue #427, opened 2022-08-05
Using the multiccd option on the Unity plug-in


mushypeas:
I'm attempting to utilize complex meshes (e.g. fruits) in my Unity scene with the MuJoCo Unity plug-in,
but I found out that the objects' behavior are rather 'unstable' when they collide with other meshes.

I tried using the same meshes in the binary build of MuJoCo, and the experimental multiccd option seemed to solve most of the issues I've been going through.
Apparently, the MuJoCo Unity plug-in doesn't support the multi-ccd option yet.
I've tried modifying the global settings script ([`MjGlobalSettings.cs`](https://github.com/deepmind/mujoco/tree/main/unity/Runtime/Components/MjGlobalSettings.cs)) in the package by simply adding multiccd entries, and it seems to have done the work - multiccd becomes available in Unity.
(In case anyone else yet suffers with unstable meshes even after enabling multiccd, try setting the Unity project timestep to 0.005, the default value of mujoco. It worked for me.)

I don't think I'm qualified to issue a pull request, hence leaving this issue.

mushypeas:
Update: I found out that modifying MjGlobalSettings.cs does enable multi-CCD, it just seems to behave slightly differently.
Looks like I'll have to look further into it.

mushypeas:
Update: I found out that modifying MjGlobalSettings.cs does enable multi-CCD, it just seems to behave slightly differently.
Looks like I'll have to look further into it. Will close issue for now.

PR #593, opened 2022-11-18
Add multiccd option in Unity MJCF parser and writer

Balint-H:
Adding the multiccd flags for the `MjGlobalSettings` class as suggested by #427. Tested for importing and exporting.


erez-tom (approved):
Approved for internal review.

PR #597, opened 2022-11-20
Add file dialog box to simulate.

kevinzakka:


saran-t:
Since we already a working, native (as opposed to AppleScript) version on macOS, I'd prefer to just implement the Win32 version ourselves (https://learn.microsoft.com/en-us/windows/win32/dlgbox/open-and-save-as-dialog-boxes), then see if we can just rely on either Zenity or KDialog for Linux users.

kevinzakka:
@saran-t Was just testing out cmake integration, but noted!

PR #507, opened 2022-09-28
unity stand  alone build support  + unity editor mesh  duplicate

MyelinsheathXD:
1.  stand  alone unity build support . encapsulated all  editor scripts with  #if UNITY_EDITOR
since mujoco namespace holding  editor modules which cases to error during stanablone build

https://github.com/MyelinsheathXD/mujocoGameTools/commit/72618db6a9b1382ab7bdb3f37ce7ca8a22c4cdc6

2. On geometry duplicate, the visable   mesh disappear in the editor. On duplicate phase in the editor  , ExecuteinedinMode along with  Update method  removing original shared mesh on duplicate phase   
fixed by calling DestroyImmediate() only when Aplication is in Play mode

https://github.com/MyelinsheathXD/mujocoGameTools/commit/100ad5c91900b8c6d554f59151327164c06015df

the branch fork link https://github.com/MyelinsheathXD/mujocoGameTools/tree/v2.1.5fromComUnity

PR #601, opened 2022-11-23
Fix disappearing meshes on Geom duplicates in Unity Editor

Balint-H:
Addresses issue mentioned by #507, except it solves it not by not deleting meshes in Edit mode (which we want to if the geom changed in edit mode), but by referencing a new mesh when the `MjMeshFilter` is first created. When the new `MeshFilter`'s `sharedMesh` is updated, edited or destroyed, the original geom will now be unaffected. This helps with copy-pasting/duplicating parts of the model quickly without needing to update the old geoms.


erez-tom (approved):
Approving for internal review.

PR #602, opened 2022-11-23
Remove unnecessary unit conversion in MjSlideJoint in Unity plugin

Balint-H:
Fix typo in `OnSyncState` of `MjSlideJoints`, which previously converted length to degrees.


erez-tom (approved):
Approving for internal review

PR #619, opened 2022-12-03
Python bindings: skip tests that cannot run on PyPy due to the use of sys.getrefcount

traversaro:
On conda-forge we also build with PyPy (https://www.pypy.org/). All tests work fine, except for two that use `sys.getrefcount` that is an implementation detail of CPython (see https://github.com/conda-forge/pysynphot-feedstock/pull/7#issuecomment-1300329664). This PR propose to skip those tests on PyPy.


saran-t:
Are you 100% sure that the bindings work with PyPy? We have not tested it at all, I'm hesitant to give an impression that this is supported unless you have good reasons to believe that this is actually functioning correctly.

traversaro:
> Are you 100% sure that the bindings work with PyPy? We have not tested it at all, I'm hesitant to give an impression that this is supported unless you have good reasons to believe that this is actually functioning correctly.

All the test suite (except for these two tests) pass fine.

traversaro:
Closing anyhow as I had a misunderstanding on the use of pytest/absltest. I will reopen when I will now that also the skip test logic is working.

traversaro:
Ok, now the PR should be working fine. Beside running all the testsuite, I also tried the rendering code in https://github.com/deepmind/mujoco/blob/2.3.0/python/LQR.ipynb with PyPy 3.9 and it was working fine.

Anyhow, clearly any change not related to a configuration that you test in your CI (including for example the use of non-vendored dependencies) is prone to break in the future, either by change in mujoco or in its chain of dependencies. The only reason I submitted this PR is that it seemed quite a minor change that did not interfere in any way with the existing functionality, and could have been useful for other users that used PyPy with MuJoCo. If you fear that this could lead misunderstanding on PyPy being actually officially supported, I think the possible options are:

* Not merge anything related to PyPy. Note that probably this could apply also to any other non-tested configuration (CMake option, dependency version, ..)
* Merge PyPy-related changes, but clearly documenting that this is not an officially supported configuration. This is what some opensource project do, see "Tier 3" in ROS2 (https://www.ros.org/reps/rep-2000.html#support-tiers), "Community Triplet" in vcpkg (https://vcpkg.readthedocs.io/en/latest/users/triplets/). 

I am fine with either choice.

saran-t:
I'm ok with adding these minor PyPy-specific escape hatches, but could you add comments next to the pypy exclusion lines that says PyPy is not supported?

traversaro:
> I'm ok with adding these minor PyPy-specific escape hatches, but could you add comments next to the pypy exclusion lines that says PyPy is not supported?

Done!

python/mujoco/bindings_test.py
```diff
@@ -128,6 +128,10 @@ def test_array_is_a_view(self):
     np.testing.assert_array_equal(
         qpos_ref, [0.789]*len(self.data.qpos))
 
+  # This test is disabled on PyPy as it uses sys.getrefcount
+  # However PyPy is not officially supported by MuJoCo
+  @absltest.skipIf(sys.implementation.name == 'pypy', 
+                      reason="requires sys.getrefcount")
```
saran-t, line 134: Please align with previous line.

python/mujoco/bindings_test.py
```diff
@@ -962,6 +966,10 @@ def __call__(self, m, d, stage):
     self.assertEqual(sensor_callback.count, 1)
     self.assertEqual(data_with_sensor.sensordata[0], 17)
 
+  # This test is disabled on PyPy as it uses sys.getrefcount
+  # However PyPy is not officially supported by MuJoCo
+  @absltest.skipIf(sys.implementation.name == 'pypy', 
+                      reason="requires sys.getrefcount")
```
saran-t, line 972: Also here.

saran-t (approved):

PR #628, opened 2022-12-06
Reuse existing mesh assets in MjMeshFilter

Balint-H:
Change in the Unity plugin. If a Geom uses a mesh at import, its shape won't be procedurally changed in the editor, therefore the existing mesh asset created at import can be used instead of creating and storing one in the scene. This significantly improves scene load and unload times on mesh-heavy scenes with about an order of magnitude.


erez-tom (approved):
Approving for internal review.

PR #629, opened 2022-12-06
Don't run Github CI for changes under `doc/*`.

kevinzakka:


saran-t:
Do you know if the check will still show up as passing when a commit only contains ignored files?

saran-t (approved):

PR #636, opened 2022-12-07
Pass glfw3 as PACKAGE_NAME in glfw's findorfetch invocations

traversaro:
The correct way to find glfw when it is installed in the system is to call `find_package(glfw3)` (see https://www.glfw.org/docs/3.3/build_guide.html#build_link_cmake_package), hence the correct parameter to pass as `PACKAGE_NAME` is glfw3 .


saran-t (approved):

traversaro:
Thanks @saran-t !

PR #638, opened 2022-12-08
Fix exception causes in introspect/type_parsing.py

cool-RR:
More context for this change here: https://blog.ram.rachum.com/post/621791438475296768/improving-python-exception-chaining-with


saran-t:
Were you actually poking around in `introspect` and ran into impossible to debug issues?

saran-t (approved):

PR #652, opened 2022-12-14
Don't access MjScene in OnDisable() if it doesn't exist

Balint-H:
Previously `MjComponent`s requested scene rebuilding at `OnDisable()` from `MjScene` even if it was already destroyed during scene transition. This generated a new MjScene during Scene cleanup, which resulted in a non-breaking error message in the Editor (and I assume some undesired side-effects).


erez-tom (approved):
Approved for internal review.

PR #657, opened 2022-12-15
Fix ignored angle units of springref during Unity import

### LINKED ISSUE #654: Robotiq Gripper closing behavior is incorrect in Unity+MuJoCo
Hi,

I'm a research engineer at a humanoid robotics company and I'm trying to use the MuJoCo Unity plug-in to dynamically simulate opening and closing the Robotiq gripper from the Mujoco Menagerie in Unity. [I previously reported an error](https://github.com/deepmind/mujoco/issues/647) I ran into that I was able to work past with the suggested hard-coded fix.

I'm now able to simulate the opening and closing behavior in both Unity (using MuJoCo) and with stand-alone MuJoCo. Unfortunately, the Unity simulation is not closing properly when actuated (and seems to start in an overextended state initially).

I've attached a video recording of the gripper actuation in Unity and in MuJoCo for comparison. Both of these are using the same `2f85.xml` file from my [minimal repo](https://github.com/james-giammona/mujoco-unity-error-repro) which has sub-sampled meshes for some of the gripper parts.

While poking around, I noticed that Unity and MuJoCo had different behavior when I deleted all equality constraints, unless I also turned off the Limits flag in the MuJoCo simulation and then the behavior of both was much closer (with the outer arm bars hanging fully down and oscillating continuously) so my guess is that the limits may not be being set properly in Unity during initialization?

Please let me know if you'd like any other information or further videos.

MuJoCo video:

https://user-images.githubusercontent.com/120140073/207705455-1250b791-ce36-430b-af31-7292dd01b499.mp4

Unity video:

https://user-images.githubusercontent.com/120140073/207705506-ae97acd2-54ca-4a63-a2c8-db1bfe79d7e4.mp4

Balint-H:
Partially fixes #654, applies multiplier depending on `MjSceneImportSettings.AnglesInDegrees`.

### DIFF
--- a/unity/Runtime/Components/Joints/MjJointSettings.cs
+++ b/unity/Runtime/Components/Joints/MjJointSettings.cs
@@ -72,7 +72,7 @@ namespace Mujoco {
       var springDamper = mjcf.GetVector2Attribute("springdamper", Vector2.zero);
       TimeConstant = springDamper.x;
       DampingRatio = springDamper.y;
-      EquilibriumPose = mjcf.GetFloatAttribute("springref", 0.0f);
+      EquilibriumPose = MjSceneImportSettings.AnglesInDegrees? mjcf.GetFloatAttribute("springref", 0.0f) : mjcf.GetFloatAttribute("springref", 0.0f) * Mathf.Rad2Deg;
       Damping = mjcf.GetFloatAttribute("damping", 0.0f);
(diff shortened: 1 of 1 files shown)


erez-tom (commented):
Approved for internal review.

erez-tom (approved):

PR #705, opened 2023-01-21
Fix infinite loop in `mj_deleteFileVFS`

hartikainen:
In its current form, the `mj_deleteFileVFS` results in infinite loop with more than one file. This should fix the issue.

### DIFF
--- a/src/engine/engine_vfs.c
+++ b/src/engine/engine_vfs.c
@@ -179,7 +179,7 @@ int mj_deleteFileVFS(mjVFS* vfs, const char* filename) {
       mju_free(vfs->filedata[i]);
 
       // scroll remaining files forward
-      for (int j = i; j < vfs->nfile - 1; ++j) {
+      for (int j=i; j<vfs->nfile-1; j++) {
         mjSTRNCPY(vfs->filename[j], vfs->filename[j+1]);
         vfs->filesize[j] = vfs->filesize[j+1];
         vfs->filedata[j] = vfs->filedata[j+1];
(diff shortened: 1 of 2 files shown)


src/engine/engine_vfs.c
```diff
@@ -179,10 +179,10 @@ int mj_deleteFileVFS(mjVFS* vfs, const char* filename) {
       mju_free(vfs->filedata[i]);
 
       // scroll remaining files forward
-      while (i<vfs->nfile-1) {
-        mjSTRNCPY(vfs->filename[i], vfs->filename[i+1]);
-        vfs->filesize[i] = vfs->filesize[i+1];
-        vfs->filedata[i] = vfs->filedata[i+1];
+      for (int j = i; j < vfs->nfile - 1; ++j) {
```
yuvaltassa, line 182: C files in `engine/` have a somewhat [idiosyncratic coding convention](https://github.com/deepmind/mujoco/blob/main/STYLEGUIDE.md). I'm not going to say that I love it, but it is what it is and consistency trumps everything when style is concerned. To wit, `for (int j=i; j<vfs->nfile-1; j++)` would be consistent with the rest of the C code.

This comment does not apply to other code (e.g. test files).

saran-t, line 182: This loop should start at i+1 not i.

hartikainen: I think `i` is correct. We delete the `i`th file and then always move e.g. `vfs->filesize[j] = vfs->filesize[j+1];`.

hartikainen: Fixed.

hartikainen (commented):
Changed

src/engine/engine_vfs.c
saran-t: You're right.

saran-t (approved):

PR #734, opened 2023-02-13
Add support for loading balloons.xml in Unity plugin

hyarion:
Hi everyone, first timer here! 👋
When testing out MuJoCo I noticed that the balloons model didn't load for me with the unity plugin, so here's a patch which makes it work.

Let me know if you have any feedback 🙂 

---

Steps to reproduce the issue, in unity with plugin setup correctly:
1. Try to load the model located at `model/balloons/balloons.xml`

Result: It fails to load

---

The issues, causing it not to work in unity as expected, was:
* Missing support for the `implicit` integrator type - model failed to load before
* Missing support for `gravcomp` - balloons where just falling down
* Default wind in imported scene - not matching the stand alone simulator, nor the documentation
* The importer skipped defaults for tendons - the strings attaching the balloons to the weight wasn't fixed in length due to missing `limited="true"` attribute

I also fixed a typo in a comment in close proximity to one of the fixes. Let me know if you would prefer this fix to be separated to a new pull request instead.

### DIFF
--- a/unity/Runtime/Components/Bodies/MjBody.cs
+++ b/unity/Runtime/Components/Bodies/MjBody.cs
@@ -20,17 +20,23 @@ namespace Mujoco {
 
   // The component represents the apex of hierarchy that defines a single rigid body.
   public class MjBody : MjBaseBody {
+
+    [Tooltip("Gravity compensation force, specified as fraction of body weight.")]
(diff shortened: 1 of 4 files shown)


erez-tom (approved):
Approved for internal review.

PR #740, opened 2023-02-17
Add support for fluid shapes to Unity plugin

hyarion:
## Background
~~Tying a buoyant object to the world off center (horizontal offset to CoM), results currently in a never ending oscillation, even if the global settings specifies a viscosity and mass + density are specified.~~ [Edit: This was not the case, see discussion]

## The problem
In the XML Reference, there's a `fluidshape` attribute that can be used to introduce drag, but this feature is not exposed in the Unity Plugin, nor is it implied.

## Implementation
I've exposed the `fluidshape` attribute for a `geom` as `FluidShapeType` in `MjGeomSettings`, as this is where more advanced settings for geometries are placed. I assume you want it to be placed here to simplify the ux for new users, but let me know if you prefer it directly in `MjGeom`.

The name `FluidShapeType` isn't a 1:1 mapping to the XML attribute, but the naming tries to conform to the style used when `ShapeType` was named.
In my opinion, it's more important that it looks similar to `ShapeType` then being called the same was as in XML.

### DIFF
--- a/unity/Runtime/Components/Shapes/MjGeomSettings.cs
+++ b/unity/Runtime/Components/Shapes/MjGeomSettings.cs
@@ -42,12 +42,17 @@ public struct MjGeomSettings {
   [Tooltip("Shape used for fluid simulation.")]
   public FluidShapeTypes FluidShapeType;
 
+  [Tooltip("Dimensionless coefficients of fluid interaction model.")]
+  public GeomFluidCoefficients FluidCoefficients;
+
+
   // Default geom settings.
   public static MjGeomSettings Default = new MjGeomSettings() {
     Solver = GeomSolver.Default,
     Filtering = CollisionFiltering.Default,
     Friction = GeomFriction.Default,
-    FluidShapeType = FluidShapeTypes.None
+    FluidShapeType = FluidShapeTypes.None,
+    FluidCoefficients = GeomFluidCoefficients.Default
   };
 
   public void FromMjcf(XmlElement mjcf) {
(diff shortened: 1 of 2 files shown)


erez-tom (commented):
If you find the "fluid" feature useful, maybe it would make sense to also expose the `fluidcoef` field (either as 5 separate numeric fields or even just a text field that's validated to have at most 5 space-separated numbers)?

yuvaltassa:
> even if the global settings specifies a viscosity and mass + density are specified

This should not occur as there is already a basic fluid model parametrized by these numbers.
Did the change you are proposing actually fix the problem?

hyarion:
@yuvaltassa You are correct, I must have missed something before causing me to misinterpret what was going on.
I recreated a simplified model, similar to what I had, which shows viscosity drag in action both in simulate and in unity:
```xml
<mujoco>
  <option density="1000" viscosity="1.5e-3"/>
  <worldbody>
    <geom type="plane" size="5 5 .01" rgba="1 1 1 1"/>

    <body pos="0 0 0.1" quat="-1 0 0 0" gravcomp="0.63">
      <joint type="free"/>
      <!-- shell, set mass as it's hollow (density=2700) -->
      <geom mass="0.889" size="0.15 0.1 0.1" type="ellipsoid"/>

      <!-- hollow volume use density of surrounding media but compensate with gravcomp -->
      <body pos="0.01 0 0" quat="-1 0 0 0" gravcomp="1.055">
        <geom density="1000" size="0.14 0.098 0.098" type="ellipsoid"/>
      </body>
    </body>
  </worldbody>
</mujoco>
```

That said, I guess it doesn't hurt to expose this if someone would need it.

---

As suggested by @erez-tom I've also added support for `fluidcoef` as a seperate commit.
I've separated the coef similar to how the settings for solver and friction is sectioned in `GeomSettings`.

erez-tom (approved):
Approved for internal review.

PR #769, opened 2023-03-14
Ensure that loaded plugins are a shared library

milutter:
The loading of the plugins in `python/mujoco/__init__.py` can fail depending on the used build environment. When the described installation process is followed, everything works fine. However, some build environments can "generously" add `__init__.py` files to all sub-directories, which causes the error `OSError: [...]/mujoco/site-packages/mujoco/plugin/__init__.py: invalid ELF header` as the added `plugin/__init__.py` file is passed to `ctypes.CDLL(...)`.

### DIFF
--- a/python/mujoco/__init__.py
+++ b/python/mujoco/__init__.py
@@ -61,8 +61,8 @@ def _load_all_bundled_plugins():
             pass
 
         else:
-            raise ValueError(f"Trying to load the plugin {os.path.join(directory, filename)}, "
-                             "which is not a shared library.")
+            raise ValueError(f"Trying to load the plugin {os.path.join(directory, filename)} "
+                             "that is not a shared library.")
 
 _load_all_bundled_plugins()
 
--- a/python/mujoco/__init__.py
+++ b/python/mujoco/__init__.py
@@ -19,6 +19,7 @@ import ctypes.util
 import os
 import platform
 import subprocess
+import warnings
 
 _SYSTEM = platform.system()
 if _SYSTEM == 'Windows':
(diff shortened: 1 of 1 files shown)


python/mujoco/__init__.py
```diff
@@ -54,7 +54,15 @@
 def _load_all_bundled_plugins():
   for directory, _, filenames in os.walk(PLUGINS_DIR):
     for filename in filenames:
-      PLUGIN_HANDLES.append(ctypes.CDLL(os.path.join(directory, filename)))
+        if os.path.splitext(filename)[-1] == ".so":
```
saran-t, line 57: This need to be platform-dependent. See https://github.com/deepmind/mujoco/blob/main/python/setup.py#L73

saran-t, line 57: Indent with 2 spaces.

saran-t, line 57: Indent with 2 spaces.

python/mujoco/__init__.py
```diff
@@ -54,7 +54,15 @@
...
+            PLUGIN_HANDLES.append(ctypes.CDLL(os.path.join(directory, filename)))
+
+        elif filename == "__init__.py":
+            pass
```
saran-t, line 61: Indent with 2 spaces.

python/mujoco/__init__.py
```diff
@@ -54,7 +54,15 @@
...
+
+        else:
+            raise ValueError(f"Trying to load the plugin {os.path.join(directory, filename)} "
```
saran-t, line 64: I prefer `warnings.warn(f'Ignoring non-library in plugin directory: {filename}', ImportWarning)`.

Also, indent with 2 spaces.

python/mujoco/__init__.py
milutter: Fixed.

python/mujoco/__init__.py
milutter: Fixed

python/mujoco/__init__.py
milutter: Fixed

python/mujoco/__init__.py
milutter: Fixed. I added the file extensions `.dll`, `.dylib` & `.so`. Are you fine with just matching the extension or do you want me to match the complete filename as in the linked `setup.py`?

PR #801, opened 2023-04-04
Fix `mj_sizeModel`

hartikainen:
Before these changes, the included `EngineIoTest.VerifySizeModel` test fails with the following:
```
mujoco/test/engine/engine_io_test.cc:79: Failure
Expected equality of these values:
  file_size
    Which is: 2416
  mj_sizeModel(model)
    Which is: 6488
```

With the fix, the test passes.

### DIFF
--- a/test/engine/engine_io_test.cc
+++ b/test/engine/engine_io_test.cc
@@ -18,7 +18,11 @@
 
 #include <array>
 #include <climits>
+#include <cstdio>
 #include <cstring>
+#include <filesystem>
+#include <fstream>
+#include <iostream>
 #include <string>
 
 #include <gmock/gmock.h>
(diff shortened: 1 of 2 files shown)


saran-t (approved):

saran-t:
This is failing on MSVC..

test/engine/engine_io_test.cc
```diff
@@ -47,6 +51,34 @@ mjModel PartialModel(const mjModel* m) {
...
+      </body>
+    </worldbody>
+  </mujoco>
+  )";
+
+  std::array<char, 1024> error;
+  mjModel* model = LoadModelFromString(xml, error.data(), error.size());
+  ASSERT_THAT(model, NotNull()) << "Failed to load model: " << error.data();
+
+  std::filesystem::path temp_file = (
```
nimrod-gileadi, line 70: This test fails on Windows (with MSVC). We have another test that creates temporary files like this [here](https://github.com/deepmind/mujoco/blob/c17375f/test/fixture.cc#L112).

Maybe you can find out why that utility works and this test doesn't? Maybe you'll need to explicitly convert the `std::filesystem::path` it to a `std::string` by calling `.string()`.

nimrod-gileadi (changes requested):

test/engine/engine_io_test.cc
hartikainen: You were right, adding `.string()` fixed the issue.

saran-t (approved):

test/engine/engine_io_test.cc
```diff
@@ -47,6 +51,34 @@ mjModel PartialModel(const mjModel* m) {
...
+    std::filesystem::temp_directory_path() / "model.mjb");
+
+  mj_saveModel(model, temp_file.string().c_str(), NULL, 0);
+
+  std::uintmax_t file_size = std::filesystem::file_size(temp_file);
+
+  std::filesystem::remove(temp_file);
+
+  EXPECT_EQ(file_size, mj_sizeModel(model));
+}
```
saran-t, line 80: The test is leaking `mjModel` (picked up by the leak checker internally). Please fix.

saran-t (changes requested):

test/engine/engine_io_test.cc
hartikainen: Done.

saran-t (approved):

PR #896, opened 2023-06-02
Fix the effective time constant function for muscle

Beanpow:
At the latest docs of modeling section, the effective time constant function for muscle just has one parameter, ``ctrl-act``. But ``act``    is used in this function. So, I think it should be $\tau(\text{ctrl}, \text{act})$ instead of $\tau(\text{ctrl} - \text{act})$.

image


yuvaltassa (approved):

yuvaltassa:
thanks!

PR #910, opened 2023-06-08
Sample Return Codes

nicholask42:
Modified samples to return EXIT_SUCCESS OR EXIT_FAILURE instead of 0 or 1. Also fixed incorrect return at the end of main method of basic.cc.

### DIFF
--- a/sample/derivative.cc
+++ b/sample/derivative.cc
@@ -284,7 +284,7 @@ int main(int argc, char** argv) {
   // print help if not enough arguments
   if (argc<2) {
     std::printf("\n Arguments: modelfile [nthread niter nwarmup nepoch nstep eps]\n\n");
-    return 1;
+    return EXIT_FAILURE;
   }
 
   // default nthread = number of logical cores (usually optimal)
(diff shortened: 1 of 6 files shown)


saran-t:
Thanks!

I'm afraid we'll need a CLA (https://cla.developers.google.com/) before we can merge this though.

nicholask42:
Sorry for the delay @saran-t. I have completed the CLA, run the check again, and it has passed.

saran-t (approved):

sample/derivative.cc
```diff
@@ -284,7 +284,7 @@ int main(int argc, char** argv) {
   // print help if not enough arguments
   if (argc<2) {
     std::printf("\n Arguments: modelfile [nthread niter nwarmup nepoch nstep eps]\n\n");
-    return 1;
+    return EXIT_FAILURE;
```
saran-t, line 287: You chose to return `EXIT_SUCCESS` for the equivalent branch ("print help") in the other programs, so let's do that here too.

saran-t (approved):

yuvaltassa:
Fixed by 5187a71

PR #916, opened 2023-06-10
Corrected index.rst example code samples

jc-cr:
Added the missing humanoid directory path to window and linux sample terminal commands


yuvaltassa (approved):
Thanks!

PR #947, opened 2023-06-22
Add MJCF exporter from Unity based on the importer class

Balint-H:
I see using the Unity plugin as a scene construction tool as one of its primary potential use cases. However the existing method for transferring scenes out from Unity is not very user friendly.

This Editor object avoids having to use the debug file from MjGlobalSettings and launching a play mode, and allows saving in a specific location. 

It creates and destroys a MjScene for this purpose to use MjScene's already implemented MJCF construction.

The generated MJCF is not exactly pretty. Having a method to automatically extract defaults or not explicitly write unchanged fields could be further improvements.


erez-tom (approved):
Approved for internal review.

PR #1006, opened 2023-08-02
Updating API to include the union types for mjuiItem

JamieMair:
Also changed the docs for running the clang command to generate the AST, as this errored without the include directory. I am not sure if this is the command that should be used, but it does not encounter any errors.

The chain change is modifying the header file `mjui.h` to add typedefs for the structs used by `mjuiItem` so that they can be discovered and generated by the `introspect` module.

@yuvaltassa @saran-t  I noticed that these typedefs were missing when I was writing code to base our Julia API on your `introspect` module. Also many thanks for your advice pointing us towards this module, it's incredibly helpful.


yuvaltassa:
closing due to staleness, please reopen a new PR if still relevant

PR #1034, opened 2023-08-27
Modify mjd_inverseFD for discrete inverse dynamics

thowell:
Modifications to `mjd_inverseFD` to compute correct Jacobians when discrete inverse dynamics are used.

When `mj_inverse` is called with discrete inverse dynamics enabled, `qacc` is corrected to a continuous time representation. As a result, we need to reset `qacc` to its discrete representation prior calling `mj_inverse`.


yuvaltassa:
Does it not make more sense to do this in mj_inverse?

yuvaltassa:
I believe this is solved by f44c6575444fc393d74ba937ba432c51f55bba2e ?

PR #1038, opened 2023-09-02
feat: added _Nonnull and _Nullable attributes to mujoco.h

### LINKED ISSUE #309: Adding _Nonull and _Nullable attributes to `mujoco.h` APIs
**Is your feature request related to a problem? Please describe.**
When using MuJoCo API, it is a bit harder to parse either automatically or from documentation when a parameter can be NULL or not. For example, `mj_step` requires `mjData* d` to be nonnull, while `mj_copyData` can take NULL `mjData* dest`. Similarly, `mjv_makeScene` can take NULL `const mjModel* m`. There seems to be no systematic way to understand which parameter can be NULL which cannot.

**Describe the solution you'd like**
New compilers, such as clang starts to support _Nullable attribute: https://clang.llvm.org/docs/AttributeReference.html#nullable
These can be `#define` away on unsupported compilers (For example, on GCC, you can define that as `__attribute__((nonnull))`: https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#Common-Function-Attributes

**Describe alternatives you've considered**
This information could be annotated in the comment like how we annotated length of array in public struct. It doesn't provide the said compiler check.

**Additional context**
These nullability attributes can be helpful when generating interface glue code to MuJoCo library. For example, [swift-mujoco](https://github.com/liuliu/swift-mujoco) right now assumes all fields to be nonnull and have to manually implement selected APIs that can be NULL. If nullability attribute added, these can be automatically generated like the rest of APIs.

varshneydevansh:
## The basics

- [x] I branched from main
- [x] My pull request is against main
- [x] My code follows the DeepMind Style Guide
- [x] I ran `git clang-format`

### Resolves

- Fixes #309 

### Proposed Changes
New compilers, such as clang starts to support _Nullable attribute: https://clang.llvm.org/docs/AttributeReference.html#nullable
These can be #define away on unsupported compilers (For example, on GCC, you can define that as __attribute__((nonnull)): https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#Common-Function-Attributes

#### Behavior Before Change
When using MuJoCo API, it is a bit harder to parse either automatically or from documentation when a parameter can be NULL or not. For example, mj_step requires mjData* d to be nonnull, while mj_copyData can take NULL mjData* dest. Similarly, mjv_makeScene can take NULL const mjModel* m. There seems to be no systematic way to understand which parameter can be NULL which cannot.

#### Behavior After Change
These nullability attributes can be helpful when generating interface glue code to MuJoCo library. For example, [swift-mujoco](https://github.com/liuliu/swift-mujoco) right now assumes all fields to be nonnull and have to manually implement selected APIs that can be NULL. If nullability attribute added, these can be automatically generated like the rest of APIs.

### Reason for Changes
When using MuJoCo API, it is a bit harder to parse either automatically or from documentation when a parameter can be NULL or not. 

### Test Coverage

I haven't tested this, as soon as I got confirmation that these are the required changes that will update the status of the Test Coverage.

### DIFF
--- a/include/mujoco/mjmacro.h
+++ b/include/mujoco/mjmacro.h
@@ -60,20 +60,21 @@
   }
 #endif
 
-//-------------------------- nullability check ----------------------------------------------------
+//-------------------------- nullability check
+//----------------------------------------------------
 
 #ifdef __clang__
 #define NONNULL_ARG _Nonnull
 #define NULLABLE_ARG _Nullable
-#define NONNULL_FUNC(...)                         /* NOT SUPPORTED */
+#define NONNULL_FUNC(...) /* NOT SUPPORTED */
 #elif defined(__GNUC__)
-#define NONNULL_ARG                               /* NOT SUPPORTED */
-#define NULLABLE_ARG                              /* NOT SUPPORTED */
+#define NONNULL_ARG  /* NOT SUPPORTED */
+#define NULLABLE_ARG /* NOT SUPPORTED */
 #define NONNULL_FUNC(...) __attribute__((nonnull(__VA_ARGS__)))
 #else
-#define NONNULL_ARG                              /* NOT SUPPORTED */
-#define NULLABLE_ARG                             /* NOT SUPPORTED */
-#define NONNULL_FUNC(...)                        /* NOT SUPPORTED */
+#define NONNULL_ARG       /* NOT SUPPORTED */
+#define NULLABLE_ARG      /* NOT SUPPORTED */
+#define NONNULL_FUNC(...) /* NOT SUPPORTED */
 #endif
 
 #endif  // MUJOCO_MJMACRO_H_
--- a/include/mujoco/mjmacro.h
+++ b/include/mujoco/mjmacro.h
@@ -60,4 +60,20 @@
   }
 #endif
 
+//-------------------------- nullability check ----------------------------------------------------
+
+#ifdef __clang__
+#define NONNULL_ARG _Nonnull
+#define NULLABLE_ARG _Nullable
+#define NONNULL_FUNC(...)                         /* NOT SUPPORTED */
+#elif defined(__GNUC__)
+#define NONNULL_ARG                               /* NOT SUPPORTED */
+#define NULLABLE_ARG                              /* NOT SUPPORTED */
+#define NONNULL_FUNC(...) __attribute__((nonnull(__VA_ARGS__)))
+#else
+#define NONNULL_ARG                              /* NOT SUPPORTED */
+#define NULLABLE_ARG                             /* NOT SUPPORTED */
+#define NONNULL_FUNC(...)                        /* NOT SUPPORTED */
+#endif
+
 #endif  // MUJOCO_MJMACRO_H_
--- a/include/mujoco/mujoco.h
+++ b/include/mujoco/mujoco.h
@@ -103,46 +103,54 @@ MJAPI void mj_deleteVFS(mjVFS* vfs);
 // Parse XML file in MJCF or URDF format, compile it, return low-level model.
 // If vfs is not NULL, look up files in vfs before reading from disk.
 // If error is not NULL, it must have size error_sz.
-MJAPI mjModel* mj_loadXML(const char* NONNULL_ARG filename, const mjVFS* NULLABLE_ARG vfs, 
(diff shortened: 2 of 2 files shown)


include/mujoco/mujoco.h
```diff
@@ -103,43 +103,54 @@ MJAPI void mj_deleteVFS(mjVFS* vfs);
...
 // Forward dynamics with skip; skipstage is mjtStage.
-MJAPI void mj_forwardSkip(const mjModel* m, mjData* d, int skipstage, int skipsensor);
+MJAPI void mj_forwardSkip(const mjModel* NONNULL_ARG m, mjData* NONNULL_ARG d,
+                          int skipstage, int skipsensor) NONNULL_FUNC(1, 2);
 
 // Inverse dynamics with skip; skipstage is mjtStage.
-MJAPI void mj_inverseSkip(const mjModel* m, mjData* d, int skipstage, int skipsensor);
-
+MJAPI void mj_inverseSkip(const mjModel* NONNULL_ARG m, mjData* NONNULL_ARG d,
+                          int skipstage, int skipsensor) NONNULL_FUNC(1, 2);
```
varshneydevansh, line 153: after running -

```elixir
➜  mujoco git:(309-Adding-_Nonull-and-_Nullable-attributes-to-mujoco.h) git clang-format HEAD~1 
changed files:
    include/mujoco/mjmacro.h
    include/mujoco/mujoco.h
➜  mujoco git:(309-Adding-_Nonull-and-_Nullable-attributes-to-mujoco.h) git clang-format HEAD~1 
changed files:
    include/mujoco/mujoco.h
```

The asterisk * is becoming insistent with the rest of the code for the pointers in the updated definitions.

Moreover I am looking at the conflict which is going with the mujoco.h file.

varshneydevansh: Is this the conflict between a version of the file that uses the NONNULL_ARG and NONNULL_FUNC macros, and a version that does not use those macros?

varshneydevansh: I will look into the changes with the comments in the later part of the mujoco.h file in the morning.

The changes I made are just before the ray-collisions. =)

include/mujoco/mujoco.h
```diff
@@ -103,43 +103,54 @@ MJAPI void mj_deleteVFS(mjVFS* vfs);
...
+                          const mjVFS* NULLABLE_ARG vfs,
+                          char* NULLABLE_ARG error, int error_sz)
+    NONNULL_FUNC(1);
 
 // Update XML data structures with info from low-level model, save as MJCF.
 // If error is not NULL, it must have size error_sz.
-MJAPI int mj_saveLastXML(const char* filename, const mjModel* m, char* error, int error_sz);
+MJAPI int mj_saveLastXML(const char* NONNULL_ARG filename,
+                         const mjModel* NONNULL_ARG m, char* NULLABLE_ARG error,
+                         int error_sz) NONNULL_FUNC(1, 2);
```
yuvaltassa, line 115: Let's always have NONNULL (formerly known as NONNULL_FUNC) on its own line?

include/mujoco/mjmacro.h
```diff
@@ -60,4 +60,21 @@
   }
 #endif
 
+//-------------------------- nullability check
+//----------------------------------------------------
+
+#ifdef __clang__
+#define NONNULL_ARG _Nonnull
```
yuvaltassa, line 67: 1. Please indent the `#ifdef` bodies.
2. I don't think we need NONNULL_ARG. Clang [supports](https://clang.llvm.org/docs/AttributeReference.html#id12) the GCC syntax.
3. Due to 2, let's rename `NONNULL_FUNC` -> `NONNULL` and `NULLABLE_ARG` -> `NULLABLE`.
4. I think this whole block belongs in mujoco.h and requires `#undef`s at the end

yuvaltassa (commented):
Preliminary review, until @saran-t takes a look 🙂

include/mujoco/mjmacro.h
varshneydevansh: Will do it and also I will restore the other unintentional changes that happened.

yuvaltassa:
Fixed in 32c7d3f085b4fe063a4de79117e6da0e49ca31ed

PR #1070, opened 2023-09-25
simulate: fix display of loading message

aftersomemath:
Model loading was moved to the physics thread sometime back (for good reason). This broke the "loading" message shown in UI because the UI did not a know a model was being loaded until after the fact.

The proposed solution adds optional methods to the Simulate class that set and clear a loading message as part of the UI's model loading state machine.

This works with C++ simulate application and the Python managed viewer. It should not affect the passive viewer.

EDIT: It also moves the loading message for command line based reloads to the left of the screen. Previously it was on the right for command line filenames and the left for UI based reloads (button or drag-n-drop) which is inconsistent.

### DIFF
--- a/simulate/simulate.h
+++ b/simulate/simulate.h
@@ -62,9 +62,17 @@ class Simulate {
   void UpdateMesh(int meshid);
   void UpdateTexture(int texid);
 
+  // Request that the Simulate UI display a "loading" message
+  // Called prior to Load or LoadMessageClear
+  void LoadMessage(const char* displayed_filename);
+
   // Request that the Simulate UI thread render a new model
-  // optionally delete the old model and data when done
   void Load(mjModel* m, mjData* d, const char* displayed_filename);
+
+  // Clear the loading message
+  // Can be called instead of Load to clear the message without
+  // requesting the UI load a model
+  void LoadMessageClear(void);
 
   // functions below are used by the renderthread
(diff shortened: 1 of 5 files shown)


simulate/simulate.h
```diff
@@ -173,6 +181,7 @@ class Simulate {
   std::atomic_int uiloadrequest = 0;
 
   // loadrequest
+  //   3: display a loading message
```
yuvaltassa, line 184: Why do we need "3" Isn't this already what "1" does (did 🙂) ?

aftersomemath: There are two stages to the loading process. Transitioning from state 3 to 2 is handled by the physics thread (this is optional for the user). Transitioning from 2 to 0 is handled by the UI thread.

The UI thread needs two states so it can put a loading message on the screen (transition from 2 to 1) prior to the blocking call to Simulate's LoadOnRenderThread (transition from 1 to 0).

yuvaltassa:
Can you resync please so the tests pass? 
I think you happened to send this right after we broke something but before we fixed it 🙂

aftersomemath:
For sure, it is done.

yuvaltassa (approved):

yuvaltassa:
@aftersomemath thanks so much ❤️

aftersomemath:
You are welcome!

Issue #691, opened 2023-01-10
Clarification - Coordinate frame of generalized forces, velocities and accelerations of free joints


Anderv21:
Hi,

I'm a PhD. student and I'm trying to use MuJoCo for the simulation of a real, legged robot. I need to design, test and deploy model-based controllers for such a robot and I have chosen MuJoCo for the testing phase. My application uses MuJoCo in a rather simple manner: sends actuator commands and receives sensor data. I already have a running application that performs the aforementioned functionality. 

I want to make sure that the simulated model behaves as close as possible to the actual hardware (same input and output list and same _dynamic behavior_). Following the documentation, I have been able to reproduce all inputs and outputs of my real system (with the coordinate frames I want for 3D quantities). Furthermore, I believe that all geometric (/kinematic) and inertial parameters are the same as the model I use in my controller. To do a final sanity check I want to compare the Mass matrix and generalized forces due to non-linear effects (gravity and velocity terms) and possibly the end-effector jacobians of my mathematical model with the matrices that MuJoCo computes during runtime. 

Here's where I need some help: the coordinate frames of the generalized velocities, accelerations and forces of the free-floating joint are not completely clear to me but I know that they differ from the definitions I chose in my mathematical model. Thus, I need to know the definition used in MuJoCo to build the appropriate transforms that will allow me to compare the aforementioned elements of the equations of motion. I have read the documentation and checked previous issues quite thoroughly and I found scattered pieces of information from which I can guess some of these definitions (I believe _most_ of them are in global coordinates) but I want some confirmation:

1 - From the documentation, I get that the qpos of this joint consists of the cartesian position of the body origin in global coordinates followed by the quaternion that represents its orientation. From what I read I assume that qvel, qacc and all the corresponding qfrc_ of this joint also present first the translational component and second the rotational one, can you confirm whether this is the case? 

2 - Is the **linear velocity** in qvel expressed in local or global coordinates? If this is expressed in global coordinates, is this the time derivative of the position or the spatial velocity (as defined by Featherstone)?

3 - Same for the **linear acceleration** in qacc, is it expressed in local or global coordinates? If in global coordinates is this the second time derivative of the position or the spatial acceleration?

4 - Are the **rotational velocity and acceleration** (in qvel and qacc respectively) expressed in local or in global coordinates?

5 - I inferred from some simulation data that the **forces** in all qfrc_ are in global coordinates, but are the **moments** in local coordinates?  

[...]

PR #1071, opened 2023-09-27
specify the frame of `qvel` of floating bodies in the doc

JewelryForge:
I surprisingly find that `qvel` of floating bodies is in the body frame instead of the world frame. It's different from that of another simulator (raisim). I think it may be better that the frame is specified in the doc


yuvaltassa:
See also [this comment](https://github.com/google-deepmind/mujoco/issues/691#issuecomment-1380347329)

Note to self: document this better.

PR #1154, opened 2023-11-01
[Unity plugin] Fix the missing .meta for copied library.

### LINKED ISSUE #1146: Missing a .meta file for the mujoco.dll in the unity plugin
I'm using MuJoCo and Unity on Windows 11. However, when I imported the package via git in the Unity editor, it promoted that the mujoco.dll does not have a .meta file and is ignored. 

After I create a .meta file for mujoco.dll manually, the plugin loads successfully. 

So, I think it is better to add .meta files for the runtime libraries.

elysia-best:
Fix #1146


yuvaltassa (approved):

yuvaltassa:
@Balint-H LGTY?

unity/Editor/Bindings/MujocoBinaryRetriever.cs
```diff
@@ -55,11 +57,38 @@ public class MujocoBinaryRetriever {
...
+      writer.WriteLine("fileFormatVersion: 2");
+      writer.WriteLine("guid: " + (uuidN.ToString()));
+      writer.WriteLine("DefaultImporter:");
+      writer.WriteLine("  externalObjects: {}");
+      writer.WriteLine("  userData: ");
+      writer.WriteLine("  assetBundleName: ");
+      writer.WriteLine("  assetBundleVariant: ");
+    }
+
+    Debug.Log("Meta file for " + file + " created!");
```
Balint-H, line 91: How about (or similar): "MuJoCo library imported to location " + filePath

unity/Editor/Bindings/MujocoBinaryRetriever.cs
```diff
@@ -55,11 +57,38 @@ public class MujocoBinaryRetriever {
...
                 mujocoPath + "\\mujoco.dll");
+            GenerateMetaFile(mujocoPath + "\\mujoco.dll");
             AssetDatabase.Refresh();
           }
         }
       }
     }
   }
+  
+  static void GenerateMetaFile(string file) {
```
Balint-H, line 68: Minor thing but consider filePath for the variable name.

Balint-H, line 68: Minor thing but consider filePath for the variable name.

Balint-H (commented):
Looks good, tested it works as intended, even as a URL referenced packages. I have some very minor suggestions regarding the message printed out after the library is imported, but otherwise its a good fix.

unity/Editor/Bindings/MujocoBinaryRetriever.cs
ganyuanzhen: OK, I've already updated the output and variable name.

ganyuanzhen:
Already add some minor changes to the prompts and change the vaiable name to `filePath`. @Balint-H

erez-tom:
Thanks for this!
Two alternatives to consider:
1. Call [Refresh](https://docs.unity3d.com/Manual/AssetDatabaseRefreshing.html) after copy.
2. Use [Unity's FileUtil](https://docs.unity3d.com/ScriptReference/FileUtil.CopyFileOrDirectory.html) instead of System.
If you have reasons to reject both, we can accept the PR.

Issue #1244, opened 2023-11-30
Support for 32 Index format for higher mesh size 


vidurvij-apptronik:
**Is your feature request related to a problem? Please describe.**
I am trying to import a mesh for my robot in Unity. It has high number of vertices, When I import the mesh directly in MuJoCo, it works but it seems like there is an issue with MuJoCo's unity plugin. It probably stems from the indexformat used to create new meshes in Unity. When I manually edit the StlParser.cs to remove the check for number of vertices and change the format for mesh to IndexFormat.UInt32, its working for MuJoCo version 3.0.1 but not for 2.3.7. I would love to have this supported natively,

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

vidurvij-apptronik:
https://github.com/google-deepmind/mujoco/pull/1254

vidurvij-apptronik:
Any progress on the pull request?

yuvaltassa:
We can't accept your pull request without the CLA.
Closing the issue, we can continue the discussion in the PR.

vidurvij-apptronik:
Can I repoen the discussion here! So I need to have a 32 bit index to store the meshes, so we can import higher resolution meshes. I am commenting here because I didn't get any reply there,

PR #1254, opened 2023-12-04
Changing indexFormat for the imported mesh from 16 bit to 32 bit.

vidurvij-apptronik:
I am changing the import format from 16 bit to 32 bit to support meshes with higher vertices count which is already supported in MuJoCo standalone.

https://github.com/google-deepmind/mujoco/issues/1244

### DIFF
--- a/unity/Editor/Importer/StlMeshParser.cs
+++ b/unity/Editor/Importer/StlMeshParser.cs
@@ -60,11 +60,13 @@ public class StlMeshParser {
         reader.ReadBytes(_headerLength);
         var numTriangles = reader.ReadUInt32();
         var numVertices = numTriangles * _verticesPerTriangle;
+        /*
         if (numVertices > _unityLimitNumVerticesPerMesh) {
           throw new IndexOutOfRangeException(
               "The mesh exceeds the number of vertices per mesh allowed by Unity. " +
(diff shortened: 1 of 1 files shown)


vidurvij-apptronik:
@yuvaltassa We can discuss the issue here?

erez-tom (approved):

Balint-H:
I think this is a reasonable addition, but I'd remove the commented out block that previously checked the vertex count. I believe a good follow-up PR would add the vertex count check back in, and use the 16 bit indexing on the mesh if it is sufficient.

vidurvij-apptronik:
@Balint-H The issue is that I cant sign the CLA yet. Is it possible to for you to make the change?

Balint-H:
Hmm, what do you see when you visit this link? https://cla.developers.google.com/clas

vidurvij-apptronik:
@Balint-H The latter one.

Balint-H:
Just an update: The PR from me will come later today or tomorrow. I managed to figure out a couple of related annoying issues first, which used to make `.stl`s generated by the plugin unopenable by Blender or other graphics software (making debugging problematic). More importantly, vertex indices will be reused now, significantly reducing the mesh asset size, and allowing the use of higher res meshes with the same amount of indices.

SebasNorenaR:
solved in #1353

Issue #1204, opened 2023-11-17
Compile failed with gcc 12.3.1 at release mode


limymy:
Mujoco source code compiling will pass at `Debug` mode but fail at `Release` mode. Here is the output:
```
[build] 在函数‘makeLabel’中,
[build]     内联自‘mjv_addGeoms’于 /home/li/bitbot-mujoco-demo/build/_deps/mujoco-src/src/engine/engine_vis_visualize.c:636:11:
[build] /home/li/bitbot-mujoco-demo/build/_deps/mujoco-src/src/engine/engine_vis_visualize.c:64:3: 错误：‘strncpy’ output may be truncated copying 99 bytes from a string of length 99 [-Werror=stringop-truncation]
[build]    64 |   strncpy(label, txt, 99);
[build]       |   ^~~~~~~~~~~~~~~~~~~~~~~
[build] 在函数‘makeLabel’中,
[build]     内联自‘mjv_addGeoms’于 /home/li/bitbot-mujoco-demo/build/_deps/mujoco-src/src/engine/engine_vis_visualize.c:676:11:
[build] /home/li/bitbot-mujoco-demo/build/_deps/mujoco-src/src/engine/engine_vis_visualize.c:64:3: 错误：‘strncpy’ output may be truncated copying 99 bytes from a string of length 99 [-Werror=stringop-truncation]
[build]    64 |   strncpy(label, txt, 99);
[build]       |   ^~~~~~~~~~~~~~~~~~~~~~~
......
[build] 在函数‘bufread’中,
[build]     内联自‘mj_loadModel’于 /home/li/bitbot-mujoco-demo/build/_deps/mujoco-src/src/engine/engine_io.c:778:3:
[build] /home/li/bitbot-mujoco-demo/build/_deps/mujoco-src/src/engine/engine_io.c:377:3: 错误：‘memcpy’ offset [0, 299] is out of the bounds [0, 0] [-Werror=array-bounds]
[build]   377 |   memcpy(dest, (char*)buf + *ptrbuf, num);
[build]       |   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
There are two errors. 
- `strncpy` function at
https://github.com/google-deepmind/mujoco/blob/4d2a176a8afecfc1b31b87aeb015602cb2772769/src/engine/engine_vis_visualize.c#L64
It can be fixed in two method. 
  1. Change it to `strncpy(label, txt, 100) ;` to fit
[...]

PR #1263, opened 2023-12-07
fix gcc warning "stringop-truncation"

limymy:
fix part of #1204

### DIFF
--- a/src/engine/engine_vis_visualize.c
+++ b/src/engine/engine_vis_visualize.c
@@ -62,7 +62,7 @@ static void makeLabel(const mjModel* m, mjtObj type, int id, char* label) {
   }
 
   // copy result into label
-  strncpy(label, txt, 99);
+  strncpy(label, txt, 100);
   label[99] = '\0';
 }
 


yuvaltassa:
Thanks!

Better to use `mjSTRNCPY` from `engine_array_safety.h` (already included).

src/engine/engine_vis_visualize.c
```diff
@@ -62,7 +62,7 @@ static void makeLabel(const mjModel* m, mjtObj type, int id, char* label) {
   }
 
   // copy result into label
-  strncpy(label, txt, 99);
+  strncpy(label, txt, 100);
```
yuvaltassa, line 65: Use `mjSTRNCPY` and then you can delete the following line.

limymy: Can't use `mjSTRNCPY` because the type of `label` is `char*` which would cause compiler error.
https://github.com/google-deepmind/mujoco/blob/1885d518ebf80b6fe6c2ba595be06027ce7306da/src/engine/engine_array_safety.h#L25-L27

yuvaltassa: Right, sorry.

yuvaltassa (approved):

PR #1268, opened 2023-12-07
[Unity plugin] Dynamic heightfield syncing

Balint-H:
In this PR the HField component optionally requests a rebuild of the scene during runtime if the Unity terrain changes (with optional rate limit), to allow dynamic deformable environments, (e.g. when resetting an episode of locomotion learning, or to simulate destructible terrain).

I had a go at trying to use `model->hfield_data` by simply writing changed heightmap pixel values to it (as floats), making sure its at the right address and right number of elements and that scene is properly initialized already. I expected that to be a cleaner and more performant approach, however this was met with crashes. Are there any limitations of writing to `hfield_data` I'm not aware of that might prevent its use with plugin this way?

This PR also provides edits to the HField geom's components to be a bit more robust to different terrains (taking into account assumptions of a MuJoCo hfield), and provides warnings and tooltips for configuring a terrain so that it matches with the hfield on the MuJoCo side. 

https://github.com/google-deepmind/mujoco/assets/41113387/5b675f09-a030-47ec-a369-57e6034f8de6


erez-tom (changes requested):
Thanks so much!  Could you please remove whitespaces at end of lines?

Balint-H:
Thanks for spotting that, I believe that should be all the whitespace issues sorted.

Balint-H:
Thanks for the patience, I made an effort now to familiarise myself with formatting tools, and will try to be more consistent from now on.

yuvaltassa:
Writing to hfield_data is supposed to work fine. We should try to understand this. 

Can you try and make a MWE (in a new issue) which recreates the crash? (Ideally via Python or C, not Unity)

Thanks!

erez-tom (approved):

Balint-H:
Thanks for encouraging a further examination Yuval. I tried to recreate a MWE in python and everything worked as I expected with dynamic hfields (although slower than real time with the interactive viewer). 

I went back to Unity and tried with the latest version of the plugin and binary (3.1.1) and the heightfield works as intended! My original try was with 3.0.1, so the issue was possibly related to that. Running the same scripts when I reverted back to 3.0.1 gave the same crashes.

In any case, the `hfield_data` version runs considerably smoother on 3.1.x, so I'll be sending a PR in the coming days with that.

Balint-H:
With a quick profiling only in editor, writing to hfield_data is approximately 5-10 times faster then recompiling the scene. Now this approach will be used if no export path for the hfield png is given for the component (the default), but I haven't removed the scene rebuilding and image exporting methods as I see them being useful in case someone is building a hfield scene in Unity and exports it for use with other MuJoCo sim options.

Exporting scenes with the editor window tool works, and multi hfield scenes work too.

erez-tom (approved):

PR #1275, opened 2023-12-13
Use positional arguments with jp.where in MJX tutorial

TyLindberg:
Due to this PR (https://github.com/google/jax/pull/18526), the MJX tutorial will now print a deprecation warning for the usage of the keyword arguments `x` and `y` with `jp.where`. This PR changes the keyword arguments to positional arguments, avoiding a deprecation error in the future and a potential confusion point for users new to MJX.


erikfrey (approved):
Thank you!

PR #1281, opened 2023-12-16
Explicitly convert arrays to scalars in Python tutorial notebook

TyLindberg:
When running the Python tutorial notebook, the following deprecation warning is printed for certain code blocks.

```
DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)
```

There are two lines of code that cause this warning.

`angle[i] = data.joint('root').qpos`: `data.joint('root').qpos` is a single element array, which is implicitly converted to a scalar.
`warning = mujoco.mjtWarning(warning_index).name`: `warning_index` is a single element array, which is implicitly converted to a scalar.

This PR adjusts those sections of code to explicitly index into the array and avoid the implicit conversion numpy is performing.


btaba (approved):
LGTM

yuvaltassa (approved):

PR #1345, opened 2024-01-18
Adding support to change the default material in MuJoCo according to …

vidurvij-apptronik:
…the rende pipeline being used.


Balint-H:
Hello! Thanks for the contribution. Have you tested it with projects of the built-in and SRP renderers?

Balint-H:
There are also some small editing nitpicks (indent size in lines 32-46). If you can make these adjustments and test it with the other pipelines will merge it, otherwise we can test it in the coming days.

yuvaltassa:
closing due to lack of activity

PR #1353, opened 2024-01-20
[Unity plugin] Reuse vertex indices when importing STL meshes

### LINKED ISSUE #1244: Support for 32 Index format for higher mesh size

Balint-H:
This PR makes modifications to the STL IO of the Unity plugin. Changes:
- Writes only the new bytes of the buffer to the file, which avoids having trailing nulls in the imported and scaled STL. Blender and other STL viewing software treated the files with trailing nulls as corrupted.
- Using a dictionary, indices of shared vertices are reused during the import. This allows loading larger meshes with the same index resolution, and the resulting files are smaller and render faster (~20%-40% smaller asset files in the MPL hand for example).
- Switch to 32bit indexing if 16 bits are not sufficient even with index reusing.

Tested with the MuJoCo mug, and the MPL hand.

It was unclear to me why was it necessary to perform axis swizzling in the STL writer part as well, and not just the parser. As I was working on this PR, meshes imported 90 degrees rotated sometimes. Removing the swizzling from there resolved that. In case I missed a case where that was key for the proper import, let me know @erez-tom.

EDIT: I re-reviewed my test cases and noticed I accidentally added an extra axis swizzling to the source mesh files when I decimated/subdivided the meshes to test with different vertex counts. Checking with the original mesh files revealed that swizzling on the writer too is necessary, so I added that back to the PR quickly.

@vidurvij-apptronik

Fixes #1244, replaces #1254

### DIFF
--- a/unity/Editor/Importer/StlMeshParser.cs
+++ b/unity/Editor/Importer/StlMeshParser.cs
@@ -12,11 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using UnityEngine;
+using UnityEngine.Rendering;
 
 namespace Mujoco {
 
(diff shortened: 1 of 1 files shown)


vidurvij-apptronik:
The decision to use UInt32 only when the vertex count exceed the limit is for memory optimizing reasons?

Balint-H:
Yeah, using 16 bit will result in smaller assets, and faster loading scenes (the impact would only be noticeable on mesh heavy scenes). With the index reusing implemented, more meshes will be able to use 16 bit. Please let us know if the new implementation works for you! 

Since the 3.1.2 dll is not released yet, and this PR is for the upcoming 3.1.2 compatible version of the plugin, if you want to try it now you'd need to backport the parser to your fork (just copy and paste the stl parser from this PR)

vidurvij-apptronik:
@Balint-H I tested this with my use case and it works perfectly.

erez-tom (approved):

PR #1413, opened 2024-02-12
USD Integration

abhihjoshi:
Three files:
- mujoco/python/mujoco/usd_exporter.py
- mujoco/python/mujoco/usd_component.py
- mujoco/python/mujoco/usd_utils.py

The USD exporter class handles all the overall USD file and materials. The USD component files handles all the objects in the scene.


python/dist/mujoco-3.0.2.tar.gz
nimrod-gileadi, line 1: Remove from this PR.

nimrod-gileadi (changes requested):
First look.

yuvaltassa:
Let's delete the `usd_` prefix to avoid stuttering. `usd/component.py` etc.

python/mujoco/__init__.py
```diff
@@ -46,6 +46,9 @@
 from mujoco._structs import *
 from mujoco.gl_context import *
 from mujoco.renderer import Renderer
+from mujoco.usd.usd_exporter import USDExporter
```
nimrod-gileadi, line 49: Remove these imports so the module can be loaded separately.

nimrod-gileadi (approved):

crewmatt (approved):

crewmatt (approved):

yuvaltassa:
This needs to be rebased.
Probably better to just close this PR and start a new one at head? 
The discussion is long and mostly irrelevant...

PR #1438, opened 2024-02-23
[Unity plugin] Support multielement springlength in tendons

Balint-H:
A quick small fix for this missing feature.


yuvaltassa:
This LGTM but I'll let @erez-tom review.

More broadly, I propose the following easy test: Load all the models under `mujoco/test`, complain if any don't load. Would this or a similar test catch future bugs where a feature is added to MJ but is missing from the plugin?

SebasNorenaR:
Closed as multielement springlength has been merged in https://github.com/google-deepmind/mujoco/commit/92e2ca9c502d67a7a6f1e422a2b8faee9073383c

PR #1447, opened 2024-02-25
Parallel factor_m

jaro-sevcik:
Just for illustration.

This patch tries to update the rows in the same model tree depth in parallel. Unfortunately, this ends up being mostly slower that the existing code updating the rows in sequence (with scatter). The parallel approach only (slightly) wins on smaller models where there is not enough parallelism.

### DIFF
--- a/mjx/mujoco/mjx/_src/smooth.py
+++ b/mjx/mujoco/mjx/_src/smooth.py
@@ -228,7 +228,6 @@ def factor_m(m: Model, d: Data) -> Data:
     return d
 
   # build up indices for where we will do backwards updates over qLD
-  dof_madr = jp.array(m.dof_Madr)
   depth = []
   for i in range(m.nv):
     depth.append(depth[m.dof_parentid[i]] + 1 if m.dof_parentid[i] != -1 else 0)
@@ -248,7 +247,7 @@ def factor_m(m: Model, d: Data) -> Data:
   qld = d.qM
 
   for _, updates in sorted(updates.items(), reverse=True):
-    # combine the updatess into one update batch (per depth level)
+    # combine the updates into one update batch (per depth level)
     rows = []
     madr_ijs = []
     pivots = []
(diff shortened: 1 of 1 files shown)


mjx/mujoco/mjx/_src/smooth.py
```diff
@@ -228,8 +228,10 @@ def factor_m(m: Model, d: Data) -> Data:
     return d
 
   # build up indices for where we will do backwards updates over qLD
-  # TODO(erikfrey): do fewer updates by combining non-overlapping ranges
   dof_madr = jp.array(m.dof_Madr)
```
erikfrey, line 231: might as well kill this temporary now and switch the other references to m.dof_Madr

mjx/mujoco/mjx/_src/smooth.py
```diff
@@ -240,21 +242,31 @@ def factor_m(m: Model, d: Data) -> Data:
...
-
-    @jax.vmap
-    def off_diag_fn(madr_d, madr_ij, qld=qld, width=out_end - out_beg):
-      qld_row = jax.lax.dynamic_slice(qld, (madr_ij,), (width,))
-      return -(qld_row[0] / qld[madr_d]) * qld_row
-
-    qld_update = jp.sum(off_diag_fn(madr_d, madr_ij), axis=0)
-    qld = qld.at[out_beg:out_end].add(qld_update)
+  for _, updates in sorted(updates.items(), reverse=True):
+    # combine the updatess into one update batch (per depth level)
```
erikfrey, line 251: s/updatess/updates

mjx/mujoco/mjx/_src/smooth.py
```diff
@@ -240,21 +242,31 @@ def factor_m(m: Model, d: Data) -> Data:
...
+    pivots = []
+    out = []
+
+    for (b, e, madr_d, madr_ij) in updates:
+      width = e - b
+      rows.append(np.arange(madr_ij, madr_ij + width))
+      madr_ijs.append(np.full((width,), madr_ij))
+      pivots.append(np.full((width,), madr_d))
+      out.append(np.arange(b, e))
+    rows = jp.array(np.concatenate(rows))
```
erikfrey, line 263: seems OK to remove the jp.array(...) wrapper for this and the next few lines?

erikfrey (changes requested):
Looks really nice, thank you!  Just a few nits.

erikfrey (approved):

PR #1449, opened 2024-02-25
Replace top-k with first-k thresholding

jaro-sevcik:
Just for sharing a prototype, not for landing.


jaro-sevcik:
I think @btaba had some prettier (and properly tested) version of this, it would be good to land that rather than this sketch.

PR #1452, opened 2024-02-26
Integrating VR into simulate

nishbo:
The current branch uses OpenVR (very close to the original mjvive project) to render the scenes in VR. It is turned on via MUJOCO_BUILD_SIMULATE_VR cmake option. If the VR could not be initialized, the simulate runs normally. I aimed at minimizing the number of changes/additions to simulate.cc. Supports definition of VR transformation in the model (instead of hardcoded default) - see balloons model.

Tested on Windows 10/11, using Steam VR:
Valve Index - works
HTC Vive XR Elite - works, streaming using Vive Streaming Hub
Meta Quest 3 - technically works (through Steam Link), but the projections are wrong, will need fixing.

Current problems:
1. Not enough documentation. Suggestions on where to add docs are welcome.
2. I need to manually add openvr_api.lib to the linker include directories in Visual Studio and copy openvr.dll into the output directory. Probably can be fixed at the cmake level, not sure how, suggestions welcome.
3. Local copy of glad.

Directions before merging:
1. Switching between VR transformations (partially complete).
2. Fixing Quest 3.
4. Adding UI bars to VR.
5. Adding controllers.


simulate/cmake/SimulateDependencies.cmake
```diff
@@ -33,7 +33,20 @@ set(MUJOCO_DEP_VERSION_glfw3
     7482de6071d21db77a7236155da44c172a7f6c9e # 3.3.8
     CACHE STRING "Version of `glfw` to be fetched."
 )
+
+# set(MUJOCO_DEP_VERSION_openxr
```
nimrod-gileadi, line 37: Remove if it's not used.

simulate/simulate.cc
```diff
@@ -1804,6 +1804,12 @@ Simulate::Simulate(std::unique_ptr<PlatformUIAdapter> platform_ui,
       uistate(this->platform_ui->state()) {
   mjv_defaultScene(&scn);
   mjv_defaultSceneState(&scnstate_);
+
+#ifdef mjBUILDSIMULATEVR
+  hmd.initHmd();
+  hmd.initTextures(this->scn);
+  hmd.transform(this->scn, this->m_);
+#endif
```
nimrod-gileadi, line 1812: Add the  `  // mjBUILDSIMULATEVR` comment.

simulate/simulate_vr/KHR/khrplatform.h
nimrod-gileadi, line 1: This isn't auto-generated, right? Could this be taken from the checked out repository, instead of checking in our own copy?

simulate/simulate_vr/simulate_vr.cc
```diff
@@ -0,0 +1,330 @@
...
+#include "simulate_vr.h"
+
+#include "glad\glad.h"
+#include <GLFW/glfw3.h>
+
+#pragma comment(lib, "openvr_api.lib")
+
+using namespace vr;
+
+// initialization funcitons
```
nimrod-gileadi, line 24: typo

CMakeLists.txt
```diff
@@ -98,6 +99,10 @@ if(MUJOCO_ENABLE_AVX_INTRINSICS)
   target_compile_definitions(mujoco PUBLIC mjUSEPLATFORMSIMD)
 endif()
 
+if (MUJOCO_BUILD_SIMULATE_VR)
+  target_compile_definitions(mujoco PUBLIC mjBUILDSIMULATEVR)
```
nimrod-gileadi, line 103: Does this affect MuJoCo itself, or only simulate? If it's the latter, than the compiler definition should go there instead.

nishbo:
For anyone who stumbles upon this PR, the OpenVR-based development is frozen, however there is a working Windows-OpenXR [branch](https://github.com/nishbo/mujoco/tree/open-xr-vr-base) and [PR draft](https://github.com/google-deepmind/mujoco/pull/2124).

PR #1477, opened 2024-03-05
Add CMake options to enable or disable installation of mujoco samples and simulate

traversaro:
The current logic uses regular CMake variables, that create confusing behavior when a CMake cache variable with the same name is set from the command line. The PR does not change the behaviour of the build system, but permits to users to enable installation of simulate and samples executables from the CMake options.

### DIFF
--- a/sample/CMakeLists.txt
+++ b/sample/CMakeLists.txt
@@ -118,7 +118,9 @@ if(APPLE AND MUJOCO_BUILD_MACOS_FRAMEWORKS)
   set(_INSTALL_SAMPLES OFF)
 endif()
 
-if(_INSTALL_SAMPLES)
+option(MUJOCO_SAMPLES_INSTALL "If ON, also install samples executables." ${_INSTALL_SAMPLES})
+
+if(MUJOCO_SAMPLES_INSTALL)
 
   include(TargetAddRpath)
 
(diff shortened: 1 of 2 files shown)


traversaro:
Unfortunatly, this can't work if we do not also duplicate the `TargetAddRpath.cmake` module in https://github.com/google-deepmind/mujoco/tree/7e87c6ac44ef980dbc5bb16f45100cd8b63c427b/sample/cmake and https://github.com/google-deepmind/mujoco/tree/7e87c6ac44ef980dbc5bb16f45100cd8b63c427b/simulate/cmake .

traversaro:
> Unfortunatly, this can't work if we do not also duplicate the `TargetAddRpath.cmake` module in https://github.com/google-deepmind/mujoco/tree/7e87c6ac44ef980dbc5bb16f45100cd8b63c427b/sample/cmake and https://github.com/google-deepmind/mujoco/tree/7e87c6ac44ef980dbc5bb16f45100cd8b63c427b/simulate/cmake .

Actually, we can make it work without duplication when `MUJOCO_ENABLE_RPATH` is OFF.

traversaro:
> > Unfortunatly, this can't work if we do not also duplicate the `TargetAddRpath.cmake` module in https://github.com/google-deepmind/mujoco/tree/7e87c6ac44ef980dbc5bb16f45100cd8b63c427b/sample/cmake and https://github.com/google-deepmind/mujoco/tree/7e87c6ac44ef980dbc5bb16f45100cd8b63c427b/simulate/cmake .
> 
> Actually, we can make it work without duplication when `MUJOCO_ENABLE_RPATH` is OFF.

Fixed this in https://github.com/google-deepmind/mujoco/pull/1478, not sure why I can't reopen this PR.

PR #1487, opened 2024-03-08
Add test to mj_loadXML

alefram:
Hi, I added a test for the mj_loadXML method."

### DIFF
--- a/test/xml/xml_api_test.cc
+++ b/test/xml/xml_api_test.cc
@@ -64,6 +64,7 @@ TEST_F(LoadXmlTest, InvalidXmlFailsToLoad) {
     mj_deleteModel(model);
   }
 }
+
 TEST_F(LoadXmlTest, MultipleBodies) {
   static constexpr char xml[] = R"(
   <mujoco>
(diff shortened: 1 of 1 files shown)


yuvaltassa:
Thanks! Requesting review from @nimrod-gileadi as he wrote the TODO in the first place 🙂

test/xml/xml_api_test.cc
```diff
@@ -64,8 +64,38 @@ TEST_F(LoadXmlTest, InvalidXmlFailsToLoad) {
...
+  EXPECT_EQ(model->nbody, 3);
+    
+  mjData* data = mj_makeData(model);
+  EXPECT_THAT(data, NotNull());
+  mj_step(model, data);
+  mj_deleteData(data);
+  mj_deleteModel(model);
+}
+
 // TODO(nimrod): Add more tests for mj_loadXML.
```
nimrod-gileadi, line 96: Do you mind deleting this line?

This TODO was written at the point where we literally only had one test in all of MuJoCo.
mj_loadXML is now very well covered by many tests.

This new test won't hurt, but I don't think we need more basic tests of this sort.

nimrod-gileadi (changes requested):

test/xml/xml_api_test.cc
alefram: Oh, I understand. I will remove that.

nimrod-gileadi (approved):

PR #1504, opened 2024-03-13
[Unity-plugin] Context menu for creating Unity terrain components

Balint-H:
A QoL addition to create and align Unity terrain with hfield geoms. Also includes tweaks to the way the MJCF is constructed (previously the local Y position of the terrain was used as the basehieght, now it can be separately defined in the HField property of the geom).


unity/Runtime/Components/Shapes/MjHeightFieldShape.cs
```diff
@@ -75,17 +80,17 @@ public class MjHeightFieldShape : IMjShape {
     var assetName = scene.GenerationContext.AddHeightFieldAsset(this);
 
     if (Application.isPlaying) {
-      scene.postInitEvent += (unused_first, unused_second) =>
```
erez-tom, line 78: I had to add these names for our internal tools, please leave them as is.

Balint-H: Ah, I thought it was my editor applying style, I'll swap it back.

erez-tom (approved):

erez-tom (approved):

kezzyhko:
Fixes #1930

Balint-H:
@erez-tom Once you are available again could you please merge this in?

erez-tom (approved):

yuvaltassa (approved):

erez-tom (approved):

Balint-H:
I had another look to see what could potentially cause this to not get merged. I have found a couple lines beyond 100 characters, which I think is the limit. I adjusted the formatting now to be under that, hopefully this will enable merging in with your approval @erez-tom.

yuvaltassa:
@Balint-H @erez-tom can I close or do we want this

Balint-H:
Yeah, this is much better for user experience when working with heightfields. I'll test if it still works as intended with the newer mujoco/plugin version after a merge, and report back!

Balint-H:
@yuvaltassa Thanks for the patience, I checked, and it works as intended when merged into main.

yuvaltassa:
Hi @Balint-H, thanks for testing this with `main`! Before we merge, we took a detailed look at the changes and had a few technical questions and observations:

### 1. Event Subscription & Lifecycle in `MjGeomEditor`
In `MjGeomEditor`, `EditorApplication.contextualPropertyMenu` is subscribed in `OnEnable()` and unsubscribed in `OnDestroy()`. In Unity, custom `Editor` instances are disabled via `OnDisable()` when the selection changes in the Hierarchy; they are not destroyed immediately.
- Unsubscribing in `OnDestroy()` rather than `OnDisable()` might leave dangling callbacks on `EditorApplication.contextualPropertyMenu` that fire against stale `target` references when inspecting other objects. Should this be moved to `OnDisable()` (calling `base.OnDisable()`)?
- Given that `MjGeomEditor` already uses `[MenuItem("CONTEXT/MjGeom/...")]` for other actions (like adding mesh renderers), would a standard component context menu item (`[MenuItem("CONTEXT/MjGeom/Add Unity Terrain")]`) be simpler and avoid global event hooking altogether?
 
### 2. `TerrainData` Persistence & Asset Lifecycle
In `MjHeightFieldShape.AddTerrain()`, a new `TerrainData` instance is created via `new TerrainData()`. In Unity, `TerrainData` is an external asset (`UnityEngine.Object`) that cannot be embedded directly into `.unity` scene files.
- Without saving the asset to disk (e.g. via `AssetDatabase.CreateAsset(...)` or standard `Terrain.CreateTerrainGameObject(terrainData)`), will `newTerrain.terrainData` become `None (Terrain Data)` / missing once the user saves, closes, and reopens the scene?
 
### 3. Parent Resolution & Editor Assembly Placement in `AddTerrain()`
To find the parent transform, `AddTerrain()` currently performs a global search across all objects:
- If no matching geom is found (e.g. on a prefab asset or temporary instance), `.First(...)` throws an unhandled `InvalidOperationException`.
- Since the caller in `MjGeomEditor` already holds the target `MjGeom` reference, can the parent `Transform` simply be passed into `AddTerrain(Transform parent)` directly to eliminate the global scan?
- Also, `MjHeightFieldShape.cs` is in the `Runtime/` assembly, but `AddTerrain()` contains editor scene authoring logic. Should `AddTerrain()` either be moved to `Editor/` or guarded with `#if UNITY_EDITOR` to avoid compiling editor logic into standalone player builds? Could `Undo.RegisterCreatedObjectUndo` / `EditorUtility.SetDirty` also be added so creating a terrain can be undone with Ctrl+Z?

### 4. Render Pipeline Compatibility
In `MjHeightFieldShape.AddTerrain()`:
```csharp
newTerrain.materialTemplate = new Material(Shader.Find("Nature/Terrain/Diffuse"));
```
Since PR #3202 updated the plugin to URP, `Nature/Terrain/Diffuse` (the legacy built-in pipeline shader) either does not exist or fails to render in URP/HDRP projects. Could this assignment be removed or delegated to `Terrain.CreateTerrainGameObject` so Unity automatically assigns the active pipeline's default terrain material?

### 5. Shadowed & Unused Properties
In `MjHeightFieldShape.cs`:
- `public float MinimumHeight { get; private set; }` is shadowed on line 115 by `var minimumHeight = ...`, leaving the public property permanently `0.0f`.
- `public float MaximumHeight { get; private set; }` is assigned, but is no longer read anywhere since `MjcfGenerationContext` now references `HeightMapScale.y`. Should these be cleaned up?
 
### 6. Height Scaling Consistency
In `MjcfGenerationContext`, `heightRange` is now set to `hFieldComponent.HeightMapScale.y`.
- In `MjHeightFieldShape.UpdateHeightFieldData`, `curData` is computed with `.Select(c => c.r * 2)`. Because MuJoCo computes elevation as data × size[2], does scaling runtime data by 2 while setting size[2] = `HeightMapScale.y` produce a terrain twice as tall when updated dynamically compared to when exported as a normalized [0, 1] PNG?

Balint-H:
1. Got rid of subscription managing completely which was previously done to avoid multiple subscriptions from multiple instances. Now the event is subscribed guaranteed once in a static constructor. The `MenuItem` approach would be inelegant as it would expose the option for all geom types, not just hfields.
2. This is fine, it creates and references an asset. The terrain asset is handled like any other persistent asset.
3. Passing the transform is a bit neater I agree. Implemented.
4. I was unaware of that function, it also lets us skip managing the collider, and the help with the material is also nice.
5. Missed this, cleaned.
6. The height is consistent with the simulation in either approach.

Issue #723, opened 2023-02-06
XML Error: unknown plugin 'mujoco.elasticity.cable' when running simulate


JohnChen1024a:
When I loaded the model "belt.xml" in the directory of model/plugin to "simulate" program of MuJoCo, I got the following error message: 
XML Error: unknown plugin 'mujoco.elasticity.cable'
Element 'required', line 20
Any suggestion for fixing this error is welcome.
My environment: Ubuntu 20.04, MuJoCo version 2.3.1

saran-t:
How did you obtain/install MuJoCo on your machine?

JohnChen1024a:
I downloaded source code from github of MuJoCo, and built it using cmake. (make build;cd build;cmake ..;make -j 8) Under bin, run "simulate" and drag "belt.xml" to it.

JohnChen1024a:
@saran-t Thank you for your help. It works. I suggest this information of searching for plugin directory should be added to the "build from source" section.

yuvaltassa:
@JohnChen1024a, would you like to submit a PR to update the docs as you propose?

tadashiK:
Hello, I encountered the same problem when I installed mujoco via pip or conda. Would you tell me how to fix the issue in this case too?

As for a PR, I would do including a solution fo this too.

To repicate the issue, please use the following colab: https://colab.research.google.com/drive/1tOP38GGEe6xyzKkhhBOgudmn69PDck87?usp=sharing

DMackRus:
Hello, 

I am having similar issues to the ones in this post. I have a custom MuJoCo fork (it's a tiny change and I was hoping it could be added to MuJoCo directly, please see this request issue (https://github.com/google-deepmind/mujoco/issues/1453)). I just wanted public access to one function `mj_stepSkip()` for my own uses.

When i follow the instructions of building from source, I do not have the first party plugins? Currently I have a working hack where I build my custom fork from source, download the latest release and copy the plugins over. This seems fine at the moment I just dont know if it might break down the line.

saran-t:
@DMackRus the plugins are actually built when you run `cmake --build`, the problem is that `cmake --install` isn't copying them over.

This is an oversight in our installation setup, but in the meantime you should be able to find these plugins in the `lib` directory of your build tree.

DMackRus:
@saran-t 

Thank you for your reply, yes that makes sense and I can just add a quick command to my shell script to copy the plugins folder over then to automate my issue, thanks!

[...]

PR #1515, opened 2024-03-15
Add installation of MuJoCo plugins as part of CMake installation logic

traversaro:
Some users in my lab found confusing the fact that `simulate` did not found the plugins out of the box if mujoco was built from source, and apparently also other users had the same confusion (see for example https://github.com/google-deepmind/mujoco/issues/723#issuecomment-1893594343). 

This PR hopefully simplifies this by also installing as part of the CMake install step the mujoco plugins in `${CMAKE_INSTALL_PREFIX}/bin/mujoco_plugin`, exactly where `simulate` expects to find them, and where they are placed in the official release binaries.

At the moment, I did not include this modifications for the case in which `MUJOCO_BUILD_MACOS_FRAMEWORKS` is `ON`, as it is not really clear to me how this should be handled. However, when building by source `MUJOCO_BUILD_MACOS_FRAMEWORKS` is set to `OFF`, so for all users that build from source, this PR ensure that the installed `simulate` will be able to find all the mujoco plugins out of the box.

fyi @francesco-romano

### DIFF
--- a/cmake/MujocoOptions.cmake
+++ b/cmake/MujocoOptions.cmake
@@ -79,6 +79,30 @@ else()
 endif()
 
 option(MUJOCO_BUILD_MACOS_FRAMEWORKS "Build libraries as macOS Frameworks" OFF)
+option(MUJOCO_INSTALL_PLUGINS "Install MuJoCo plugins" ON)
+mark_as_advanced(MUJOCO_INSTALL_PLUGINS)
+
+# Define function to simplify installations of plugins
+function(mujoco_install_plugin plugin_target_name)
+  # Do not install plugin on macOS when using macOS Frameworks
+  if(MUJOCO_INSTALL_PLUGINS AND NOT MUJOCO_BUILD_MACOS_FRAMEWORKS)
+    if(WIN32)
+      install(TARGETS ${plugin_target_name}
+              # consistently with the official mujoco binary format, we do
+              # not want to install .lib import libraries on Windows, and
+              # only install .dll . To do so, we install non-.dll files to
+              # a fake install directory that is not actually in the install path
+              ARCHIVE DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/fake_plugin_install"
+              LIBRARY DESTINATION "${CMAKE_CURRENT_BINARY_DIR}/fake_plugin_install"
+              RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}/mujoco_plugin")
+    else()
+      install(TARGETS ${plugin_target_name}
+              ARCHIVE DESTINATION "${CMAKE_INSTALL_BINDIR}/mujoco_plugin"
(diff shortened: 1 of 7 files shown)


nimrod-gileadi (approved):

cmake/MujocoOptions.cmake
```diff
@@ -79,6 +79,30 @@ else()
 endif()
 
 option(MUJOCO_BUILD_MACOS_FRAMEWORKS "Build libraries as macOS Frameworks" OFF)
+option(MUJOCO_INSTALL_PLUGINS "Install MuJoCo plugins" ON)
```
nimrod-gileadi, line 82: Do you mind making exactly the same changes to SimulateOptions.cmake?

SimulateOptions.cmake is used for SIMULATE_STANDALONE here: https://github.com/google-deepmind/mujoco/blob/91192393a0293d97bbb4bdd0fe7a16d8faf43bc0/simulate/CMakeLists.txt#L60

and the two files should be kept in sync.

traversaro: Done and rebased. 

Just an OT curiosity, if the two files must be kept in sync, there is any reason why we can't use the same `MujocoOptions.cmake` directly in `mujoco/simulate/CMakeLists.txt`, for example using `include(../cmake/MujocoOptions.cmake)` instead of `include(SimulateOptions)` ? Indeed inclusion of a file in a sibling directory is a bit ugly, but perhaps it would be easier to manage then two files that need to be manually kept in sync (unless you have something on the Google side to keep them in sync, in that case probably the problem is just for external contributors).

yuvaltassa: Internally they are just one file that gets split in our CI so ya, this is a problem we save specifically for external contributors 🙂

yuvaltassa (approved):

cmake/MujocoOptions.cmake
yuvaltassa: Annoyingly it looks like copybara still refuses to ingest this 😠 

@nimrod-gileadi should I just copy-paste this into a CL?
Alternatively we could implement @traversaro's proposal to do the file duplication/referencing in cmake rather than copybara?

I'll approve yet again, maybe it'll work... 🤷‍♂️

yuvaltassa (approved):

saran-t:
I'm back at work now. Will see if we can finally get this merged.

saikishor:
Hello!

What's the status of this PR?

traversaro:
> Hello!
> 
> What's the status of this PR?

The problem was the copybara propagation of the changes. As one year and half passed, perhaps just opening the same PR will work out of the box, if you want to do that and check if that works feel free to do it, thanks!

saikishor:
Thanks

yuvaltassa:
Hi @traversaro, closing this pull request due to staleness and earlier Copybara synchronization conflicts as discussed above. Anyone needing this plugin installation logic in CMake is welcome to submit a fresh PR rebased on current `main`.

PR #1646, opened 2024-05-10
[Unity plugin] Remove sensor noise properties

Balint-H:
Remove references in MjGlobalOptions and sensor base class to the deprecated sensornoise attribute.

### DIFF
--- a/unity/Runtime/Components/MjGlobalSettings.cs
+++ b/unity/Runtime/Components/MjGlobalSettings.cs
@@ -73,7 +73,6 @@ public struct MjcfOptionFlag {
   public EnableDisableFlag Override;
   public EnableDisableFlag Energy;
   public EnableDisableFlag FwdInv;
-  public EnableDisableFlag SensorNoise;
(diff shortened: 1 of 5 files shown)


yuvaltassa (commented):
Removing the enable flag: Great!

Removing the sensor attribute: Why? It's still there for the user to write into, if they like... Maybe just update the tooltip?

Balint-H:
I think it could be easily misleading, e.g. they have learned some RL controller but want to make it more robust with sensor noise, they see the option and just put a value there without it actually having the intended effect. In any component they use to collect sensor readings with, they can inject noise there just as well.

erez-tom (approved):

erez-tom:
Could you please run the tests?  Both MjBaseSensorTests and MjcfImporterTests make references to this now-absent field.

Balint-H:
Somehow I missed the test framework implementation in the package (by default it is not compiled)! From now on I'll make sure to run them before a PR, and I'll amend this one.

erez-tom (approved):

PR #1714, opened 2024-06-03
Tendons in USD

abhihjoshi:
- Adding support for tendons in USD
- Updating geom visibility logic for changing visibility across multiple frames
- Adding user friendly demo script for generating USD file for inputted XML file


python/mujoco/usd/shapes.py
```diff
@@ -0,0 +1,183 @@
...
+            radius=mesh_config[shape]["radius"],
+            height=mesh_config[shape]["height"],
+            resolution=resolution,
+            create_uv_map=True,
+        )
+
+      if "transform" in config:
+        if "rotate" in config["transform"]:
+          tmp = o3d.geometry.TriangleMesh.create_coordinate_frame() # TODO: add rotation matrix code gen to utils.py
+          R = tmp.get_rotation_matrix_from_xyz(config["transform"]["rotate"])
```
yuvaltassa, line 163: you could use `mujoco.mju_euler2Quat` and `mujoco.mju_quat2Mat`  here

saran-t (approved):

python/mujoco/usd/exporter.py
```diff
@@ -413,19 +401,47 @@ def add_camera(
...
   def save_scene(self, filetype: str = "usd"):
+    assert filetype in ["usd", "usda", "usdc"]
     self.stage.SetEndTimeCode(self.frame_count)
+
+    # post-processing for visibility of geoms in scene
+    for _, geom_ref in self.geom_refs.items():
+      geom_ref.update_visibility(False, geom_ref.last_visible_frame+1)
+
     self.stage.Export(
         f"{self.output_directory_root}/{self.output_directory_name}/frames/frame_{self.frame_count}_.{filetype}"
```
xieleo5, line 422: Hi, is this a typo to have an additional `_` after `{self.frame_count}`?

xieleo5, line 422: Hi, is this a typo to have an additional `_` after `{self.frame_count}`?

xieleo5 (commented):
Hi, I used the branch and the exported file name will have an additional `_`. Is this intentional or a typo?

python/mujoco/usd/exporter.py
abhihjoshi: I remember this was intentional prior to when we were able to render a full trajectory. However, now, I do not believe there is a reason. I can remove this.

yuvaltassa (approved):

yuvaltassa (approved):

Issue #1394, opened 2024-02-02
Issue with Unity and MuJoCo 3 loading elasticity plugin


dblanm:
Hi,

I'm a PhD candidate trying to use MuJoCo 3.1.1 with Unity for simulating deformable objects.

I have followed the instructions [here](https://mujoco.readthedocs.io/en/latest/unity.html) and managed to set-up MuJoCo properly.
Now I am able to import MuJoCo scenes (I have tried importing [humanoid](https://github.com/google-deepmind/mujoco/blob/main/model/humanoid/22_humanoids.xml) and [sotfbox](https://github.com/google-deepmind/mujoco/blob/main/model/composite/softbox.xml) and the model is loaded into Unity).
I have tried to import the sample scene of the [poncho](https://github.com/google-deepmind/mujoco/blob/main/model/plugin/elasticity/poncho.xml).

However I get the following error when trying to do so:

```bash
IOException: Error loading the model: XML Error: unknown plugin 'mujoco.elasticity.shell'
Element 'plugin', line 24

Mujoco.MjEngineTool.LoadModelFromFile (System.String fileName) (at C:/Users/dblancom/Documents/UnityHub project/mujoco_src/unity/Runtime/Tools/MjEngineTool.cs:50)
Mujoco.MjImporterWithAssets.ImportFile (System.String filePath) (at C:/Users/dblancom/Documents/UnityHub project/mujoco_src/unity/Editor/Importer/MjImporterWithAssets.cs:53)
Mujoco.MjImporterEditorWindow.Apply () (at C:/Users/dblancom/Documents/UnityHub project/mujoco_src/unity/Editor/Importer/MjImporterEditorWindow.cs:28)
```

OS: Windows 11
Unity Version: Unity 2022.3.19f1

For setting up MuJoCo in Unity I did the following:
1. Create a Unity project
2. Download the [release 3.1.1](https://github.com/google-deepmind/mujoco/releases) of MuJoCo for Windows.
3. Paste the mujoco.dll into my project Assets folder.
4. Clone the source files of MuJoCo.
5. Import the MuJoCo package in Unity by clicking: Window->Package Manager-> Import from local disk

I have tried copying the file `elasticity.dll` from the MuJoCo release that is inside Mujoco/bin/mujoco_plugin  into my project Assets folder but this doesn't solve the issue.

How can I use the latest elasticity plugin with Unity?
I'm guessing @Balint-H should know about this?

Thanks in advance!

[...]

Issue #845, opened 2023-05-01
IOException: Error loading the model: XML Error: problem reading attribute 'range' Element 'joint', line 1


kaizu3:
**How to ask for help**

First, read our quick guide to
[asking good questions](https://github.com/deepmind/mujoco#asking-questions).
Below is a template for you to use:

Hi,

I'm a (student / professor / engineer) and I'm trying to use MuJoCo for a project.

I'm looking for some help with this error when I import the humanoid mujoco scene or any mujococ scene in unity I get this error ""IOException: Error loading the model: XML Error: problem reading attribute 'range' Element 'joint', line 1 "".

Here is a screenshot / video, illustrating my question:
![image](https://user-images.githubusercontent.com/130810535/235503934-acc6ccd1-31af-4b2d-ad0b-198eea8f08dd.png)

Please

erez-tom:
If you input a path to the text box called "Debug File Name" (at the top of the "Mj Global Settings" component) you can save the XML that's generated from your scene.  You can then inspect it (in particular, search for mentions of "range") and see which element is causing the problem.  Is this a fresh import from the humanoid XML, or a scene/prefab you've been modifying?

kaizu3:
It is a fresh import and i tried multiple models from different sources and versions of mujoco and i get the same error.
I tried it in mujoco and it worked just fine.
I linked the xml debug file if you can see any problem.
Thank you very much.

[HUMANOID.txt](https://github.com/deepmind/mujoco/files/11368750/HUMANOID.txt)

Issue #1709, opened 2024-06-03
unity  'mujoco.elasticity.cable' error


mg638:
hello. I'm trying to test Mujoco in Unity, but when loading the model, the error is displayed and I can't use it.

IOException: Error loading the model: XML Error: unknown plugin 'mujoco.elasticity.cable'

I am using version 3.1.3. Does Unity not support elasticity yet?

Balint-H:
Hello! Yes, at the moment elasticity plugin support is not added to the official branch yet, although preliminary tests were already done to confirm it is possible to add. I intended to revisit the topic at the start of July.

If it is a critical feature for you, and you don't mind tweaking the source code of the Unity plugin, I can help you get started before then though.

mg638:
If it's available, what should I do? It's okay to modify the plug-in.

mg638:
@Balint-H 

hello. The elasticity plugin error was previously resolved. However, when I play, the following error message appears. How do I resolve this?

IOException: Error loading the model: XML Error: Schema violation: unrecognized attribute: 'sensornoise'

Element 'flag', line 1

Balint-H:
Hello, this was an issue for a while for version 3.1.4 and 3.1.5 of the Unity plugin, but is fixed for 3.1.6. Could you please test with that? Thanks!

kevinzakka:
Going to close the issue assuming that upgrading fixes the problem. If that is not the case, please re-open the issue.

PR #1721, opened 2024-06-05
Fix MjHingeJoint.cs range with MakeLocaleInvariant

stonfute:
Seems to fix issues #1394  #845 by using localeInvariant on range

### DIFF
--- a/unity/Runtime/Components/Joints/MjHingeJoint.cs
+++ b/unity/Runtime/Components/Joints/MjHingeJoint.cs
@@ -83,7 +83,7 @@ namespace Mujoco {
       if (RangeLower > RangeUpper) {
         throw new ArgumentException("Lower range value can't be bigger than Higher");
       }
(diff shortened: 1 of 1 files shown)


Balint-H:
@erez-tom LGTM, Thanks for the contribution @stonfute! It doesn't completely fix #1709, but #845 could be solved by this (pending on confirmation from the issue's original author)

stonfute:
> It doesn't completely fix #1709, but #845 could be solved by this (pending on confirmation from the issue's original author)
My bad, wrong issue mentioned. I meant it will fix #845 according to sample xml provided. And also related to #1394 (hinge joint part with range issue)

Balint-H:
@erez-tom Hello Tom, I think this is still a fix pending that we need. Is there anything else that needs to be done before merging this?

erez-tom (approved):

PR #1749, opened 2024-06-21
UI Improvements

emotodorov:
When OpenGL buffer size is too small to hold entire UI with all sections open, older sections (as determined by latest mouse click) are automatically closed as needed.

To test this new functionality, uncomment the test near the top of mjui_resize().

UI items now have userid, which can be used for event handling.  It is set through the (new) last field of mjuiDef, called otherint.

Sections titles can now have a checkmark - in which case they are rendered differently. Set via mjuiDef.otherint: 0- none (as before), 1- rendering changed but box not shown, 2- box shown.  This can be used to label sections of special interest to the application.

### DIFF
--- a/include/mujoco/mjui.h
+++ b/include/mujoco/mjui.h
@@ -26,6 +26,7 @@
 #define mjMAXUIRECT     25      // maximum number of rectangles
 
 #define mjSEPCLOSED     1000    // closed state of adjustable separator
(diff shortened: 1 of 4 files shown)


yuvaltassa (approved):
Thanks!

yuvaltassa (approved):

PR #1753, opened 2024-06-22
'nstack' property of data is now removed, updating comment in mjData

JamieMair:
This is assuming that the size of `mjData.arena` is `mjData.narena*sizeof(mjtNum)` bytes


yuvaltassa:
Thanks for noticing this! Now corrected (it was actually just bytes).

PR #1769, opened 2024-06-30
Use Renderer context manager format in tutorial.py

### LINKED ISSUE #1746: Modify Renderer to use the context manager format in tutorials
Based on the suggestions from https://github.com/google-deepmind/mujoco/pull/1713#issuecomment-2162842911, 

> One thing that we probably need to do is change all uses of Renderer to use the [context manager format](https://github.com/google-deepmind/mujoco/blob/64cd44252fdf73d289970ff9fb2bc4c6a845850f/python/mujoco/renderer_test.py#L51), people keep tripping up on Renderer instances hanging around e.g. https://github.com/google-deepmind/mujoco/issues/1185 ...

I'll modify Renderer in the python tutorials to use context manager.

fanminshi:
TEST=I ran a side-sde comparison between before and this commit.  I observed the resulting simulation outputs are the same.
FIX #1746


yuvaltassa (approved):
Thanks!

yuvaltassa:
Thanks for this! As a side note, I had to manually fix some issues: You had assignments with no space (like `height=800`) and some indentations were 4 spaces instead of 2.

PR #1827, opened 2024-07-21
Open3d Dependency Removal from USD Module

abhihjoshi:
- Removing Open3D dependency for USD module.
- Also adding texture scaling and cube mapping


python/mujoco/usd/objects.py
```diff
@@ -16,13 +16,14 @@
 
 import abc
 import collections
-from typing import Optional, Dict, Any
+from typing import Optional, Dict, Any, Tuple
```
nimrod-gileadi, line 19: Sort these: Any, Dict, Optional, Tuple

python/mujoco/usd/camera.py
```diff
@@ -14,7 +14,7 @@
 # ==============================================================================
 """Camera handling for USD exporter."""
 
-import mujoco.usd.utils as utils_component
+import mujoco.usd.utils as utils_modules
```
nimrod-gileadi, line 17: Why the plural `modules`?

python/mujoco/usd/exporter.py
```diff
@@ -251,12 +251,10 @@ def _load_geom(self, geom: mujoco.MjvGeom):
 
     assert geom_name not in self.geom_names
 
-    texture_file = (
-        self.texture_files[
-            self.model.mat_texid[geom.matid][mujoco.mjTEXROLE_RGB]
-        ]
-        if geom.matid != -1
-        else None
+    geom_textures = (
```
nimrod-gileadi, line 254: Since in downstream code you don't have different handling of None vs an empty list (you just check `if self.geom_textures`), then you should prefer to just use the empty list to indicate no textures. This means you don't have to have Optional in all your function signatures.

Also, this is a complicated expression. Can you break it up into:

```
if geom.matid == -1:
  geom_textures = []
else:
  geom_textures = [(self.texture_files[i], self.model.tex_type[i]) if i != -1 else None for i in self.model.mat_texid[geom.matid]]
```

python/mujoco/usd/objects.py
```diff
@@ -50,16 +51,18 @@ class USDObject(abc.ABC):
   def __init__(
       self,
       stage: Usd.Stage,
+      model: mujoco.MjModel,
       geom: mujoco.MjvGeom,
       obj_name: str,
       rgba: np.ndarray = np.array([1, 1, 1, 1]),
-      texture_file: Optional[str] = None,
+      geom_textures: Optional[Tuple[str, mujoco.mjtTexture]] = None
```
nimrod-gileadi, line 58: Based on the previous file, I think this is actually

`Optional[List[Optional[Tuple[str, mujoco.mjtTexture]]]]`

Or if you took my suggestion on the last file:
`List[Optional[Tuple[str, mujoco.mjtTexture]]]`

Same below.

python/mujoco/usd/objects.py
```diff
@@ -334,42 +336,56 @@ def __init__(
...
 
   def _get_uv_geometry(self):
-    assert self.prim_mesh
+    assert self.prim_mesh and self.prim_mesh.triangle_uvs is not None
 
     mesh_texcoord = np.array(self.prim_mesh.triangle_uvs)
     mesh_facetexcoord = np.asarray(self.prim_mesh.triangles)
-    # TODO(etom): bring back support for rescaling the texture coordinates.
+    
+    geom_rgb_texture = self.geom_textures[mujoco.mjtTextureRole.mjTEXROLE_RGB][1]
```
nimrod-gileadi, line 369: Don't you need to handle the case where the texture tuple is None?

python/mujoco/usd/shapes.py
```diff
@@ -14,34 +14,244 @@
 # ==============================================================================
 """Built-in shapes for USD exporter."""
 
-from typing import Dict, Any
+from typing import Dict, Any, Tuple, Optional, Union
```
nimrod-gileadi, line 17: Sort

python/mujoco/usd/shapes.py
```diff
@@ -14,34 +14,244 @@
...
 
 import mujoco
 import numpy as np
-from open3d import open3d as o3d
 
-
-def create_hemisphere(
-    radius: float, theta_steps: int = 50, phi_steps: int = 50
+def get_triangle_uvs(
+    vertices: np.array,
```
nimrod-gileadi, line 23: The type is not np.array (which is a function). You can either use np.ndarray or use numpy.typing.ArrayLike

python/mujoco/usd/shapes.py
```diff
@@ -14,34 +14,244 @@
...
+      assert 0 <= u and u <= 1 and 0 <= v and v <= 1
+
+      triangle_uvs.append([u, v])
+  elif texture_type == mujoco.mjtTexture.mjTEXTURE_SKYBOX:
+    # defaults to 2D mapping temporarily
+    triangle_uvs = [[vertices[i][0], vertices[i][1]] for i in np.nditer(triangles)]
+
+  return np.array(triangle_uvs)
+
+class TriangleMesh():
```
nimrod-gileadi, line 73: Drop the `()`

python/mujoco/usd/shapes.py
```diff
@@ -14,34 +14,244 @@
...
+  """ Store UV and geometry information for a primitve mesh
```
nimrod-gileadi, line 74: Remove space after `"""`, fix "primitve" typo, add a period at the end of the sentence and move the closing `"""` up to this line.

python/mujoco/usd/shapes.py
```diff
@@ -14,34 +14,244 @@
...
+  """
+  def __init__(self,
+               vertices: np.array,
```
nimrod-gileadi, line 77: Fix the type

python/mujoco/usd/shapes.py
```diff
@@ -14,34 +14,244 @@
...
+    self.triangle_uvs = triangle_uvs
+
+  @classmethod
+  def create_box(
+      cls,
+      width: float,
+      height: float,
+      depth: float,
+      texture_type: Optional[mujoco.mjtTexture]
+  ):
```
nimrod-gileadi, line 91: Add return type. Same for other class methods.

nimrod-gileadi (changes requested):

python/mujoco/usd/camera.py
abhihjoshi: Ah, that was a mistake. Changed it!

python/mujoco/usd/objects.py
abhihjoshi: I believe I make this check prior to calling the function. In the __init__ function, I say:
```
if geom.matid and self.geom_textures[mujoco.mjtTextureRole.mjTEXROLE_RGB]:
      # setting mesh uv properties
      mesh_texcoord, _ = self._get_uv_geometry()
```
The if statement checks that the geom has a material and the RGB texture exists.

nimrod-gileadi (approved):

PR #1881, opened 2024-08-06
remove unused mjx.model class attribute

WangHanSolo:
As title :)

The recent release adds additional fields to mjx.model for further compatigbility with the mujoco structs, however `_sizes` doesn't seem to exist in mjModel.


WangHanSolo:
I stand corrected. `_sizes` is a field in mujoco python bindings.

PR #1901, opened 2024-08-13
Fixing USD pipeline after copybara changes

abhihjoshi:
- Checking if texture_type is None (`if not texture_type` is not the same as `if texture_type is None` since `texture_type` can have a value of 0)
- Wrapping class name as a string for return types for classmethods
- Adding tex uniform check for scaling textures


python/mujoco/usd/shapes.py
yuvaltassa, line 1: translate(), rotate() and scale() have bad annotations:
- `np.array` should be `np.ndarray`
- missing return annotation `-> None`

python/mujoco/usd/objects.py
yuvaltassa, line 1: Annotations in this file are bad
`geom_textures: List[Optional[Tuple[str, mujoco.mjtTexture]]] = None`
is bad because `None` is not a list.

I changed it to 
`geom_textures: List[Optional[Tuple[str, mujoco.mjtTexture]]] | None = None`


but then Nimrod said in his review:

"Two problems:

In the review with Abhishek, I pointed out that if the empty list and None are equivalent, it's better to just take an empty list and not have to check for Nones everywhere.

Instead of making the type optional and having a default of None, please change the type to Sequence, and default to empty tuple. If downstream code complains that it's a Sequence and not a List, you can change the assignment to `self.geom_textures = list(geom_textures)`. Note that using [] as a default value is problematic in Python."

The oldest Python version we target externally doesn't support | in type specifications. Need to use Optional or Union.

yuvaltassa (commented):
Sorry for breaking the `"TriangleMesh"` annotation, that was my bad.

I don't think there is any way that the test is currently running and passing, the golden file is all wrong and needs to be updated.

Also I think the test calls `exporter.save_scene('export.usda')` instead of `exporter.save_scene('usda')`, right?

If you fix the test you'll encounter other bugs, e.g. `prim_mesh` in objects.py can be `None`

yuvaltassa (approved):

Issue #1875, opened 2024-08-02
mj_ray performance regression from 3.1.1 to 3.2.0


michael-projectx:
I've recently upgraded from v3.1.1 to v3.2.0 and I'm seeing some large regressions in performance in simulations that call `mj_ray`. We use `mj_ray` to determine the visibility of certain objects in a scene. This post is twofold:
1. Perhaps we're using mj_ray inefficiently and there are some optimizations required on our end.
2. Perhaps there's been a regression in `mj_ray`.

For all of our simulations we track simulated time and wall clock time and provide a "realtime ratio" (wall clock time / simulated time) as an output. I noticed that many of our simulations showed little difference in realtime ratio between both mujoco version, but some were quite different. Using one of the simulations that showed a large regression in wall clock time (the sim takes much longer to run), I did some comparisons.

 I've run the same simulation using 3.1.1 and 3.2.0 to compare performance. Here is some data I've collected:

For both simulations the simulated time is 12.102 seconds
When running with v3.1.1, the wall clock time is XXX seconds (a ratio of X.XX)
When running with v3.2.0, the wall clock time is 110.86 seconds (a ratio of 9.16)

In order to dig deeper, I've run the above simulation using both versions of mujoco under `perf` to try to pinpoint what was running slower. It turns out that `mj_ray` is the culprit.

When running with v3.1.1, the calls to `mj_ray` account for 13.9% of the total execution time of the simulation
When running with v3.2.0, the calls to `mj_ray` account for 78.6% of the total execution time of the simulation

Diving deeper, `mju_rayTree` accounts for 13.1% of the simulation time for v3.1.1 and 78.3% of the simulation time for v3.2.0

`ray_triangle` which is called by `mju_rayTree` takes less that 1% under both versions.

What additional information can I provide in order to help debug this issue?

quagla:
Hi, could you provide the value of `model.nbvh` for your model before and after the change?

michael-projectx:
With mujoco 3.2.0 we have:
  model.nbody=96
  model.nbvh=2622528
  model.nbvhstatic=2622528
  model.nbvhdynamic=0

With mujoco 3.1.1 we have:
  model.nbody=96
  model.nbvh=2622599
  model.nbvhstatic=2622599
  model.nbvhdynamic=0

So, pretty similar, but not exactly he same.

quagla:
How much time is going into `mju_raySlab` for the two cases? Or where otherwise is the time going?

[...]

PR #1906, opened 2024-08-16
Fixed incorrect mesh face circumradius calculation

michael-ahn:
- In [this commit in v3.17](https://github.com/google-deepmind/mujoco/commit/8e2f830fd2d079d4391541bcbb4971d12e164e88), this was changed from `mju_norm3` to `mjuu_normvec`, but the latter will mutate its input to be normalized instead of just returning the norm.
  - Perhaps `mju_norm3` was mistaken as `mju_normalize3` which does mutate its input.
  - It looks like this was the only mistakenly changed callsite; other calls to `mju_norm3` were changed to `sqrt(mjuu_dot3(x, x))`
- This causes mesh ray casting to be incorrect; on some meshes I see a large increase in node visits within `mju_rayTree` (this fixes [issue #1875](https://github.com/google-deepmind/mujoco/issues/1875))
  - Before commit 8e2f830 on my mesh: `nstack` in `mju_rayTree` ~= 500. After the bug, nstack ~= 30000. With this fix, nstack drops back to ~500.

### DIFF
--- a/src/user/user_mesh.cc
+++ b/src/user/user_mesh.cc
@@ -1915,7 +1915,7 @@ void mjCMesh::MakeCenter(void) {
     // compute circumradius
     double norm_a_2 = mjuu_dot3(a, a);
     double norm_b_2 = mjuu_dot3(b, b);
-    double area = mjuu_normvec(nrm, 3);
+    double area = sqrt(mjuu_dot3(nrm, nrm));
 
     // compute circumcenter
     double res[3], vec[3] = {


quagla:
Amazing @michael-ahn well spotted!

yuvaltassa:
1. @michael-ahn, how the hell did you catch this??? 🥇 
2. @michael-ahn, @quagla, what would a test look like that would have caught this regression?

quagla:
> 2. @michael-ahn, @quagla, what would a test look like that would have caught this regression?

The golden tests did catch this, but there were so many other changes that we missed it.

yuvaltassa:
What about adding a test that asserts that `bvh_active` is smaller than X?
Evidently regressions in the BVH code are very hard to spot...

quagla:
We should probably have a test that checks the bvh for a given simple mesh, like one or two triangles.

iron76:
Thanks @michael-ahn! Well spotted!

PR #2078, opened 2024-09-23
Enable binary `.mjb` in `testspeed.py` and `mjx/viewer.py`

hartikainen:


mjx/mujoco/mjx/testspeed.py
```diff
@@ -50,7 +50,12 @@ def _main(argv: Sequence[str]):
   path = epath.resource_path('mujoco.mjx') / 'test_data'
   path = _BASE_PATH.value or path
   f = epath.Path(path) / _MJCF.value
-  m = mujoco.MjModel.from_xml_path(f.as_posix())
+  if f.suffix == '.xml':
```
btaba, line 53: This might be more appropriate:

```
if f.suffix == 'mjb':
  m = mujoco.MjModel.from_binary_path(f.as_posix())
else:
  m = mujoco.MjModel.from_xml_path(f.as_posix())
```

Since it [1] defaults to XML unless the file extension is mjb and [2] lets the MuJoCo compiler determine if the XML format is unknown.

btaba (changes requested):

hartikainen:
I added similar logic for `mjx/viewer.py`.

nimrod-gileadi (approved):

PR #2148, opened 2024-10-17
Custom camera for USD export

abhihjoshi:
Adding custom camera for USD export.

### DIFF
--- a/python/mujoco/usd/exporter.py
+++ b/python/mujoco/usd/exporter.py
@@ -18,14 +18,10 @@ import os
 from typing import List, Optional, Tuple, Union
 
 import mujoco
-# import mujoco.usd.camera as camera_module
-# import mujoco.usd.lights as light_module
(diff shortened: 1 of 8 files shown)


yuvaltassa:
CI is failing?

python/mujoco/usd/exporter.py
```diff
@@ -15,13 +15,16 @@
...
 import os
-from typing import List, Optional, Union
+from typing import List, Optional, Tuple, Union
 
 import mujoco
 import mujoco.usd.camera as camera_module
 import mujoco.usd.lights as light_module
 import mujoco.usd.objects as object_module
 import mujoco.usd.shapes as shapes_module
+from mujoco import _enums
```
nimrod-gileadi, line 25: You shouldn't be importing these internal modules. Everything from these modules is exposed directly from the mujoco module.

nimrod-gileadi (changes requested):

PR #2160, opened 2024-10-21
Fix running tests out of a installed mujoco python package

traversaro:
This PR should fix running `pytest --pyargs mujoco` in an environment in which `mujoco` is installed not in editable mode, by moving the `model.xml` file used by `spec_test.py` to the `testdata` folder of python bindings (as suggested by @saran-t and @yuvaltassa).

Furthermore, I modified the github actions test to ensure that they do not run in the `python/dist` folder, to ensure that they can catch this kind of regressions in the future.

I also integrated @hartikainen suggestions on the use of `epath`.

### DIFF
--- a/python/mujoco/specs_test.py
+++ b/python/mujoco/specs_test.py
@@ -18,6 +18,7 @@ import inspect
 import textwrap
 
 from absl.testing import absltest
+from etils import epath
 import mujoco
 import numpy as np
 
@@ -433,10 +434,11 @@ class SpecsTest(absltest.TestCase):
     )
 
   def test_load_xml(self):
-    filename = '../../test/testdata/model.xml'
     state_type = mujoco.mjtState.mjSTATE_INTEGRATION
 
     # Load from file.
+    file_path = epath.resource_path("mujoco") / "testdata" / "model.xml"
+    filename = file_path.as_posix()
     spec1 = mujoco.MjSpec.from_file(filename)
     model1 = spec1.compile()
     data1 = mujoco.MjData(model1)
@@ -692,9 +694,8 @@ class SpecsTest(absltest.TestCase):
                      mujoco.mjtGeom.mjGEOM_BOX)
 
   def test_delete(self):
-    filename = '../../test/testdata/model.xml'
-
-    spec = mujoco.MjSpec.from_file(filename)
+    file_path = epath.resource_path("mujoco") / "testdata" / "model.xml"
+    spec = mujoco.MjSpec.from_file(file_path.as_posix())
 
     model = spec.compile()
     self.assertIsNotNone(model)
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -239,7 +239,6 @@ jobs:
     - name: Test Python bindings
       if: ${{ runner.os != 'Windows' }}
       shell: bash
-      working-directory: python/dist
       env:
         MUJOCO_GL: disable
       run: >
--- a/python/mujoco/testdata/model.xml
+++ b/python/mujoco/testdata/model.xml
@@ -1,0 +1,178 @@
+<mujoco>
+  <option jacobian="dense" density="1.225" viscosity="1.8e-5" wind="0 0 1">
+    <flag fwdinv="enable" energy="enable" island="enable"/>
+  </option>
+
+  <asset>
+    <mesh name="icosahedron" scale=".05 .05 .05"
+          vertex="0         1       1.618
+                  0        -1       1.618
+                  0         1      -1.618
+                  0        -1      -1.618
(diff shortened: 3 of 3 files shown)


traversaro:
Actually this will not work as https://github.com/google-deepmind/mujoco/tree/680fb3e5ddbf21fb1e08877f2d03589ad01968ea/python/mujoco/testdata and https://github.com/google-deepmind/mujoco/tree/680fb3e5ddbf21fb1e08877f2d03589ad01968ea/test/testdata are not the same directory.

python/mujoco/specs_test.py
```diff
@@ -433,7 +435,9 @@ def test_kwarg(self):
     )
 
   def test_load_xml(self):
-    filename = '../../test/testdata/model.xml'
+    filename = os.path.join(
+      epath.resource_path("mujoco"), "testdata", "model.xml"
+    )
```
hartikainen, line 440: Would this work?
```suggestion
        file_path = epath.Path(__file__).parents[2] / "test" / "testdata" / "model.xml"
```

hartikainen: Then, below, you'd do:
```
spec1 = mujoco.MjSpec.from_file(file_path.as_posix())
```

python/mujoco/specs_test.py
```diff
@@ -692,8 +696,9 @@ def test_include(self):
                      mujoco.mjtGeom.mjGEOM_BOX)
 
   def test_delete(self):
-    filename = '../../test/testdata/model.xml'
-
+    filename = os.path.join(
+      epath.resource_path("mujoco"), "testdata", "model.xml"
+    )
     spec = mujoco.MjSpec.from_file(filename)
```
hartikainen, line 702:
```suggestion
    file_path = epath.Path(__file__).parents[2] / "test" / "testdata" / "model.xml"
    spec = mujoco.MjSpec.from_file(file_path.as_posix())
```

python/mujoco/specs_test.py
```diff
@@ -15,9 +15,11 @@
 """Tests for mjSpec bindings."""
 
 import inspect
+import os
```
hartikainen, line 18:
```suggestion
```
No need of `os` if you use paths.

python/mujoco/specs_test.py
traversaro: I am not sure. I guess the problem is that the `test/testdata` folder is never installed, so no matter which relative path you pass, if you install `mujoco` via for example a wheel, the needed file `model.xml` is not found anywhere.

yuvaltassa:
Hold on, is the issue that we're not exporting some file/files in the wheel? If yes then presumably that's easy to fix?

traversaro:
> Hold on, is the issue that we're not exporting some file/files in the wheel? If yes then presumably that's easy to fix?

Yes, the nutshell of the issue is that. The problem is that the Python MjSpec test access the model.xml in a folder that is not copied to the wheel. So the possible solutions I can think of are: 
* Duplicate the model.xml to the python testdata folder (easier, but you end up with duplicated files)
* Copy the model.xml in the python testdata folder in the make_sdist.sh script 


Which solution do you prefer?

traversaro:
To avoid regression of this kind, I opened https://github.com/google-deepmind/mujoco/pull/2163 (that should fail until this issue is fixed).

yuvaltassa:
I don't think the Python tests should depend on files external to that directory. I think forking the model file into Python is the correct thing, duplication is fine in this case.

@saran-t WDYT?

saran-t:
Yeah fork the model into the test data dir.

traversaro:
> Yeah fork the model into the test data dir.

Ok, I updated the PR to do that and modified the CI to catch future errors of this kind.

yuvaltassa:
Hi @traversaro, I ran the CI, something is obviously wrong. Let me know if you think you've fixed this and I will rerun (requires manual retriggering by one of us)/

traversaro:
My bad, I introduced @hartikainen suggestions without double checking, let me fix that.

traversaro:
> Let me know if you think you've fixed this and I will rerun (requires manual retriggering by one of us)/

Note to self: I did not realized that Actions run also on the fork, so the easiest option is to check there (in this case https://github.com/traversaro/mujoco/actions) instead of waiting for CI to be approved here (testing locally is obviously possible as well, but not straightforward w.r.t. to regular C++ with Python bindings projects due to the`make_sdist.sh` + `pip install` workflow instead of having a single command).

traversaro:
Ok, the CI seems to be green now, the PR is actually ready for review, thanks! @saran-t @yuvaltassa

saran-t (approved):

PR #2163, opened 2024-10-22
Do not test python bindings in python/dist directory

traversaro:
This will permit to early catch tests that rely on files not installed in the wheel, i.e. catch if `pytest --pyargs mujoco` fails in a vanilla environment.

This PR intentionally fails in CI to show the problem discussed in https://github.com/google-deepmind/mujoco/pull/2160 .


traversaro:
Closing as the same test was integrated in https://github.com/google-deepmind/mujoco/pull/2160 .

PR #2170, opened 2024-10-23
Adding the ability to install libsimulate library and associated headers during installation.

vidurvij-apptronik:
Aim
I want to use the libsimulate header in my internal simulator. I want to leverage the default mujoco visualizer instead of writing my own. Wanted to add the ability to install the libsimulate and headers in my dev container directly from CMake.

### DIFF
--- a/simulate/CMakeLists.txt
+++ b/simulate/CMakeLists.txt
@@ -41,16 +41,12 @@ if(APPLE)
   enable_language(OBJCXX)
 endif()
 
-if(NOT DEFINED expose_headers)
-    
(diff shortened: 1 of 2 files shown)


simulate/CMakeLists.txt
```diff
@@ -41,12 +41,16 @@ if(APPLE)
   enable_language(OBJCXX)
 endif()
 
+if(NOT DEFINED expose_headers)
```
saran-t, line 44: Why is there an empty conditional here?

PR #2185, opened 2024-10-28
Add precisions about position and orientation of meshes

dyl1n:
Add mathematical precisions about meshes position and orientation in the XML Reference section of the documentation, under the asset/mesh (*)  subsection

### DIFF
--- a/doc/XMLreference.rst
+++ b/doc/XMLreference.rst
@@ -1151,10 +1151,34 @@ be created as follows:
 Positioning and orienting is complicated by the fact that vertex data in the source asset are often relative to
 coordinate frames whose origin is not inside the mesh. In contrast, MuJoCo expects the origin of a geom's local frame to
 coincide with the geometric center of the shape. We resolve this discrepancy by pre-processing the mesh in the compiler,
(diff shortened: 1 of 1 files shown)


yuvaltassa:
Thanks for this proposal! I rendered this locally and unfortunately I don't think it's an improvement. This homogeneous transform notation is used nowhere else in the docs and introducing it here is just adding more complication to the reader.

I still think this section could probably be improved somewhat, but I think it should refer to existing fields in `mjModel` and `mjData`, rather than use mathematical notation.

Sorry if this is a disappointing review

dyl1n:
I understand your point about maintaining consistency with the existing documentation style. I agree that introducing new mathematical notations can have a counterproductive effect. 

As I mentioned in my original post, math notation was what really helped me clarify my own understanding, and I’m not the best person to explain things purely in English or with code snippets. I hope that my original post in the Discussion will still be helpful to others who come across it looking for clarity on this topic. If in the future I come up with new ideas to refine this section in a way that aligns better with the existing style, I’d be happy to revisit it. But for now I’ll close this PR.

Thank you for the feedback and for taking the time to review my proposal!

PR #2190, opened 2024-10-30
Fix typo in MJDATA_VECTOR definition

jjyyxx:


yuvaltassa (approved):
I can't believe nothing broke due to this very obvious bug :open_mouth: 

Thank you

PR #2218, opened 2024-11-07
Enable distance sensing with SDF geoms

### LINKED ISSUE #2216: Add distance sensing for SDF geoms

The distance sensor returns the distance between two bodies/geoms normally, but if one of the elements is an SDF geom, then it will only return the penetration distance if the geoms are in contact. So if the objects are not in contact then the distance sensor doesn't actually sense the distance and will just return the `cutoff` value.

I'll make this a feature request rather than a bug since this behavior seems intentional/place-holding based on the [mjc_SDF](https://github.com/google-deepmind/mujoco/blob/a2a0b95de01eebf08fa83e5f895a8212d3410a90/src/engine/engine_collision_sdf.c#L662) function that gets used when a distance sensor between an SDF and most other geom types is created.

### Alternatives

_No response_

### Additional context

I used the following model to test:
```
<mujoco model="distance_testing">
    <extension>
        <plugin plugin="mujoco.sdf.torus">
            <instance name="torus">
                <config key="radius1" value="0.35"/>
                <config key="radius2" value="0.15"/>
            </instance>
        </plugin>
    </extension>

    <asset>
        <mesh name="torus">
            <plugin instance="torus"/>
        </mesh>
    </asset>
    <option sdf_iterations="10" sdf_initpoints="40"/>

    <worldbody>
        <light name="top" pos="0 0 1"/>
        <body name="sphere" pos="1.0 0. 0." >
            <joint name="trans_1" type="slide" damping="100" axis="1 0 0"/>
            <joint name="trans_2" type="slide" damping="100" axis="0 1 0"/>
            <geom name="sphere_geom" type="sphere" size=".2 .2 .2" rgba="1 0 0 1" margin="1.0" gap="1.0"/>
        </body>

        <body name="sphere2" pos="0.0 1.0 0." >
            <geom name="sphere2_geom" type="sphere" size=".2 .2 .2" rgba="0 1 0 1" margin="1.0" gap="1.0"/>
        </body>

        <body name="torus" pos="0.0 0.0 0.0">
            <geom type="sdf" name="torus_geom" mesh="torus" rgba="0.4 0.4 0.4 1" margin="1.0" gap="1.0">
                <plugin instance="torus"/>
            </geom>
        </body>
    </worldbody>

    <actuator>
        <position name="trans_jnt_1" joint="trans_1" ctrlrange="-1 1" ctrllimited="true" kp="4000" kv="600"/>
        <position name="trans_jnt_2" joint="trans_2" ctrlrange="-1 1" ctrllimited="true" kp="4000" kv="600"/>
    </actuator>

    <sensor>
        <distance name="sphere_torus_distance_sensor" geom1="sphere_geom" geom2="torus_geom" cutoff="10"/>
        <distance name="sphere_sphere_distance_sensor" geom1="sphere_geom" geom2="sphere2_geom" cutoff="10"/>
    </sensor>
</mujoco>
```

You can see that the distance sensor reading between the sphere and the sphere changes value as you slide the mobile sphere around. But the distance sensor reading between the torus (SDF) and the mobile sphere never change unless the two geoms come into contact.

dyackzan:
Resolves #2216

### DIFF
--- a/src/engine/engine_collision_sdf.c
+++ b/src/engine/engine_collision_sdf.c
@@ -334,9 +334,9 @@ static int isknown(const mjtNum* points, const mjtNum x[3], int cnt) {
 // adds candidate point to result
 static int addContact(mjtNum* points, mjContact* con, const mjtNum x[3],
                       const mjtNum pos2[3], const mjtNum quat2[4], mjtNum dist,
-                      int cnt, const mjModel* m, const mjSDF* s, mjData* d) {
+                      int cnt, const mjModel* m, const mjSDF* s, mjData* d, mjtNum margin) {
   // check if there is a collision
-  if (dist > 0 || isknown(points, x, cnt)) {
+  if (dist > margin || isknown(points, x, cnt)) {
     return cnt;
   } else {
     mju_copy3(points+3*cnt, x);
@@ -652,7 +652,7 @@ int mjc_MeshSDF(const mjModel* m, const mjData* d, mjContact* con, int g1, int g
   // add only the first mjMAXCONPAIR pairs
   for (int i=0; i < mju_min(ncandidate, mjMAXCONPAIR); i++) {
     cnt = addContact(points, con, candidate + 3*index[i], pos2true, sdf_quat,
-                     dist[index[i]], cnt, m, &sdf, (mjData*)d);
+                     dist[index[i]], cnt, m, &sdf, (mjData*)d, margin);
   }
 
   return cnt;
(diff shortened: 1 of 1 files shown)


yuvaltassa:
Thanks!

Can you do the CLA thing please? 🙏

dyackzan:
Oh I thought I took care of it, is there something more I need to do? I see "✅ All contributors are covered under a CLA with Google" in the check output now: https://github.com/google-deepmind/mujoco/pull/2218/checks

dyackzan:
@kbayes In the new implementation will there be a solution for computing distances for SDF geometries? For my application I need to used SDF geometries and I need to be able to get the distance to the nearest contact with the distance sensor. If there is another solution to #2216 then I'm ok not merging this

yuvaltassa:
not longer relevant, I think?

PR #2225, opened 2024-11-12
Bazel build

bd-jahn:
# Summary

This PR introduces a Bazel build system that compiles the C library, Python bindings, and `mjx` library. While the original CMake build system supports multiple architectures and configurations, this Bazel build is currently configured only for Ubuntu 22.04 on x86_64.

This setup is not yet fully comprehensive. Some functionalities are still lacking, such as a hermetic toolchain, support for multiple OS and architecture builds, and compatibility with various rendering backends. However, this PR aims to initiate a discussion on the best direction for further improvements.

# Testing

## Build and test using the default Clang compiler:
Run `bazel build //...` and `bazel test //...`.

## Build and test using the GCC compiler:
Run `bazel build --config=linux_gcc_x86_64 //...` and `bazel test --config=linux_gcc_x86_64 //...`.

# Security

This build process fetches third-party dependencies through Bazel's `http_archive` rule, ensuring each dependency includes its license file.


tools/workspace/MarchingCubeCpp/LICENSE
saran-t, line 1: This file shouldn't be here

tools/workspace/abseil-cpp/repository.bzl
```diff
@@ -0,0 +1,10 @@
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+def abseil_cpp_repository(name = "abseil-cpp"):
+    version = "20240722.0"
```
saran-t, line 4: We really should have a mechanism to ensure that this is kept in sync with CMakeFiles...

tools/workspace/cereal/BUILD.bazel
saran-t, line 1: There are several libraries that aren't being pulled in for our CMake builds (cereal, spdlog, libglvnd, ...) -- why are these needed for Bazel?

saran-t (commented):
General comment: have you validated that you're passing the same compiler flags for the dependencies as we do for CMake builds?

WORKSPACE
```diff
@@ -0,0 +1,94 @@
+workspace(name = "mujoco")
```
hartikainen, line 1: This is using the [soon-to-be deprecated](https://bazel.build/external/migration) WORKSPACE system. It might be worth migrating this to use the new Bzlmod setup. I'll take a look at this if I have some extra time in the upcoming days. No guarantees though.

python/mujoco/__init__.py
```diff
@@ -14,6 +14,8 @@
 # ==============================================================================
 """Python bindings for MuJoCo."""
 
+__path__ = __import__('pkgutil').extend_path(__path__, __name__)
```
hartikainen, line 17: This line already exists below.

bd-jahn: Thanks, removed!

tools/workspace/cereal/BUILD.bazel
bd-jahn: These are transitive dependencies:  
For instance, `SdfLib` depends on `cereal` and `spdlog`, while `libglfw3` depends on `libglvnd`.  

The goal is to avoid using system-installed libraries and instead fetch them with a fixed SHA to ensure hermeticity.

tools/workspace/abseil-cpp/repository.bzl
bd-jahn: This is a good idea. Let me add that mechanism.

WORKSPACE
bd-jahn: Good point. I noticed that some dependencies are already in [Bazel Central Registry](https://registry.bazel.build/). Let me know how it goes - I’m happy to explore this as well.

bd-jahn:
I see that some of my changes to the include paths broke the CMake build system. I'll fix this issue.

bd-jahn:
Any idea of why [this](https://github.com/bd-jahn/mujoco/actions/runs/12014576432/job/33490695340#step:21:277) is failing? I only touched [generate_functions.py](https://github.com/google-deepmind/mujoco/pull/2225/files#diff-9d888cf694901278e3151db84a0cc08c8422f7d59690927375dc87b570d6caa4), [generate_functions.py](https://github.com/google-deepmind/mujoco/pull/2225/files#diff-9d888cf694901278e3151db84a0cc08c8422f7d59690927375dc87b570d6caa4), and [generate_structs.py](https://github.com/google-deepmind/mujoco/pull/2225/files#diff-ab123e89722b6d47dbd555d4094fb351ff48dfbda7fa310bd8bcebb7723aca6f), which shouldn't affect the test.

milutter:
Hey @saran-t 
we want to push a bit more on this PR. What are the minimal set of changes to merge the PR (other than rebase the branch to the current main and making sure everything still works)?

I think the two big discussion points are:
* using `bzlmod` instead of the `WORKSPACE` system
* supporting all platforms including MacOS & Windows. The current state is only working for linux using `gcc` & `clang` 

What is your opinion, on merging this PR without these changes? And extending the scope of the bazel build incrementally afterwards? 
 
 CC: @yuvaltassa, @hartikainen, @bd-jahn

saran-t (approved):

saran-t:
I'm so sorry about forgetting about this PR and thanks for prodding. I'm going to try and see how far I can get with importing this into our internal repo and setting up continuous builds for Bazel. In the meantime, could you please validate if everything is still working correctly at HEAD?

As an alternative to merging this into our own repo, would you consider maintaining a separate BD-owned repo for Bazel building infrastructure for MuJoCo? Given that you guys will be the de factor owner of this stuff, it is likely going to be less overhead and annoyance if you can avoid having to wait for us to merge PRs all the time. I'd suggest having something like a `mujoco_bazel` repo that has `mujoco` as a submodule.

bd-jahn:
@saran-t Thanks for your response. I'll update the PR and make sure everything works with the latest master. Looking forward to catching up in person for the next step!

hartikainen:
I took a stab at porting this into bzlmod setup earlier this year. I intended to finish that stuff but, as often happens, haven't had found time to do that. I'll see if I can get back to it sometime in the upcoming weekends. For me, it would also be important to get the macos build working as most of my development happens on a macbook.

saran-t (approved):

AustinSchuh:
Did this go anywhere?  I'm another bazel user who would love to not have to write my own build files.

manumerous:
Same here! :)

hartikainen:
@bd-jahn, I tried running the the bazel build from the latest main (v3.3.3) and had to make some changes on my way. Feel free to pick those changes here. They're available at https://github.com/google-deepmind/mujoco/compare/main...hartikainen:mujoco:bazel-build-v3.3.3. I haven't verified that things fully work but `bazel build --config=linux_gcc_x86_64 //...` at least succeeds without errors.

AustinSchuh:
2 thoughts.

1) This probably needs to now have bzlmod support.
2) Any reason not to just submit this as a bzlmod module to bcr directly at this point?  The build files could be overlays.

wep21:
I'm trying to add mujoco as bzlmod into bcr.
https://github.com/bazelbuild/bazel-central-registry/pull/9702

wep21:
mujoco is now available from bcr.

yuvaltassa:
Ping on this, do people still want first-party Bazel support?

hartikainen:
We would use it. As noted by @wep21, someone has created a MuJoCo package for [Bazel Central Registry](https://registry.bazel.build/modules/mujoco) but it's a bit older version.

PR #2246, opened 2024-11-24
Remove nroll argument from rollout

aftersomemath:
`nroll` can always be implied from the other arguments of rollout.

I grepped the codebase and did not find any cases where `nroll` was being passed to rollout other than its threading test.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -15,6 +15,7 @@ Python bindings
 - Provide prebuilt wheels for Python 3.13.
 - Added ``bind`` method and removed id attribute from :ref:`mjSpec` objects. Using ids is error prone in scenarios of repeated attachment and
   detachment. Python users are encouraged to use names for unique identification of model elements.
+- Removed ``nroll`` argument from :ref:`rollout<PyRollout>` because its value can always be inferred.
 
 Bug fixes
 ^^^^^^^^^
--- a/python/mujoco/rollout.py
+++ b/python/mujoco/rollout.py
@@ -29,7 +29,6 @@ def rollout(model: mujoco.MjModel,
             *,  # require subsequent arguments to be named
             control_spec: int = mujoco.mjtState.mjSTATE_CTRL.value,
             skip_checks: bool = False,
-            nroll: Optional[int] = None,
             nstep: Optional[int] = None,
             initial_warmstart: Optional[npt.ArrayLike] = None,
             state: Optional[npt.ArrayLike] = None,
@@ -50,7 +49,6 @@ def rollout(model: mujoco.MjModel,
       ([nroll or 1] x [nstep or 1] x ncontrol)
     control_spec: mjtState specification of control vectors.
     skip_checks: Whether to skip internal shape and type checks.
-    nroll: Number of rollouts (inferred if unspecified).
     nstep: Number of steps in rollouts (inferred if unspecified).
     initial_warmstart: Initial qfrc_warmstart array (optional).
       ([nroll or 1] x nv)
@@ -74,7 +72,7 @@ def rollout(model: mujoco.MjModel,
   #   don't allocate output arrays
   #   just call rollout and return
   if skip_checks:
-    _rollout.rollout(model, data, nroll, nstep, control_spec, initial_state,
+    _rollout.rollout(model, data, nstep, control_spec, initial_state,
                      initial_warmstart, control, state, sensordata)
     return state, sensordata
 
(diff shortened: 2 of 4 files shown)


doc/changelog.rst
```diff
@@ -56,20 +56,24 @@ MJX
 11. Added support for ``eq_active``. Fixes :github:issue:`2173`.
 12. Added ray intersection with ellipsoid.
 
+Python bindings
+^^^^^^^^^^^^^^^
+13. Removed ``nroll`` argument from rollout because its value can always be implied
```
yuvaltassa, line 61: This is the wrong section? Should be at the top of the file "Upcoming version (not yet released)"

python/mujoco/rollout.py
```diff
@@ -69,12 +67,12 @@ def rollout(model: mujoco.MjModel,
     ValueError: bad shapes or sizes.
   """
   # skip_checks shortcut:
-  #   don't infer nroll/nstep
```
yuvaltassa, line 72: I understand why we don't need the argument, but it still needs to be inferred, right?

For example, if I have a single initial state but multiple control sequences for that state, or a single state and control sequence, but multiple warmstart vectors.

Can you confirm that this is still supported? (maybe add a test for these if they do not already exist)

yuvaltassa (commented):
Looks good, some comments inline

doc/changelog.rst
aftersomemath: Whoops, my apologies.

python/mujoco/rollout.py
aftersomemath: Yes it is still supported, though when using the `skip_checks` shortcut the user must ensure all of these arrays have a leading dimension of `nroll`. This is what the non-`skip_checks` version of rollout does before calling into the bindings.

It seems there is only one test related to inferring `nroll`. It is `test_one_rollout_fixed_ctrl`. It tests the case of passing a single initial state and control in but with nroll state and sensordata vectors. I will add tests for some of the other cases.

yuvaltassa: My point is that "#   don't infer nroll/nstep" is still a correct comment, is it not? nroll is not an argument of the python API but still an argument of the low level API and that is inferred when skip_checks is False. Right?

doc/changelog.rst
```diff
@@ -18,6 +18,7 @@ MJX
 Python bindings
 ^^^^^^^^^^^^^^^
 - Provide prebuilt wheels for Python 3.13.
+- Removed ``nroll`` argument from rollout because its value can always be implied
```
yuvaltassa, line 21: from rollout -> from :ref:`rollout<PyRollout>`

(apologies for the markdown, please look at the source of this comment)

python/mujoco/rollout.py
```diff
@@ -29,15 +29,14 @@ def rollout(model: mujoco.MjModel,
...
-            nroll: Optional[int] = None,
             nstep: Optional[int] = None,
             initial_warmstart: Optional[npt.ArrayLike] = None,
             state: Optional[npt.ArrayLike] = None,
             sensordata: Optional[npt.ArrayLike] = None):
   """Rolls out open-loop trajectories from initial states, get subsequent states and sensor values.
 
   Python wrapper for rollout.cc, see documentation therein.
-  Infers nroll and nstep.
+  Infers nstep.
```
yuvaltassa, line 39: Same as below, nroll is still inferred, just that now we are inferring something that is not exposed to Python

python/mujoco/rollout.py
yuvaltassa: Yes please, add that test, and then I think we're done

aftersomemath: You're right, the old comment is the correct comment. My thinking was along the lines of "`nroll` is not an argument therefore it cannot be an inferred argument". But the comment string is not specific to arguments.

python/mujoco/rollout.py
aftersomemath: Yes, for the same reason as the other comment.

doc/changelog.rst
yuvaltassa: I'm asking to convert the word "rollout" to an RST link

yuvaltassa: Like here
https://github.com/google-deepmind/mujoco/blob/e9a2f055f044fa2817b41bbc6ddd23a20f58c830/doc/changelog.rst?plain=1#L558

aftersomemath: Yes, got it!

aftersomemath:
One test was added for each argument that can be used to infer `nroll`.

doc/changelog.rst
```diff
@@ -18,6 +18,7 @@ MJX
 Python bindings
 ^^^^^^^^^^^^^^^
 - Provide prebuilt wheels for Python 3.13.
+- Removed ``nroll`` argument from :ref:`rollout<PyRollout>` because its value can always be implied
```
yuvaltassa, line 21: 1. implied -> inferred
2. Add full stop

yuvaltassa:
Please sync to latest

doc/changelog.rst
aftersomemath: Done

aftersomemath:
Squashed and rebased

PR #2247, opened 2024-11-24
Fix mj_saveXMLString's return value in case of success

JafarAbdi:
- Fix mj_saveXMLString's return value in case of success.
- Fix test case for mj_saveXMLString.


quagla:
@JafarAbdi Thanks for the PR, could you please also update the comment above the function, in the header, and in doc/includes/references.h?

JafarAbdi:
> @JafarAbdi Thanks for the PR, could you please also update the comment above the function, in the header, and in doc/includes/references.h?

Thanks for the review, updated!

quagla (approved):

PR #2250, opened 2024-11-26
Allow rollout to accept a list of models of length nroll

aftersomemath:
`rollout` can now accept a list of models of length `nroll`, similar to its other arguments.

The models must be compatible, meaning `nstate`, `ncontrol`, `nv`, and `nsensordata` must be equal.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -16,6 +16,7 @@ Python bindings
 - Added ``bind`` method and removed id attribute from :ref:`mjSpec` objects. Using ids is error prone in scenarios of repeated attachment and
   detachment. Python users are encouraged to use names for unique identification of model elements.
 - Removed ``nroll`` argument from :ref:`rollout<PyRollout>` because its value can always be inferred.
+- :ref:`rollout<PyRollout>` can now accept lists of MjModel of length ``nroll``.
 
 Bug fixes
 ^^^^^^^^^
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -16,7 +16,8 @@ Python bindings
 - Added ``bind`` method and removed id attribute from :ref:`mjSpec` objects. Using ids is error prone in scenarios of repeated attachment and
   detachment. Python users are encouraged to use names for unique identification of model elements.
 - Removed ``nroll`` argument from :ref:`rollout<PyRollout>` because its value can always be inferred.
-- :ref:`rollout<PyRollout>` can now accept lists of MjModel of length ``nroll``.
+- :ref:`rollout<PyRollout>` can now accept lists of MjModel of length ``nroll``. ``nroll`` argument deprecated because
+  its value can always be inferred.
 
 Bug fixes
 ^^^^^^^^^
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -16,7 +16,7 @@ Python bindings
 - Added ``bind`` method and removed id attribute from :ref:`mjSpec` objects. Using ids is error prone in scenarios of repeated attachment and
   detachment. Python users are encouraged to use names for unique identification of model elements.
 - Removed ``nroll`` argument from :ref:`rollout<PyRollout>` because its value can always be inferred.
-- :ref:`rollout<PyRollout>` can now accept lists of MjModel of length ``nroll``. ``nroll`` argument deprecated because
+- :ref:`rollout<PyRollout>` can now accept sequences of MjModel of length ``nroll``. ``nroll`` argument deprecated because
   its value can always be inferred.
 
 Bug fixes
--- a/python/mujoco/rollout.cc
+++ b/python/mujoco/rollout.cc
@@ -73,26 +73,28 @@ void _unsafe_rollout(std::vector<const mjModel*>& m, mjData* d, int nroll, int n
   if (!(control_spec & mjSTATE_XFRC_APPLIED)) {
     mju_zero(d->xfrc_applied, 6*nbody);
   }
-  if (!(control_spec & mjSTATE_MOCAP_POS)) {
-    for (int i = 0; i < nbody; i++) {
-      int id = m[0]->body_mocapid[i];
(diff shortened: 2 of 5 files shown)


python/mujoco/rollout_test.py
```diff
@@ -677,13 +706,17 @@ def py_rollout(model, data, initial_state, control,
   control = ensure_3d(control)
   nroll = initial_state.shape[0]
   nstep = control.shape[1]
-  nstate = mujoco.mj_stateSize(model, mujoco.mjtState.mjSTATE_FULLPHYSICS)
+
+  if isinstance(model, mujoco.MjModel):
+    model = [model]*nroll
```
yuvaltassa, line 711: I see what you're doing here, but I'm not sure I like it. Imagine the "regular" case of single-model, multiple data. Say that I want a really wide batch 10k rollouts of 1 step each. In this case this line, whether called here or by the user could get very expensive due to 10k copies of MjModel.

The solution is for the pybind module and unsafe_rollout_ to correctly handle the special case where the list is of length 1, rather than to always require nroll MjModels. Does that make sense?

python/mujoco/rollout.cc
```diff
@@ -37,7 +37,7 @@ const auto rollout_doc = R"(
 Roll out open-loop trajectories from initial states, get resulting states and sensor values.
 
   input arguments (required):
-    model              instance of MjModel
+    model              list of MjModel instances of length nroll
```
yuvaltassa, line 40: - list -> array
- MjModel -> mjModel

Also, I think this function needs another argument that tells it if there is one model or `nroll` models, see my reasoning below. This should probably be `int nmodel` which is guaranteed to be either `nroll` or 1.

python/mujoco/rollout_test.py
aftersomemath: So, this doesn't have a big performance impact because this copies the reference not the model. For 10k models, it takes 0.031 milliseconds to construct this list. The rest of the rollout functions checks (prior to calling into the bindings) takes 1.5 milliseconds. Taking one step over 10k `humanoid.xml` models takes about 250 milliseconds single threaded.

I can still make it a special case if wanted, but it won't affect performance. At the least a comment is in order.

(also this review comment is on the test, but the actual rollout function in rollout.py uses the same list expansion).

yuvaltassa: Ah that's good to know! Resolving.

doc/changelog.rst
```diff
@@ -16,6 +16,7 @@ Python bindings
 - Added ``bind`` method and removed id attribute from :ref:`mjSpec` objects. Using ids is error prone in scenarios of repeated attachment and
   detachment. Python users are encouraged to use names for unique identification of model elements.
 - Removed ``nroll`` argument from :ref:`rollout<PyRollout>` because its value can always be inferred.
+- :ref:`rollout<PyRollout>` can now accept lists of MjModel of length ``nroll``.
```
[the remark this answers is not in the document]
aftersomemath: Done

python/mujoco/rollout.cc
yuvaltassa: You're right, I misread.

yuvaltassa (approved):

python/mujoco/rollout.cc
```diff
@@ -188,15 +188,20 @@ PYBIND11_MODULE(_rollout, pymodule) {
...
          const PyCArray state0,
          std::optional<const PyCArray> warmstart0,
          std::optional<const PyCArray> control,
          std::optional<const PyCArray> state,
          std::optional<const PyCArray> sensordata
          ) {
-        const raw::MjModel* model = m.get();
+        // get raw pointers
+        int nroll = state0.shape(0);
+        const raw::MjModel* model_ptrs[nroll];
```
yuvaltassa, line 201: Hold on, you can't do this, this would be a variable length array on the stack.

yuvaltassa: Saran says: use std::vector here

aftersomemath: fixed

yuvaltassa (approved):

yuvaltassa:
I'm getting 

```
[ RUN      ] MuJoCoRolloutTest.test_detect_divergence
[third_party/crosstool/v18/stable/src/libcxx/include/__vector/vector.h:1114](https://cs.corp.google.com/piper///depot/google3/third_party/crosstool/v18/stable/src/libcxx/include/__vector/vector.h?l=1114&ws=tap-presubmit-rerun-server/51936822&snapshot=2): assertion __n < size() failed: vector[] index out of bounds
```

In line 233, shouldn't you be passing `model_ptrs.data()` ? But I don't understand how the test are passing on GH...

aftersomemath:
It looks like the internal check passed, but I did not do anything. I'm more worried that there is a bug now (and passing a raw pointer using `model_ptrs.data()` would just mask that). Was anything done that made the internal check pass?

aftersomemath:
I found the bug, I created the `std::vector` incorrectly.

nimrod-gileadi (changes requested):

python/mujoco/rollout.cc
```diff
@@ -75,26 +75,26 @@ void _unsafe_rollout(const mjModel* m, mjData* d, int nroll, int nstep, unsigned
   }
   if (!(control_spec & mjSTATE_MOCAP_POS)) {
     for (int i = 0; i < nbody; i++) {
-      int id = m->body_mocapid[i];
-      if (id >= 0) mju_copy3(d->mocap_pos+3*id, m->body_pos+3*i);
+      int id = m[0]->body_mocapid[i];
+      if (id >= 0) mju_copy3(d->mocap_pos+3*id, m[0]->body_pos+3*i);
```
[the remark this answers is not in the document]
aftersomemath: Yes, thanks! There is no guarantee that the models use the same default values for these unspecified controls.

aftersomemath:
All requested changes implemented.

saran-t (approved):

PR #2276, opened 2024-12-07
Reduce cmake test boilerplate

traversaro:
Before this PR, all instances of `mujoco_test` in the codebase were followed by a call to `target_link_libraries(<testname> fixture gmock)`.

As anyhow `mujoco_test` already was calling target_link_libraries to some predefined list of targets (`mujoco` and `gtest_main`), this PR adds to the list of default linked targets also `fixture` and `gmock`, to reduce the boilerplate required to add a test.

Furthermore, to completely remove the need for calling target_link_libraries after a call to `mujoco_test`, this PR also adds to the `mujoco_test` macro the `ADDITIONAL_LINK_LIBRARIES` argument, that can be used if a given test needs to link some additional targets beside the default ones.

To permit to use these new features also for mujoco benchmarks, this PR adds a `MAIN_TARGET` parameter to mujoco_test, to select if `gtest_main` or another target is used to provide the main
entry point to the test executable, and migrate the existing usage of `mujoco_benchmark_test` to use `mujoco_test`.

This change is a requirement to add a `TAGS` argument to the `mujoco_test` macro, to permit to skip the compilation of a given test under some condition.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -17,10 +17,22 @@ set(MUJOCO_TEST_WORKING_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 
 include(GoogleTest)
 
+# This macro is used to add a C++ test to MuJoCo CMake build system
+#
+# The CMake targets linked by default with target_link_libraries are:
+#   mujoco fixture gmock
+#
+# The macro supports the following parameters:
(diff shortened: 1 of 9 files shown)


traversaro:
If CI fails, it will be due to https://github.com/google-deepmind/mujoco/pull/2277 .

PR #2282, opened 2024-12-12
Add a persistent thread pool to rollout

aftersomemath:
`rollout` now accepts a list of MjData of length `nthread`. If passed, the `nroll` rollouts are parallelized.

The thread pool is persistent across calls to rollout, so long as the number of MjData instances is constant.

The thread pool implementation was copied from MJPC. It is not threadsafe. This means that if rollout is using threads, it should not be called from multiple threads.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -11,8 +11,9 @@ Bug fixes
 
 Python bindings
 ^^^^^^^^^^^^^^^
-- :ref:`rollout<PyRollout>` can now accept sequences of MjData of length ``nthread``. If passed, :ref:`rollout<PyRollout>`
-  will automatically create a persistent threadpool and parallelize rollouts.
+- :ref:`rollout<PyRollout>` now features native multi-threading. If a sequence of MjData instances
+  of length ``nthread`` is passed in, ``rollout`` will automatically create a persistent threadpool
+  and parallelize the computation. Contribution by :github:user:`aftersomemath`.
 
 Version 3.2.6 (Dec 2, 2024)
 ---------------------------
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -9,6 +9,11 @@ Bug fixes
 ^^^^^^^^^
 - Fixed a bug in the box-sphere collider, depth was incorrect for deep penetrations (:github:issue:`2206`).
 
+Python bindings
+^^^^^^^^^^^^^^^
+- :ref:`rollout<PyRollout>` can now accept sequences of MjData of length ``nthread``. If passed, :ref:`rollout<PyRollout>`
+  will automatically create a persistent threadpool and parallelize rollouts.
+
 Version 3.2.6 (Dec 2, 2024)
 ---------------------------
 
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -12,8 +12,10 @@ Bug fixes
 Python bindings
 ^^^^^^^^^^^^^^^
 - :ref:`rollout<PyRollout>` now features native multi-threading. If a sequence of MjData instances
-  of length ``nthread`` is passed in, ``rollout`` will automatically create a persistent threadpool
-  and parallelize the computation. Contribution by :github:user:`aftersomemath`.
+  of length ``nthread`` is passed in, ``rollout`` will automatically create a thread pool and parallelize
+  the computation. The thread pool can resused across calls, but then the function cannot be called simultaneously
+  from multiple threads. To run multiple threaded rollouts simultaneously, use the new class ``Rollout`` which
+  encapsulates the thread pool. Contribution by :github:user:`aftersomemath`.
 
 Version 3.2.6 (Dec 2, 2024)
 ---------------------------
--- a/doc/python.rst
+++ b/doc/python.rst
@@ -711,18 +711,20 @@ The ``mujoco`` package contains two sub-modules: ``mujoco.rollout`` and ``mujoco
 rollout
 -------
 
-``mujoco.rollout`` shows how to add additional C/C++ functionality, exposed as a Python module via pybind11. It is
-implemented in `rollout.cc <https://github.com/google-deepmind/mujoco/blob/main/python/mujoco/rollout.cc>`__
+``mujoco.rollout`` and ``mujoco.Rollout`` shows how to add additional C/C++ functionality, exposed as a Python module
+via pybind11. It is implemented in `rollout.cc <https://github.com/google-deepmind/mujoco/blob/main/python/mujoco/rollout.cc>`__
 and wrapped in `rollout.py <https://github.com/google-deepmind/mujoco/blob/main/python/mujoco/rollout.py>`__. The module
 performs a common functionality where tight loops implemented outside of Python are beneficial: rolling out a trajectory
 (i.e., calling :ref:`mj_step` in a loop), given an intial state and sequence of controls, and returning subsequent
-states and sensor values. The basic usage form is
+states and sensor values. The rollouts are run in parallel with an internally managed thread pool if multiple MjData instances
+(one per thread) are passed as an argument. The basic usage form is
 
 .. code-block:: python
 
    state, sensordata = rollout.rollout(model, data, initial_state, control)
 
 ``model`` is either a single instance of MjModel or a sequence of compatible MjModel of length ``nroll``.
+``data`` is either a single instance of MjData or a sequence of compatible MjData of length ``nthread``.
 ``initial_state`` is an ``nroll x nstate`` array, with ``nroll`` initial states of size ``nstate``, where
 ``nstate = mj_stateSize(model, mjtState.mjSTATE_FULLPHYSICS)`` is the size of the
 :ref:`full physics state<geFullPhysics>`. ``control`` is a ``nroll x nstep x ncontrol`` array of controls. Controls are
@@ -732,13 +734,39 @@ specified by passing an optional ``control_spec`` bitflag.
 If a rollout diverges, the current state and sensor values are used to fill the remainder of the trajectory.
 Therefore, non-increasing time values can be used to detect diverged rollouts.
 
-The ``rollout`` function is designed to be completely stateless, so all inputs of the stepping pipeline are set and any
+The ``rollout`` function is designed to be computationally stateless, so all inputs of the stepping pipeline are set and any
 values already present in the given ``MjData`` instance will have no effect on the output.
 
-Since the Global Interpreter Lock can be released, this function can be efficiently threaded using Python threads. See
-the ``test_threading`` function in
+By default ``rollout.rollout`` creates a new thread pool every call if ``len(data) > 1``. To reuse the thread pool
+over multiple calls use the ``persistent_pool`` argument. ``rollout.rollout`` is not thread safe when using
+a persistent pool. The basic usage form is
+
+.. code-block:: python
+
+   state, sensordata = rollout.rollout(model, data, initial_state, persistent_pool=True)
+
+The pool is shutdown on interpreter shutdown or by a call to ``rollout.shutdown_persistent_pool``.
+
+To use multiple thread pools from multiple threads, use ``Rollout`` objects. The basic usage form is
+
+.. code-block:: python
+
+   # Pool shutdown upon exiting block
+   with rollout.Rollout(nthread) as rollout_:
+    rollout_.rollout(model, data, initial_state)
+
+or
+
+.. code-block:: python
+
+  # pool shutdown on object deletion, interpreter shutdown, or call to rollout_.shutdown_pool
+  rollout_ = rollout.Rollout(nthread)
+  rollout_.rollout(model, data, initial_state)
+
+Since the Global Interpreter Lock is released, this function can also be threaded using Python threads. However, this
+is less efficient than using native threads. See the ``test_threading`` function in
 `rollout_test.py <https://github.com/google-deepmind/mujoco/blob/main/python/mujoco/rollout_test.py>`__ for an example
-of threaded operation (and more generally for usage examples).
+of threaded operation (and for more general usage examples).
 
 .. _PyMinimize:
 
--- a/doc/python.rst
+++ b/doc/python.rst
@@ -759,9 +759,11 @@ or
 
 .. code-block:: python
 
-  # pool shutdown on object deletion, interpreter shutdown, or call to rollout_.shutdown_pool
+  # pool shutdown on object deletion or call to rollout_.shutdown_pool
+  # to ensure clean shutdown of threads, call shutdown_pool before interpreter exit
   rollout_ = rollout.Rollout(nthread)
   rollout_.rollout(model, data, initial_state)
+  rollout_.shutdown_pool()
 
 Since the Global Interpreter Lock is released, this function can also be threaded using Python threads. However, this
 is less efficient than using native threads. See the ``test_threading`` function in
--- a/doc/python.rst
+++ b/doc/python.rst
@@ -711,7 +711,7 @@ The ``mujoco`` package contains two sub-modules: ``mujoco.rollout`` and ``mujoco
 rollout
 -------
 
-``mujoco.rollout`` and ``mujoco.Rollout`` shows how to add additional C/C++ functionality, exposed as a Python module
+``mujoco.rollout`` and ``mujoco.rollout.Rollout`` shows how to add additional C/C++ functionality, exposed as a Python module
 via pybind11. It is implemented in `rollout.cc <https://github.com/google-deepmind/mujoco/blob/main/python/mujoco/rollout.cc>`__
 and wrapped in `rollout.py <https://github.com/google-deepmind/mujoco/blob/main/python/mujoco/rollout.py>`__. The module
 performs a common functionality where tight loops implemented outside of Python are beneficial: rolling out a trajectory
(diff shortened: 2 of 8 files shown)


python/mujoco/rollout.cc
```diff
@@ -158,6 +161,47 @@ void _unsafe_rollout(std::vector<const mjModel*>& m, mjData* d, int nroll, int n
   }
 }
 
+// C-style threaded version of _unsafe_rollout
+static ThreadPool* pool = nullptr;
```
aftersomemath, line 165: I used a raw pointer here instead of a smart pointer because the [style guide states](https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables) that objects with static duration must be trivially destructible. My understanding is that smart pointers are not trivially destructible.

doc/changelog.rst
```diff
@@ -9,6 +9,11 @@ Bug fixes
 ^^^^^^^^^
 - Fixed a bug in the box-sphere collider, depth was incorrect for deep penetrations (:github:issue:`2206`).
 
+Python bindings
+^^^^^^^^^^^^^^^
+- :ref:`rollout<PyRollout>` can now accept sequences of MjData of length ``nthread``. If passed, :ref:`rollout<PyRollout>`
```
yuvaltassa, line 14:
```
- :ref:`rollout<PyRollout>` now features native multi-threading. If a sequence of MjData instances
  of length ``nthread`` is passed in, ``rollout`` will automatically create a persistent threadpool
  and parallelize the computation. Contribution by :github:user:`aftersomemath`.
```

python/mujoco/rollout.cc
```diff
@@ -230,9 +280,24 @@ PYBIND11_MODULE(_rollout, pymodule) {
...
           // call unsafe rollout function
-          InterceptMjErrors(_unsafe_rollout)(
-              model_ptrs, data, nroll, nstep, control_spec, state0_ptr,
-              warmstart0_ptr, control_ptr, state_ptr, sensordata_ptr);
+          if (nthread > 1 && nroll > 1) {
+            int chunk_size_final = 1;
+            if (!chunk_size.has_value()) {
+              chunk_size_final = std::max(1, nroll / (10 * nthread));
+            }
+            else {
```
yuvaltassa, line 288: The else on a new line thing is something that we do in C code, and only to facilitate putting a comment above the block. Let's attach all else's `} else {` in C++ code as in the style guide.

python/mujoco/rollout.cc
```diff
@@ -158,6 +161,47 @@ void _unsafe_rollout(std::vector<const mjModel*>& m, mjData* d, int nroll, int n
...
+void _unsafe_rollout_threaded(std::vector<const mjModel*>& m, std::vector<mjData*>& d,
+                              int nroll, int nstep, unsigned int control_spec,
+                              const mjtNum* state0, const mjtNum* warmstart0,
+                              const mjtNum* control, mjtNum* state, mjtNum* sensordata,
+                              int nthread, int chunk_size) {
+  int nfulljobs = nroll / chunk_size;
+  int chunk_remainder = nroll % chunk_size;
+  int njobs = (chunk_remainder > 0) ? nfulljobs + 1 : nfulljobs;
+
+  if (pool == nullptr) {
```
yuvaltassa, line 175: added block comments, slightly refactored
```
  // if existing thread pool of the wrong size, delete
  if (pool->NumThreads() != nthread) {
    delete pool;
  }

  // make threadpool if required
  if (pool == nullptr) {
    pool = new ThreadPool(nthread);
  } else {
    pool->ResetCount();
  }
```

python/mujoco/rollout.cc
```diff
@@ -158,6 +161,47 @@ void _unsafe_rollout(std::vector<const mjModel*>& m, mjData* d, int nroll, int n
...
+    pool = new ThreadPool(nthread);
+  }
+  else if (pool->NumThreads() != nthread) {
+    delete pool;
+    pool = new ThreadPool(nthread);
+  } else {
+    pool->ResetCount();
+  }
+
+  for (int j = 0; j < nfulljobs; j++) {
```
yuvaltassa, line 185: // schedule all jobs of full (chunk) size

python/mujoco/rollout.cc
```diff
@@ -158,6 +161,47 @@ void _unsafe_rollout(std::vector<const mjModel*>& m, mjData* d, int nroll, int n
...
+    auto task = [=, &m, &d](void) {
+      int id = pool->WorkerId();
+      _unsafe_rollout(m, d[id], j*chunk_size, (j+1)*chunk_size,
+        nstep, control_spec, state0, warmstart0, control, state, sensordata);
+    };
+    pool->Schedule(task);
+  }
+
```
yuvaltassa, line 193: // schedule any remaining jobs of size < chunk_size

python/mujoco/rollout.cc
```diff
@@ -158,6 +161,47 @@ void _unsafe_rollout(std::vector<const mjModel*>& m, mjData* d, int nroll, int n
...
+  if (chunk_remainder > 0) {
+    auto task = [=, &m, &d](void) {
+      _unsafe_rollout(m, d[pool->WorkerId()], nfulljobs*chunk_size, nfulljobs*chunk_size+chunk_remainder,
+        nstep, control_spec, state0, warmstart0, control, state, sensordata);
+    };
+    pool->Schedule(task);
+  }
+
+  pool->WaitCount(njobs);
```
yuvaltassa, line 202: // increment job counter for this thread

python/mujoco/rollout.cc
```diff
@@ -230,9 +280,24 @@ PYBIND11_MODULE(_rollout, pymodule) {
           py::gil_scoped_release no_gil;
 
           // call unsafe rollout function
-          InterceptMjErrors(_unsafe_rollout)(
-              model_ptrs, data, nroll, nstep, control_spec, state0_ptr,
-              warmstart0_ptr, control_ptr, state_ptr, sensordata_ptr);
+          if (nthread > 1 && nroll > 1) {
```
yuvaltassa, line 283: // call unsafe rollout function, multi or single threaded

python/mujoco/rollout.py
```diff
@@ -59,6 +60,8 @@ def rollout(
       (nroll x nstep x nstate)
     sensordata: Sensor data output array (optional).
       (nroll x nstep x nsensordata)
+    chunk_size: Determines threadpool chunk size. If unspecified,
+                chunk_size = max(1, nroll / (nthread * 10)
```
yuvaltassa, line 64: missing closing paren

yuvaltassa (commented):
Initial cosmetic review

python/mujoco/rollout.cc
nimrod-gileadi, line 175: Ironically, the management of this pool object is not thread safe.

i.e. if the rollout function is called from multiple threads, and we drop the Python GIL (https://github.com/google-deepmind/mujoco/blob/main/python/mujoco/rollout.cc#L230), then you may end up deleting a thread pool that is being used, or creating two threadpools.

I would lean towards creating a new threadpool on every call to rollout. Have you measured a serious performance issue with that?

yuvaltassa: @nimrod-gileadi, this is clearly stated in the PR description and also discussed in the related design doc 🙂 

The short answer is that yes, there is a measurable cost to starting up the threadpool, especially for smallish tasks. @aftersomemath can perhaps provide some numbers here to guide the discussion. I think the options are:
1. Come up with a mechanism to make this threadsafe, that does not hurt performance or overcomplicate the API.
2. Clearly document that multi-threading is not thread safe.

I think option 2 is fine, even though when pressed I can think of use cases for multi-multi-threaded rollouts. Perhaps @kevinzakka and/or @btaba would have an opinion on options 1 vs 2?

aftersomemath: I just measured threadpool creation overhead for the case that the pool is recreated every call. I tested the models, hopper, go 2, and humanoid. nroll = [50, 3000, 100] while nstep = [1000, 1, 50] respectively. For hopper the overhead is 2-30%. unitree go 2 is 0.3-4%. humanoid is 0.5-9%. This seems significant, at least in my view.

Yes, the management of the pool is not threadsafe. But also the pool cannot be used from multiple threads because of the way the condition variable and WaitCount work.

Some alternatives that allow a persistent threadpool and allow rollout to be threadsafe are:

1. let the user pass a reference to the pool to rollout (requires wrapping the pool with pybind, or at least managing it)
2. make rollout a function object (easy)
3. let the user start/stop the threadpool and change the threadpools design to allow usage from multiple threads (this will result in a rather complex threadpool implementation)

I'm in favor of implementing both the current approach and, if threading-threaded-rollouts is, important, option 2. The current approach is not mutually exclusive with 2 and 2 is a breaking API change that requires the user to manage an object instead of just calling a function.

aftersomemath: Technically, it the threadpool is threadsafe, the issue is just that the pool is not compatible with multithreaded usage. My bad, I should have been more clear.

python/mujoco/rollout.cc
aftersomemath: I went with `// wait for job counter to incremented up to the number of jobs submitted by this thread`

aftersomemath:
fixed all cosmetics

python/mujoco/rollout.cc
aftersomemath: I'm ok with just option 2 as well.

yuvaltassa: Pardon my ignorance, can someone spell out Option 2 more clearly?

aftersomemath: Usage would look like this if implemented exactly as discussed above:

```python
rollout = rollout.Rollout() # creates an object Rollout that acts like a function
rollout(models, data) # Creates or reuses member variable `rollout.pool`
rollout = None # reference lost, the object will be destroyed and the thread pool shut down cleanly
```

Usage could also look like this:
```python
with Rollout() as rollout:
    rollout(models, data)
```

In either case the thread pool will be shut down properly and automatically. Currently, this does not happen and the OS has to shut down the threads when the interpreter exits.

Also threading-threaded-rollout would be possible by using multiple Rollout instances.

At first glance, `rollout.rollout()` would still exist, but it would have to start and stop its thread pool every call. The result is that otherwise functional algorithms would not be able to use persistent threadpools without an additional argument. This seems verbose.

A solution (which I am inclined towards) is to have `rollout.rollout()` create both a persistent Rollout instance and a cleanup handler for it (with [`atexit`](https://docs.python.org/3/library/atexit.html)) that runs on interpreter shutdown.

yuvaltassa: I get now, thank you. 

I'll let @nimrod-gileadi comment on the details but I will say this: the current single-threadedness and pain of using Python threads mean that this module is not highly used in the wild and when it is, it's by expert users. Ergo, it's okay to break existing behaviour. As long as the changelog includes a "breaking changes" section with migration instructions for existing code, we're good. No need to bend over backwards so existing code runs unmodified.

milutter (commented):
Thank you so much @aftersomemath & @yuvaltassa for making this happen. 🙏

I have two questions. How much would it take to make the `Threadpool` explicit / threadsafe as far as I understand the `ThreadPool` is global and not accessible via python, right? For example in `rl` one might want to use two threadpools with different number of threads in parallel. For example use `n - 4` threads for training and use `4` threads just for evaluation. Would that be possible or require too much work?

Do I understand it correctly that the `gil` is released in the python rollout function, so I can call the rollout function using a thread and do something else in python right?

aftersomemath:
Glad you like this! 🙂

It looks like multiple threadpools will be supported. In the detailed comments above, we discussed changing the current approach to allow this. Basically, rollout would become an object where each instance has its own pool. It's not hard to implement, especially given a real use case.

 The gil is released during rollouts, so yes, it can run in parallel with another python thread.

python/mujoco/rollout.cc
milutter: I only saw this discussion after I posted my comment. 

I think making the `rollout` an object is a good solution as it also gives me the option to start and stop the threadpool whenever I like.

nimrod-gileadi: The examples you posted look good, @aftersomemath. Thanks. It would be good to have some configuration options in the Rollout constructor, to choose threadpool size, whether to use a threadpool at all, etc.

@yuvaltassa, I think it's very easy to maintain the current behaviour of the rollout function, as @aftersomemath said. I wouldn't say it's "bending over backwards" and I don't see a reason to break it unnecessarily.

python/mujoco/rollout_test.py
```diff
@@ -461,7 +461,7 @@ def test_threading(self):
     model = mujoco.MjModel.from_xml_string(TEST_XML)
     nstate = mujoco.mj_stateSize(model, mujoco.mjtState.mjSTATE_FULLPHYSICS)
     num_workers = 32
-    nroll = 10000
+    nroll = 100
```
aftersomemath, line 464: I reduced the number of rollouts for all threading tests. They were taking a noticeable amount of time to run.

aftersomemath:
Rollout has been converted to a class as discussed. Additional tests and associated documentation were updated.

`rollout.rollout` is still available. By default the thread pool is not persistent, but an optional argument makes the pool persistent (and the function is no longer thread safe). `rollout.rollout`'s persistent pool is shutdown cleanly on interpreter exit. It can also be shutdown manually by the user.

doc/python.rst
```diff
@@ -759,9 +759,11 @@ or
 
 .. code-block:: python
 
-  # pool shutdown on object deletion, interpreter shutdown, or call to rollout_.shutdown_pool
+  # pool shutdown on object deletion or call to rollout_.shutdown_pool
+  # to ensure clean shutdown of threads, call shutdown_pool before interpreter exit
   rollout_ = rollout.Rollout(nthread)
```
nimrod-gileadi, line 764: Please use a keyword argument to specify nthread. I can imagine we'll want to add more at some point.

doc/python.rst
```diff
@@ -759,9 +759,11 @@ or
 
 .. code-block:: python
 
-  # pool shutdown on object deletion, interpreter shutdown, or call to rollout_.shutdown_pool
+  # pool shutdown on object deletion or call to rollout_.shutdown_pool
+  # to ensure clean shutdown of threads, call shutdown_pool before interpreter exit
```
nimrod-gileadi, line 763: Please rename `shutdown_pool` to `close`, since we may want to release other resources in a future implementation, and `close()` is quite a common name for this kind of method.

python/mujoco/rollout.cc
```diff
@@ -176,7 +176,7 @@ void _unsafe_rollout_threaded(std::vector<const mjModel*>& m, std::vector<mjData
                               int nroll, int nstep, unsigned int control_spec,
                               const mjtNum* state0, const mjtNum* warmstart0,
                               const mjtNum* control, mjtNum* state, mjtNum* sensordata,
-                              std::shared_ptr<ThreadPool> pool, int chunk_size) {
+                              std::shared_ptr<ThreadPool>& pool, int chunk_size) {
```
nimrod-gileadi, line 179: Nit-picking: Since this function should modify the shared_ptr reference count, just make it take a raw `ThreadPool*` and let the caller dereference the shared_ptr.

doc/python.rst
```diff
@@ -711,18 +711,20 @@ The ``mujoco`` package contains two sub-modules: ``mujoco.rollout`` and ``mujoco
 rollout
 -------
 
-``mujoco.rollout`` shows how to add additional C/C++ functionality, exposed as a Python module via pybind11. It is
-implemented in `rollout.cc <https://github.com/google-deepmind/mujoco/blob/main/python/mujoco/rollout.cc>`__
+``mujoco.rollout`` and ``mujoco.Rollout`` shows how to add additional C/C++ functionality, exposed as a Python module
```
nimrod-gileadi, line 714: Are we exposing mujoco.Rollout, or mujoco.rollout.Rollout?

python/mujoco/rollout.cc
```diff
@@ -225,93 +233,121 @@ mjtNum* get_array_ptr(std::optional<const py::array_t<mjtNum>> arg,
   return static_cast<mjtNum*>(info.ptr);
 }
 
+class Rollout {
+  public:
+    Rollout(int nthread) : nthread_(nthread) {
+      if (this->nthread_ > 0) {
+        this->pool_ = std::shared_ptr<ThreadPool>(
```
nimrod-gileadi, line 240: Use `std::make_shared<ThreadPool>(this->nthread_);`

python/mujoco/rollout.py
```diff
@@ -22,6 +23,228 @@
 import numpy as np
 from numpy import typing as npt
 
+class Rollout:
+  def __init__(self, nthread: int = None):
```
nimrod-gileadi, line 27: Could you make nthread a keyword-only argument?

nimrod-gileadi (changes requested):

doc/python.rst
aftersomemath: `mujoco.rollout.Rollout`

aftersomemath:
All requested changes implemented. Also fixed a bug where `close` was not being called in `rollout.rollout()`.

aftersomemath:
One question is, the docstring for rollout appears twice. Once in `rollout.rollout` and again in `rollout.Rollout.rollout`. Do we want to just have one copy and let one refer to the other?

python/mujoco/rollout.py
```diff
@@ -74,132 +302,38 @@ def rollout(
...
+    control,
+    control_spec=control_spec,
+    skip_checks=skip_checks,
+    nstep=nstep,
+    initial_warmstart=initial_warmstart,
+    state=state,
+    sensordata=sensordata,
+    chunk_size=chunk_size)
+
+  if not persistent_pool:
```
nimrod-gileadi, line 332: Can you put this in a `try ... finally` block, so `close()` is called even if an exception is thrown?

nimrod-gileadi:
> One question is, the docstring for rollout appears twice. Once in `rollout.rollout` and again in `rollout.Rollout.rollout`. Do we want to just have one copy and let one refer to the other?

Having two copies is OK.

nimrod-gileadi (changes requested):
Looks great. Just a couple of fixes needed.

aftersomemath:
All changes implemented. I double checked and I think we caught all the missing `.close()` calls.

python/mujoco/rollout.py
```diff
@@ -308,29 +310,33 @@ def rollout(
...
-    nstep=nstep,
-    initial_warmstart=initial_warmstart,
-    state=state,
-    sensordata=sensordata,
-    chunk_size=chunk_size)
-
-  if not persistent_pool:
-    rollout.close()
+  try:
+    ret = rollout.rollout(
```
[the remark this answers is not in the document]
aftersomemath: Done

yuvaltassa (approved):

yuvaltassa:
Merged.

PR #2291, opened 2024-12-14
Fix MjSpec introspection with `visual.rgba`, `visual.headlight`, and `visual.global_`.

AaronYoung5:
Fixes issue mentioned here in https://github.com/google-deepmind/mujoco/discussions/2289.

### DIFF
--- a/python/mujoco/codegen/generate_spec_bindings.py
+++ b/python/mujoco/codegen/generate_spec_bindings.py
@@ -70,6 +70,19 @@ def _value_binding_code(
 
   return f'{classname}.def_property({",".join(def_property_args)});'
 
+def _struct_binding_code(
+    field: ast_nodes.AnonymousStructDecl, classname: str = '', varname: str = ''
+) -> str:
+  code = ''
+  name = classname + varname.title()
+  # explicitly generate for nested fields with arrays
+  if any(isinstance(f.type, ast_nodes.ArrayType) for f in field.fields):
+    for subfield in field.fields:
+      code += _binding_code(subfield, name)
+  # generate for the struct itself
+  field = ast_nodes.ValueType(name=name)
+  code += _value_binding_code(field, classname, varname)
+  return code
 
 def _array_binding_code(
     field: ast_nodes.ArrayType, classname: str = '', varname: str = ''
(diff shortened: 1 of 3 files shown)


yuvaltassa (commented):
What about the other members of mjVisual (map, scale, quality)?

Can you please do the CLA thing?

yuvaltassa:
Also, thank you! 😊 🙏

yuvaltassa:
Got it, makes sense.

Question: are you sure we need the underscore suffix? I.e. did you try not adding it and got an actual error? Sometimes the syntax highlighting will make you think you need it but `global` is actually OK.

AaronYoung5:
Yeah, I get the following error.

```bash
$ python -c 'import mujoco; print(mujoco.MjSpec().visual.global)'
  File "<string>", line 1
    import mujoco; print(mujoco.MjSpec().visual.global)
                                                ^^^^^^
SyntaxError: invalid syntax
```

quagla:
Great, thanks for doing this! Is there anything that prevents you to put the code inside `_value_binding_code` like I did for e.g. mjStatistic? I don't think the top-level function of the binding generation makes it easy to maintain the code. That would also make it easier to generalize it to arrays.

AaronYoung5:
Thanks for taking the time! 

I don't think `_value_binding_code` has access to the subfields, which are the array types that need to be defined explicitly, right? Like since only the `ValueType` is passed, it can't iteratively define the arrays. The latest commit might be more maintainable, as you suggested

quagla (approved):

yuvaltassa (approved):

yuvaltassa:
Can you remove the trailing spaces please?

Lines 986, 988 and 989 in specs.cc

AaronYoung5:
Done

quagla (approved):

yuvaltassa (approved):

AaronYoung5:
Hi @yuvaltassa , would it be possible for this PR to be included in the next release? I'm currently using my branch as a dependency, but would be nice to use upstream mujoco. Is there anything I need to do to fix the CI? Thanks in advance!

PR #2314, opened 2024-12-27
Energy calculations in MJX

simeon-ned:
This PR adds support for kinetic and potential energy calculations in MJX, matching MuJoCo's implementation. 

Key changes:

- Add energy_pos() and energy_vel() functions to sensor.py for calculating potential and kinetic energy respectively
- Add EnableBit.ENERGY flag support to control energy calculations

### DIFF
--- a/mjx/mujoco/mjx/_src/io.py
+++ b/mjx/mujoco/mjx/_src/io.py
@@ -363,6 +363,7 @@ def make_data(
         '_qM_sparse': (m.nM, float),
         '_qLD_sparse': (m.nM, float),
         '_qLDiagInv_sparse': (m.nv, float),
+        'energy': (2, float),
     }
 
     if not _full_compat:
--- a/mjx/mujoco/mjx/_src/io.py
+++ b/mjx/mujoco/mjx/_src/io.py
@@ -55,9 +55,9 @@ def _make_option(
     if o.solver not in set(types.SolverType):
       raise NotImplementedError(f'{mujoco.mjtSolver(o.solver)}')
 
-    for i in range(mujoco.mjtEnableBit.mjNENABLE):
-      if o.enableflags & 2**i:
-        raise NotImplementedError(f'{mujoco.mjtEnableBit(2 ** i)}')
+    # Check enable flags using enum pattern
+    if types.EnableBit(o.enableflags) not in set(types.EnableBit):
+      raise NotImplementedError(f'Unsupported enable flags: {mujoco.mjtEnableBit(o.enableflags)}')
 
   has_fluid_params = o.density > 0 or o.viscosity > 0 or o.wind.any()
   implicitfast = o.integrator == mujoco.mjtIntegrator.mjINT_IMPLICITFAST
@@ -71,6 +71,7 @@ def _make_option(
   fields['jacobian'] = types.JacobianType(o.jacobian)
   fields['solver'] = types.SolverType(o.solver)
   fields['disableflags'] = types.DisableBit(o.disableflags)
+  fields['enableflags'] = types.EnableBit(o.enableflags)
   fields['has_fluid_params'] = has_fluid_params
 
   return types.Option(**fields)
--- a/mjx/mujoco/mjx/_src/io.py
+++ b/mjx/mujoco/mjx/_src/io.py
@@ -56,8 +56,8 @@ def _make_option(
       raise NotImplementedError(f'{mujoco.mjtSolver(o.solver)}')
 
     # Check enable flags using enum pattern
-    if types.EnableBit(o.enableflags) not in set(types.EnableBit):
-      raise NotImplementedError(f'Unsupported enable flags: {mujoco.mjtEnableBit(o.enableflags)}')
+    if types.EnableBit(o.enableflags) not in set(types.EnableBit) and o.enableflags != 0:
+      raise NotImplementedError(f'{mujoco.mjtEnableBit(o.enableflags)}')
 
   has_fluid_params = o.density > 0 or o.viscosity > 0 or o.wind.any()
   implicitfast = o.integrator == mujoco.mjtIntegrator.mjINT_IMPLICITFAST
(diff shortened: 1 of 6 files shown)


erikfrey:
@simeon-ned what a nice PR!  Thank you for creating it.  Things are still slow due to the holidays, but someone will get to this soon.

simeon-ned:
Hello @erikfrey, is there anything I can do to help with this PR? I could add tests similar to those implemented in mjx src.

mjx/mujoco/mjx/_src/io.py
```diff
@@ -55,9 +55,9 @@ def _make_option(
     if o.solver not in set(types.SolverType):
       raise NotImplementedError(f'{mujoco.mjtSolver(o.solver)}')
 
-    for i in range(mujoco.mjtEnableBit.mjNENABLE):
-      if o.enableflags & 2**i:
-        raise NotImplementedError(f'{mujoco.mjtEnableBit(2 ** i)}')
+    # Check enable flags using enum pattern
+    if types.EnableBit(o.enableflags) not in set(types.EnableBit) and o.enableflags != 0:
+      raise NotImplementedError(f'{mujoco.mjtEnableBit(o.enableflags)}')
```
erikfrey, line 60: I don't think `mujoco.mjtEnableBit(o.enableflags)` works when you have multiple bits set, you will get an error like this:

```
ValueError: Invalid int value for mjtEnableBit: 3
```

mjx/mujoco/mjx/_src/types.py
```diff
@@ -1467,3 +1477,4 @@ class Data(PyTreeNode):
   _qM_sparse: jax.Array = _restricted_to('mjx')  # pylint:disable=invalid-name
   _qLD_sparse: jax.Array = _restricted_to('mjx')  # pylint:disable=invalid-name
   _qLDiagInv_sparse: jax.Array = _restricted_to('mjx')  # pylint:disable=invalid-name
+  energy: jax.Array
```
erikfrey, line 1480: We try to match the field order in mjdata.h - can you move this below time?  And move in the docstring too.

mjx/mujoco/mjx/_src/sensor.py
```diff
@@ -600,3 +602,79 @@ def _framelinacc(cvel, cacc, offset):
...
+        return d
+    
+    # Add gravitational potential energy for each body
+    if not m.opt.disableflags & DisableBit.GRAVITY:
+        for i in range(1, m.nbody):  # Skip world body
+            energy -= m.body_mass[i] * jp.dot(m.opt.gravity, d.xipos[i])
+    
+    # Add joint spring potential energy
+    if not m.opt.disableflags & DisableBit.PASSIVE:
+        for i in range(m.njnt):
```
erikfrey, line 623: I challenge you to figure out how to do this without this for loop iterating over joints.  You should find code elsewhere that does this: look for usage of `scan.flat` in `smooth.py` for some examples.  

`scan` is a bit gnarly and not well super well documented, but should be used wherever possible.

mjx/mujoco/mjx/_src/sensor.py
```diff
@@ -600,3 +602,79 @@ def _framelinacc(cvel, cacc, offset):
...
+                dif = math.quat_sub(quat, m.qpos_spring[padr:padr+4])
+                energy += 0.5 * stiffness * jp.dot(dif, dif)
+                
+            elif m.jnt_type[i] in (JointType.SLIDE, JointType.HINGE):
+                dif = d.qpos[padr] - m.qpos_spring[padr]
+                energy += 0.5 * stiffness * dif * dif
+
+    # Add tendon spring potential energy
+    if not m.opt.disableflags & DisableBit.PASSIVE:
+        for i in range(m.ntendon):
```
erikfrey, line 650: this for loop can be removed, and you should not need `scan` at all.

you can do something like:

```py
lower = m.tendon_lengthspring[0::2]
upper = m.tendon_lengthspring[1::2]
length = d.ten_length

displacement = jp.where(length > upper, upper - length, 0)
displacement = jp.where(length < lower, lower - length, displacement)
...
```

mjx/mujoco/mjx/_src/sensor.py
```diff
@@ -600,3 +602,79 @@ def _framelinacc(cvel, cacc, offset):
...
 
   return d.replace(sensordata=sensordata)
+
+
+def energy_pos(m: Model, d: Data) -> Data:
+    """Calculates position-dependent energy (potential).
+    """
+        
+    # Initialize potential energy
+    energy = jp.array(0.0)
```
erikfrey, line 612: mega nitpick: please move this variable definition to right after the if statement guard

erikfrey (changes requested):

mjx/mujoco/mjx/_src/sensor.py
```diff
@@ -600,3 +603,85 @@ def _framelinacc(cvel, cacc, offset):
...
+            if jnt_type == JointType.FREE:
+                # Position springs
+                quat = qpos[padr:padr+4] 
+                quat = math.normalize(quat)
+                dif = quat - qpos_spring[padr:padr+4]
+                energy = 0.5 * stiffness * jp.dot(dif[:3], dif[:3])
+                
+            elif jnt_type in (JointType.FREE, JointType.BALL):
+                # Convert quaternion difference to angular displacement
+                quat = qpos[padr:padr+4]
```
mattephi, line 635: Probably slicing syntax in this form is not supported, as jax requires static start/stop/step to be used with NumPy indexing syntax.

```
IndexError: Array slice indices must have static start/stop/step to be used with NumPy indexing syntax. Found slice(Traced<ShapedArray(int32[1])>with<BatchTrace> with
  val = Traced<ShapedArray(int32[1,1])>with<DynamicJaxprTrace>
  batch_dim = 0, Traced<ShapedArray(int32[1])>with<BatchTrace> with
  val = Traced<ShapedArray(int32[1,1])>with<DynamicJaxprTrace>
  batch_dim = 0, None). To index a statically sized array at a dynamic position, try lax.dynamic_slice/dynamic_update_slice (JAX does not support dynamically sized arrays within JIT compiled functions).
```

This happens when I try to jit compile functions which take derivatives of energies.

mattephi (changes requested):

PR #2322, opened 2025-01-02
Documentation: added instructions for running tests

spikedoanz:
## Description
While trying to run MuJoCo tests locally, I discovered that they fail silently when run from build/bin. After digging through the test source code, I realized this happens because many tests rely on relative paths to model files in the test/ directory. This PR adds documentation to help other developers avoid the same confusion I encountered.

### Documentation changes
Add a new Testing section that explains:
- Tests must be run from the test/ directory (this isn't obvious and took me some investigation to figure out!)
- How to correctly run tests using relative paths to build/bin
- Example commands that actually work

### Motivation
I spent some time figuring out why my tests were failing with no helpful error messages. Rather than having other developers go through the same process of reading test source code to understand the path requirements, we should document this clearly.

The change is small but hopefully saves others the troubleshooting time I spent!

### DIFF
--- a/src/engine/engine_util_sparse.c
+++ b/src/engine/engine_util_sparse.c
@@ -742,7 +742,8 @@ void mju_sqrMatTDSparse(mjtNum* res, const mjtNum* mat, const mjtNum* matT,
   int* markers = mjSTACKALLOC(d, nc, int);
 
   for (int i=0; i < nc; i++) {
-    int* cols = res_colind+res_rowadr[i];
+    int rowadr_i = res_rowadr[i];
+    int* cols = res_colind + rowadr_i;
 
     res_rownnz[i] = 0;
     buffer[i] = 0;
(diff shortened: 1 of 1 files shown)


traversaro:
Are you familiar with ctest? This is the command from cmake meant to run tests from the build directory, that can be used also to run single tests using the -R option.

traversaro:
See also in  https://github.com/google-deepmind/mujoco/tree/main/.github%2Fworkflows to see how ctest is used to run test in CI.

spikedoanz:
Ah thank you for pointing this out! You're right - I totally missed `ctest`. This is a much better solution than my manual approach.

Would it make sense to modify the documentation to:
1. Primarily recommend using `ctest` from the build directory (since this is the proper/intended way)
2. Keep the manual test execution instructions as a "Note" or alternative approach, since understanding the test directory requirement can still be helpful for debugging?

The new version could also mention the `-R` flag for running specific tests, since that's quite useful for development.

Let me know if you'd like me to update the PR with these changes!

PR #2348, opened 2025-01-10
Fix contact frame inconsistency in visualization with mjContact.frame

### LINKED ISSUE #2335: Inconsistency between contact frame and its visualization
### Intro

Hi!

I am a graduate student at HKU, I use MuJoCo for my research on robotic manipulation.

### My setup

MuJoCo 3.2.6, Python and C, Linux

Mujoco is using the X axis as the contact normal direction in `mjContact.frame`, as explicitly stated in the documentation:
> Thus, given that MuJoCo uses row-major format, the contact normal axis (which is the X axis of the contact frame by our convention) is in position mjContact.frame[0-2], the Y axis is in [3-5] and the Z axis is in [6-8]. The reason for this arrangement is because ...

However, when visualizing with `mjFRAME_CONTACT`, the normal direction becomes the Z axis in blue, as explicitly switched in https://github.com/google-deepmind/mujoco/blob/4d82ab57626f2036e6b7b928628456264f88fea7/src/engine/engine_vis_visualize.c#L134-L137

Both conventions are perfectly fine for me, but the inconsistency may create confusion during debugging, especially when dealing with anisotropic friction in the way recommended by documentation:
> For a predefined contact pair we know the two geom types in advance, and the corresponding collision function always generates contact frames oriented in the same way – which we do not describe here but it can be seen in the visualizer.

### Steps for reproduction

-

_No response_

_No response_

### Confirmations

- [X] I searched the [latest documentation](https://mujoco.readthedocs.io/en/latest/overview.html) thoroughly before posting.
- [X] I searched previous [Issues](https://github.com/google-deepmind/mujoco/issues) and [Discussions](https://github.com/google-deepmind/mujoco/discussions), I am certain this has not been raised before.

jjyyxx:
Fixes #2335

### DIFF
--- a/src/engine/engine_vis_visualize.c
+++ b/src/engine/engine_vis_visualize.c
@@ -115,7 +115,7 @@ static void makeTriangle(mjvGeom* thisgeom, const mjtNum v0[3], const mjtNum v1[
 static void addContactGeom(const mjModel* m, mjData* d, const mjtByte* flags,
                            const mjvOption* vopt, mjvScene* scn) {
   int objtype = mjOBJ_UNKNOWN, category = mjCAT_DECOR;
-  mjtNum mat[9], tmp[9], vec[3], frc[3], confrc[6], axis[3];
+  mjtNum mat[9], vec[3], frc[3], confrc[6], axis[3];
   mjtNum framewidth, framelength, scl = m->stat.meansize;
   mjContact* con;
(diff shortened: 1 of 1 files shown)


yuvaltassa (approved):

yuvaltassa:
Note I modified your code

PR #2439, opened 2025-02-20
rollout use size_t for array size check and pointer arithmetic

aftersomemath:
`rollout.cc` used `int` types for pointer arithmetic and checking the array size. This limited the state/sensordata arrays to about 2^31 - 1 elements due to an `int` limited size check in `rollout.cc`. The fix is to use `size_t` in the appropriate places.

I wrote a test that initially failed and now passes. But if the test passes, it means a very long rollout was run. With 24 threads the test took just over 3 minutes. So I did not add the test to `rollout_test.py`. The test function is:

```python
  def test_large_state(self):
    model = mujoco.MjModel.from_xml_string(TEST_XML)
    nstate = mujoco.mj_stateSize(model, mujoco.mjtState.mjSTATE_FULLPHYSICS)
    data = mujoco.MjData(model)

    nthread = 24
    nbatch = nthread

    nstep = ((2**31) // (nstate*nbatch)) + 2
    assert nstep * nstate * nbatch > 2**31

    initial_state = np.random.randn(nbatch, nstate)
    rollout.rollout(model, [copy.copy(data) for _ in range(nthread)],
                    initial_state, nstep=nstep)
```

### DIFF
--- a/python/mujoco/rollout.cc
+++ b/python/mujoco/rollout.cc
@@ -75,10 +75,11 @@ void _unsafe_rollout(std::vector<const mjModel*>& m, mjData* d, int start_roll,
                      const mjtNum* state0, const mjtNum* warmstart0,
                      const mjtNum* control, mjtNum* state, mjtNum* sensordata) {
   // sizes
-  int nstate = mj_stateSize(m[0], mjSTATE_FULLPHYSICS);
-  int ncontrol = mj_stateSize(m[0], control_spec);
-  int nv = m[0]->nv, nbody = m[0]->nbody, neq = m[0]->neq;
-  int nsensordata = m[0]->nsensordata;
+  size_t nstate = static_cast<size_t>(mj_stateSize(m[0], mjSTATE_FULLPHYSICS));
+  size_t ncontrol = static_cast<size_t>(mj_stateSize(m[0], control_spec));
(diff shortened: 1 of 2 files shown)


yuvaltassa:
@nimrod-gileadi is there a way to add a **disabled test**?

One very ugly way is to add it in a comment with `# uncomment below to test for large sizes`. If we add a return at the top the linter will complain about unreachable code. Is there an "official" way to add a test that you don't actually want to be run automatically? Like `notap` but in public code...

aftersomemath:
I pushed the test with a skip decorator.

nimrod-gileadi (approved):

python/mujoco/rollout.cc
```diff
@@ -226,7 +227,8 @@ mjtNum* get_array_ptr(std::optional<const py::array_t<mjtNum>> arg,
   py::buffer_info info = arg->request();
 
   // check size
-  int expected_size = nbatch * nstep * dim;
+  size_t expected_size =
```
yuvaltassa, line 230: Trailing whitespace on this line, please fix

aftersomemath:
fFixed cosmetics in `rollout.cc` and ran isort/pyink on `rollout_test.py`.

nimrod-gileadi (approved):

PR #2493, opened 2025-03-11
Allow run/pause toggling from passive UI and handle, closes #2481

### LINKED ISSUE #2481: Enable pausing in passive viewer

While the **passive viewer** is more flexible than the **managed viewer**, it lacks several convenient features (e.g., reload, history) due to its passive nature. Among the current restrictions, the inability to **pause** via GUI or <kbd>Space</kbd> seems like a limitation that could be lifted without extensive refactoring.  

### **Proposed Solution**  
When the user requests a pause, the program could check the `run` flag and simply stop calling `mj_step`, effectively pausing the simulation. This would enable **direct `qpos` manipulation and pose perturbation** while paused.  

### **Feasibility Check**  
I have tested this locally by manually setting `run = 0` via a hacky memory write by offset. The results indicate that:  
- The flag update works as expected.  
- The pause state can be confirmed via a Python handle.  
- The `qpos` slider becomes usable.  
- Float body perturbation functions correctly.  

### **Potential Concern**  
For a robust API implementation, synchronization is necessary since: `run` would be set in the **rendering thread**, and it would be read from the **main thread**. Maybe add an atomic variable like `exitrequest` is helpful.

### Alternatives

_No response_

### Additional context

_No response_

jjyyxx:
Closes #2481.

The actual behavior of paused simulation in passive viewers is up to the user. To work as intended, the user should check `handle.run` to choose between `mj_step` and `mj_forward`:
```python
import time
import mujoco, mujoco.viewer
model = mujoco.MjModel.from_xml_string("""
<mujoco>
    <worldbody>
        <body name="box" pos="0 0 0">
            <joint type="slide" axis="1 0 0"/>
            <geom type="box" size="1 1 1" rgba="1 0 0 1"/>
        </body>
        <body name="sphere" pos="3 0 0">
            <joint type="free"/>
            <geom type="sphere" size="1" rgba="0 1 0 1"/>
        </body>
    </worldbody>
</mujoco>
""")
data = mujoco.MjData(model)
handle = mujoco.viewer.launch_passive(model, data)
while handle.is_running():
    if handle.run:
        mujoco.mj_step(model, data)
    else:
        # Forward to reflect direct position perturb
    handle.sync()
    time.sleep(0.001)
```
When the simulation is paused, the user can apply direct qpos manipulation and pose perturbation in the simulator.

Several concerns:
1. The `Handle` class already has a method `is_running` to check if the simulation window is running or exited, while the `simulator` uses `run=1/0` to indicate run/pause. The naming would potentially cause confusion.
2. **(Off-topic)** The following line does not quite make sense to me (and opposite to several other cases above and below). Should `!sim->is_passive_` be `sim->is_passive_`?

    https://github.com/google-deepmind/mujoco/blob/14dc6fd33b9fbeb88d09bbde397c558f0f8b52d4/simulate/simulate.cc#L1620

### DIFF
--- a/python/mujoco/simulate.cc
+++ b/python/mujoco/simulate.cc
@@ -277,8 +277,10 @@ PYBIND11_MODULE(_simulate, pymodule) {
       .def_property_readonly("busywait",
                              GetIfNotNull(&mujoco::Simulate::busywait),
                              py::call_guard<py::gil_scoped_release>())
-      .def_property_readonly("run", GetIfNotNull(&mujoco::Simulate::run),
-                             py::call_guard<py::gil_scoped_release>())
+      .def_property("run",
+                    GetIfNotNull(&mujoco::Simulate::run),
+                    SetIfNotNull(&mujoco::Simulate::run),
+                    py::call_guard<py::gil_scoped_release>())
 
       .def_property_readonly("exitrequest",
                              CallIfNotNull(+[](mujoco::Simulate& sim) {
--- a/python/mujoco/viewer.py
+++ b/python/mujoco/viewer.py
@@ -115,6 +115,21 @@ class Handle:
       return sim.viewport
     return None
 
+  @property
+  def run(self):
+    sim = self._sim()
+    if sim is not None:
+      return sim.run
+    return None
+
+  @run.setter
+  def run(self, value):
+    if value != 0 and value != 1:
+      raise ValueError('run must be 0 or 1')
+    sim = self._sim()
+    with sim.lock():
+      sim.run = value
+
   def set_figures(self, viewports_figures):
     sim = self._sim()
     if sim is not None:
--- a/simulate/simulate.cc
+++ b/simulate/simulate.cc
@@ -1553,7 +1553,7 @@ void UiEvent(mjuiState* state) {
   if (state->type==mjEVENT_KEY && state->key!=0) {
     switch (state->key) {
     case ' ':                   // Mode
-      if (!sim->is_passive_ && sim->m_) {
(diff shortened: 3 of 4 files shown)


yuvaltassa:
Thanks for this! We will check it soon.

josechenf:
Hello!

I use the passive viewer from time to time, and I'm wondering if you can achieve the same with something like this (untested pseudo-code):

```
_IS_RUNNING = True
def _key_callback(key: int) -> None:
  if key == 32:  # Space bar
    _IS_RUNNING = not _IS_RUNNING
 
handle = mujoco.viewer.launch_passive(
      model,
      data,
      key_callback=_key_callback,
  )
while handle.is_running():
   with handle.lock()
      if _IS_RUNNING:
        mujoco.mj_step(model, data)
      else:
        # Forward to reflect direct position perturb
      handle.sync()
    time.sleep(0.001)
 ```

The advantage of this is that you can also create much more complex behavior.

josechenf:
I am curious about what a user should to visualize a trajectory of saved states without running `mj_step`. Is the user expected to set `handle.run = False`, and not press spacebar?

RE(2), this is currently possible by calling `mjv_applyPerturbForce` and `mjv_applyPerturbPose`.

michael-lutz:
Seems like a pretty useful feature - been looking to pause and modify joint qpos.

yuvaltassa:
Closing as the old passive viewer is basically deprecated. More info can be found here and soon in the proper docs
https://github.com/google-deepmind/mujoco/blob/main/doc/skills/studio/SKILL.md

Issue #2494, opened 2025-03-11
Exposing handles for plotting, image overlay and text boxes from simulate.cc and viewer.py


Andrew-Luo1:
### The feature, motivation and pitch

My RL workflow is Brax for its PPO, MJX for physics, and Madrona-MJX rendering for pixel observations. Especially with the detailed reward shaping for manipulation tasks and the visual domain-randomization required for sim2real transfer of pixel-based policies, visualizing rollouts throughout the training process across different envs is very helpful.

Late last year, I wrote *rscope*, a fork of Mujoco's viewer for visualizing RL training. It has become a key part of my workflow and I'd love to open source it.

https://github.com/user-attachments/assets/2785e4cc-93a8-4835-afe0-bac832d0e6cc

Rscope allows the user to inspect Brax training runs at no additional cost. Brax already punctuates the training run with *evaluator rollouts* for policy evaluation. Rscope visualizes the transitions generated from these rollouts.

**Mujoco Change Proposal**
To cleanly release rscope, I would like to implement three general mujoco viewer features that have previously been requested by other users: text boxes, plotting and image buffer overlay. I believe the following changes would be required:

`mujoco/simulate/simulate.cc`
Methods to initialize, draw and toggle the proposed assets.

`mujoco/python/mujoco/simulate.cc`
Python bindings for the new methods.

`mujoco/python/mujoco/viewer.py`
Expose the python bindings by modifying `class Handle`, which is returned by the passive viewer.

The required changes should be quite minor (<300 lines). No build changes should be needed, since these new features would simply use API's already called elsewhere in the files.

### Alternatives

**Plan B**. An alternative that would require less or no changes to Mujoco would be to essentially re-implement a simpler version of `mujoco/simulate/simulate.cc` in Python via the visualization bindings. [This repository](https://github.com/rohanpsingh/mujoco-python-viewer) provides an example of that.

While this is possible, I'm currently of the opinion that Plan A is the way to go. Plan B would likely require either importing that github repo as an additional dependency or adding ~900 lines of code. Additionally, the three proposed features can be useful to the many existing users of the passive viewer.

yuvaltassa:
This is a great proposal! 🚀 

One question is, do we first need to add the `mjui` structs to `python/mujoco/introspect` ?

Also, I believe @josechenf and @silviacruciani have worked on something like this internally, perhaps they have some opinions.

vikashplus:
[...]

PR #2503, opened 2025-03-17
Add handles for overlaying user text and images to the Passive Viewer

Andrew-Luo1:
See issue #2494 

Added the text and image overlay handles in the style of commit [37d759](https://github.com/google-deepmind/mujoco/commit/37d7591ce0d1424f6853dd9293eec1579c8e623f).

This supports:
[Brax PR 584](https://github.com/google/brax/pull/584)
[Playground PR 83](https://github.com/google-deepmind/mujoco_playground/pull/83)

### DIFF
--- a/python/mujoco/viewer.py
+++ b/python/mujoco/viewer.py
@@ -25,10 +25,12 @@ import threading
 import time
 from typing import Callable, List, Optional, Tuple, Union
 import weakref
+
 import glfw
+import numpy as np
+
 import mujoco
 from mujoco import _simulate
-import numpy as np
 
 if not glfw._glfw:  # pylint: disable=protected-access
   raise RuntimeError('GLFW dynamic library handle is not available')
@@ -114,7 +116,9 @@ class Handle:
       return sim.viewport
     return None
 
-  def set_figures(self, viewports_figures: List[Tuple[mujoco.MjrRect, mujoco.MjvFigure]]):
+  def set_figures(
+      self, viewports_figures: List[Tuple[mujoco.MjrRect, mujoco.MjvFigure]]
+  ):
     sim = self._sim()
     if sim is not None:
       sim.set_figures(viewports_figures)
@@ -125,8 +129,8 @@ class Handle:
       sim.clear_figures()
 
   def overlay_text(self, overlay_texts: List[Tuple[int, int, str, str]]):
-    """ Overlay text on the viewer.
-    
+    """Overlay text on the viewer.
+
     Args:
       overlay_texts: List of tuples of (font, gridpos, text1, text2)
       let:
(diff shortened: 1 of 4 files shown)


simulate/simulate.cc
```diff
@@ -2597,6 +2605,16 @@ void Simulate::Render() {
     ShowFigure(this, viewport, &figure);
```
aftersomemath, line 2605: These calls to `ShowFigure`, `ShowOverlayText` and `ShowImage` are being made from Simulate's `Render` function. `Render` is called from the `RenderLoop` after unlocking Simulate's lock (this helps with parallelism and makes the viewer run smoother).

However, `ShowFigure`, `ShowOverlayText` and `ShowImage` use vectors, `user_figures_`, `user_text_`, and `user_images_` which can be modified concurrently by the user code. The result is that undefined behavior can occur.

See top level review comment about how we can resolve this issue in general.

python/mujoco/viewer.py
```diff
@@ -125,6 +128,49 @@ def clear_figures(self):
...
+    if sim is not None:
+      sim.clear_overlay_text()
+
+  def set_images(
+      self, viewports_images: List[Tuple[mujoco.MjrRect, np.ndarray]]
+  ):
+    sim = self._sim()
+    if sim is not None:
+      # Nearest neighbor resize
+      resize = lambda a, s: a[(np.arange(s[0]) * a.shape[0]) // s[0]][
```
aftersomemath, line 157: Does resizing by slicing have good performance? It might be a good idea to check. If not, I would just force the user to pass in images of the correct size.

Andrew-Luo1: TODO: will check if resize > 1m fps, else check sizes match

Andrew-Luo1: Turns out it's only about 7k FPS. Will remove

josechenf:
We'd be super keen to get this in and use this!
@aftersomemath Let us know if this works for you!

python/mujoco/viewer.py
```diff
@@ -115,16 +116,96 @@ def viewport(self):
...
+      if isinstance(viewports_figures, tuple):
+        viewports_figures = [viewports_figures]
       sim.set_figures(viewports_figures)
 
   def clear_figures(self):
     sim = self._sim()
     if sim is not None:
       sim.clear_figures()
 
+  def set_text(self, overlay_texts: Union[Tuple[Optional[int], Optional[int], Optional[str], Optional[str]],
```
aftersomemath, line 142: nit: if the other functions are plural (`set_images`, `set_figures`) then this should be `set_texts`.

python/mujoco/viewer.py
```diff
@@ -115,16 +116,96 @@ def viewport(self):
...
+    if sim is not None:
+      # Convert single tuple to list if needed
+      if isinstance(viewports_images, tuple):
+        viewports_images = [viewports_images]
+
+      processed_images = []
+      for viewport, image in viewports_images:
+        targ_shape = (viewport.height, viewport.width)
+        # Check if image is already the correct shape
+        if image.shape[:2] != targ_shape:
```
aftersomemath, line 197: Since we are no longer resizing the image, passing a viewport with the image is no longer required, the size is implicit and this check is not needed. However, I'm ok with leaving the check here to avoid breaking the API in the future if resizing is brought back somehow.

python/mujoco/viewer.py
```diff
@@ -115,16 +116,96 @@ def viewport(self):
...
     sim = self._sim()
     if sim is not None:
       sim.clear_figures()
 
+  def set_text(self, overlay_texts: Union[Tuple[Optional[int], Optional[int], Optional[str], Optional[str]], 
+                                            List[Tuple[Optional[int], Optional[int], Optional[str], Optional[str]]]]):
+    """Overlay text on the viewer.
+
+    Args:
+      overlay_texts: Single tuple or list of tuples of (font, gridpos, text1, text2)
```
aftersomemath, line 147: nit: since `set_figures` argument is `viewports_figures`, this argument should be `fonts_gridpos_texts`, but that is cumbersome, how about `texts`? We don't need to include the `overlay` prefix.

aftersomemath (approved):
Looks good to me. My comments are just nits.

Thanks for adding a docstring, typing, etc to set_figures.

I think merging this with the race condition is fine because `set_figures` is already on main with the race condition. I will submit a PR for fixing the race condition in a few days.

After reviewing this a second time, I came to the conclusion that signatures of the `set_X` functions should be more pythonic (maybe each argument should be its own list like with `rollout`). But we don't need to fix that in this PR.

Andrew-Luo1:
Updated!

yuvaltassa (approved):

yuvaltassa (approved):

Issue #2526, opened 2025-03-24
Systematic asset reuse for attached models


jjyyxx:
### The feature, motivation and pitch

`asset/model` + `<attach>` and the `mjs_attachBody`, `mjs_attachFrame` equivalent in model editing API is the most systematic and convenient way to combine components into a complex scene. But a potential limitation is that: when a model is attached multiple times, or in any other case that assets (mainly meshes) are effectively shared by multiple child models, the assets are unconditionally duplicated in the main model.

In our use case, the model size quickly grows to over 300MiB with over 1M mesh faces. The major sources of bloating are several child models with SDF collision geom (SDF marching cube resolution is hard-coded) attached several times. If we further scale up the scene (only by attaching the existing child models more times), the model size would surely exceed 1GiB. The bloating model size would not cause significant problem for "single model, multiple data" simulation on its own, but:
1. We are trying to store MJB binary for archiving and exact reproducibility;
2. We plan to experiment with certain domain randomization that goes beyond the scope of MjData, necessitating one model for each parallel worker.

### Alternatives

1. Refactor the namespacing/scoping mechanism. The current prefix/suffix mechanism is somewhat inflexible.
2. Centralized asset management and asset sharing across models: It is theoretically possible with current mjModel interface, by changing asset pointers currently pointing to the main buffer to external buffers.

yuvaltassa:
This is a valuable feature request, thanks.

I think there are two separate but related requests:
1. Stop copying assets unconditionally during attach, only copy assets which are actually referenced in the child.
2. Detect duplicate assets and merge them. This is obviously valuable during attach, but could potentially happen much later in the compilation step, for any model, whether created using attach or not.

Is this a correct reflection of your request, or did I miss something?

demoncoder-crypto:
I have done a very rough implementation of this, let me know if it works, i will refine it iteratively

demoncoder-crypto:
#2540 Find it here

PR #2540, opened 2025-03-28
Systematic asset reuse for attached models #2526

demoncoder-crypto:
# Asset Deduplication for Model Attachment

 Changes:
- Enhanced the `FindDuplicateAsset` method to perform thorough comparisons of different asset types:
  - For meshes: Compares file paths and buffer sizes
  - For textures: Compares file paths and dimensions
  - For height fields: Compares file paths and dimensions
  - For skins: Compares vertex and face counts
- Updated the `AppendSpec` method to check for existing assets before appending new ones

This is a very rough Implementation of the issue let me know if its ok to be accepted


src/user/user_objects.cc
quagla, line 1: Something weird happened to this file. It looks like the code is repeated many times?

quagla (changes requested):

src/user/user_objects.cc
demoncoder-crypto: When I stashed git changes it said no changes registered added a case deleted the code tried to commit again user changes have not been stashed try merge head and after that I don't what the code file copy pasted itself. I am so sorry about this mess-up i was non intentional. I will fix this in next commit

src/user/user_model.cc
```diff
@@ -1166,10 +1166,101 @@ mjCPlugin* mjCModel::AddPlugin() {
...
+    mj_deleteSpec(spec);
+    return;
+  }
+  
+  // No duplicate found, add the new spec
   specs_.push_back(spec);
 }
 
+// Find duplicate assets and return the existing one if found
+mjSpec* mjCModel::FindDuplicateAsset(const mjSpec* spec) {
```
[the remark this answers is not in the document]
demoncoder-crypto: For sure i will do it very soon

yuvaltassa:
Comparison by metadata is not enough. You can exit early if the metadata is different, but if it's the same then you have to compare the content. Adding @kbayes who has some related infra in the asset cache.

demoncoder-crypto:
You are absolutely right as the current asset deduplication check in FindDuplicateAsset only compares metadata like file paths and dimensions. To make it more robust, you are suggesting as i understand to check if the metadata matches, then perform a byte-by-byte comparison of the actual asset content using memcmp? I can try to Implement this logic in user_model.cc. Please let me know if I am on the right track.

demoncoder-crypto:
I have tried to implement this. I am very unsure of this. Any help is appreciated

demoncoder-crypto:
@yuvaltassa any updates on this?

yuvaltassa:
Closing due to staleness, apologies. Please reopen if you like

PR #2561, opened 2025-04-08
[mjspec] user_api: add mjs_findAllBodies + mjs_findAllElements

Tadinu:
#2556
This PR adds util functions that find all child specs (of elements, bodies, geoms, etc.) in either a `mjSpec` or a `mjsBody`, largely based on `FindAllImpl()` in specs.cc. Specifically:
Private API:
```
template <typename TSpec, typename TChildSpec,
          typename = std::enable_if<std::is_same_v<TSpec, mjSpec> | std::is_same_v<TSpec, mjsBody>>>
std::vector<TChildSpec*> _mjPRIVATE__findAll(TSpec* base_spec, mjtObj type, int recurse)
```
Public API:
* `MJAPI mjsBodyVec mjs_findAllBodies(mjSpec* s, int recurse)` with `using mjsBodyVec = std::vector<mjsBody*>;`
* `MJAPI mjsElementVec mjs_findAllElements(mjSpec* s, mjtObj type, int recurse);` with `using mjsElementVec = std::vector<mjsElement*>;`
-> This requires `-Wno-return-type-c-linkage` to compile in Clang for `mjsBodyVec, mjsElementVec`

Notes:
* Further API, finding specific child specs (bodies, geoms, sites, etc.) of a body spec, can be also added eg:
```
std::vector<mjsGeom*> mjs_findAllChildGeoms(mjsBody* body, int recurse) {
  return _mjPRIVATE__findAll<mjsBody, mjsGeom>(body, mjOBJ_GEOM, recurse);
}
```
, however, is left for the future to keep this one usefully minimal
* A follow-up PR on python binding can utilize `mjs_findAllElements()` in `FindAllImpl()`


yuvaltassa:
@Tadinu we discussed internally and the requirement of adding new pointer array types to the API makes the PR a bit too cumbersome and is not justified by the functionality. 

We think the user who wants this can basically copy your implementation.

I hope that's not too disappointing.

Tadinu:
@yuvaltassa Oh sure, it's understandable. I actually had the same thought as making the PR, also realizing that point. Maybe that's why we didn't have `FindAllImpl` in C in the first place.

Issue #2596, opened 2025-04-24
Possible incorrect accelerations from accelerometer


varunagrawal:
### Intro

Hi!

I am a graduate student at Georgia Tech, where I use MuJoCo for my research on legged robot state estimation.

### My setup

I am using Mujoco Playground to train a model for Go1 walking and then use the model to collect a rollout as a trajectory.
Since Playground already attaches IMUs and other sensors to the Go1, this made it easy to collect ground truth IMU measurements.
```python
env.get_accelerometer(state.data)
```
Additionally, a velocimeter is defined on the Go1 so I can get the ground truth linear velocity in the IMU frame:
```python
env.get_local_linvel(state.data)
```

In the [Go1 xml](https://github.com/google-deepmind/mujoco_playground/blob/024adbeb301457f2ee0ab6e014c3b54f5696fa5f/mujoco_playground/_src/locomotion/g1/xmls/g1_mjx_feetonly.xml), the integration scheme is specified as `Euler`.

The trained model looks good in that the Go1 is consistently walking with faltering or falling down. Please see the video below for a qualitative result.

https://github.com/user-attachments/assets/c2982381-f3c9-4bb0-82f1-a55bb3afc90d

The code I use for computing the rollout is similar to the one in the [Locomotion jupyter notebook](https://colab.research.google.com/github/google-deepmind/mujoco_playground/blob/main/learning/notebooks/locomotion.ipynb) on Playground's website.

Below is the code I use. Please focus particularly on the `get_rollout` function.

```python
"""Showcase accelerometer bug on Go1"""

# pylint: disable=protected-access

import functools
from typing import Callable, Tuple

import jax
from brax.io import model
from brax.training.agents.ppo import networks as ppo_networks
from brax.training.agents.ppo import train as ppo
from jax import numpy as jp
from mujoco_playground import registry, wrapper
from mujoco_playground.config import locomotion_params
from tqdm import tqdm


def sample_pert(rng, env, state, velocity_kick_range, kick_duration_range):
    """Sample and add perturbations to the state."""
    rng, key1, key2 = jax.random.split(rng, 3)
    pert_mag = jax.random.uniform(key1,
                                  minval=velocity_kick_range[0],
                                  maxval=velocity_kick_range[1])
    duration_seconds = jax.random.uniform(key2,
                                          minval=kick_duration_range[0],
                                          maxval=kick_duration_range[1])
    duration_steps = jp.round(duration_seconds / env.dt).astype(jp.int32)
[...]

PR #2611, opened 2025-05-06
Add note to docs regarding sensor information

varunagrawal:
I encountered an issue when using Mujoco Playground, where the simulation would be run `n` times between each control input, where `n = ctrl_dt / sim_dt`.

This made the sensor data obtained a bit confusing, and there was no mention of this as a potential issue in any of the docs.
Hence I made a note with this PR.

The original issue which led me to this was #2596.

### DIFF
--- a/doc/XMLreference.rst
+++ b/doc/XMLreference.rst
@@ -5941,6 +5941,7 @@ corresponding to body accelerations and interaction forces. Some of these quanti
 certain sensors (force, acceleration etc.) but even if no such sensors are defined in the model, these quantities
 themselves are "features" that could be of interest to the user.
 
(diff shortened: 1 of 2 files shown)


kevinzakka:
I honestly don't think this should be clarified in the docs. There is no mention of control frequency in the MuJoCo docs, (nor should there be).

yuvaltassa:
I think this is fully covered [here](https://mujoco.readthedocs.io/en/latest/computation/index.html#consistency-in-mjdata): "After an mj_step, which terminates immediately after updating the state, quantities in mjData correspond to the previous state".

In any case, the note that you added is not really about MuJoCo, but about someone calling MuJoCo `n` times and being surprised. I.e., this is in a sense "outside" of MuJoCo itself.

varunagrawal:
Fair points. Thank you for the comments, Kevin and Yuval. I'll go ahead and close this.

PR #2645, opened 2025-05-21
Corrected A=J*inv(M)*J' in engine_setconst.c

sunilgora:
There is an error in the computation of A=J*inv(M)*J' 
```
        // inverse spatial inertia: A = J*inv(M)*J'
        mj_jacBodyCom(m, d, jac, jac+3*nv, i);
        mj_solveM(m, d, tmp, jac, 6);
        mju_mulMatMatT(A, jac, tmp, 6, nv, 6); 
```
This calculates $A=JJ^T({M^{-1}})^T$, which is incorrect. 
I did the corrections in the code to compute $A=J M^{-1}J^T$.

### DIFF
--- a/src/engine/engine_setconst.c
+++ b/src/engine/engine_setconst.c
@@ -66,6 +66,7 @@ static void set0(mjModel* m, mjData* d) {
   mjtNum A[36] = {0}, pos[3], quat[4];
   mj_markStack(d);
   mjtNum* jac = mjSTACKALLOC(d, 6*nv, mjtNum);
+  mjtNum* jacT = mjSTACKALLOC(d, 6*nv, mjtNum);
   mjtNum* tmp = mjSTACKALLOC(d, 6*nv, mjtNum);
   mjtNum* moment = mjSTACKALLOC(d, nv, mjtNum);
   int* cammode = 0;
@@ -148,8 +149,9 @@ static void set0(mjModel* m, mjData* d) {
       if (nv) {
         // inverse spatial inertia: A = J*inv(M)*J'
         mj_jacBodyCom(m, d, jac, jac+3*nv, i);
-        mj_solveM(m, d, tmp, jac, 6);
-        mju_mulMatMatT(A, jac, tmp, 6, nv, 6);
+        mju_transpose(jacT, jac, 6, 6);
+        mj_solveM(m, d, tmp, jacT, 6);
+        mju_mulMatMat(A, jac, tmp, 6, nv, 6);
       }
 
       // average diagonal and assign


src/engine/engine_setconst.c
```diff
@@ -148,8 +149,9 @@ static void set0(mjModel* m, mjData* d) {
       if (nv) {
         // inverse spatial inertia: A = J*inv(M)*J'
         mj_jacBodyCom(m, d, jac, jac+3*nv, i);
-        mj_solveM(m, d, tmp, jac, 6);
-        mju_mulMatMatT(A, jac, tmp, 6, nv, 6);
+        mju_transpose(jacT, jac, 6, 6);
```
yuvaltassa, line 152: this should be `mju_transpose(jacT, jac, 6, nv);`

src/engine/engine_setconst.c
```diff
@@ -148,8 +149,9 @@ static void set0(mjModel* m, mjData* d) {
...
+        mj_solveM(m, d, tmp, jacT, 6);
```
yuvaltassa, line 153: This is also wrong.

yuvaltassa (commented):
I think you correctly identified a bug (thanks!), but your solution needs work.
I'll try to follow up myself, this is pretty straighforward

yuvaltassa:
Okay, I think I understand where your confusion comes from. The documentation of [mj_solveM](https://mujoco.readthedocs.io/en/latest/APIreference/APIfunctions.html#mj-solvem) is written in column-major form, since that is the standard in math notation. However, solveM notionally operates on a set of vectors, which in C conventions are row vectors. If operating on a single vector $y \in R^n$ it's fine to say that we compute $x = M^{-1} y$ and we don't care if $x, y$ are column or row vectors (notionally $x^T = y^T M^{-1}$). Once we have $k$ vectors this becomes confusing. So if $y$ is $k$ vectors then in MuJoCo they are in row major form $y \in R^{k \times n}$, one $n$ vector after another in memory, $k$ times. In this case it would be more accurate, in terms of the C code, to say that solveM computes $x = y M^{-1}$ rather than $x = M^{-1} y$, but this is confusing in another way since math notation is column major. TBH I never liked row-majorness for exactly this reason, but hey, you get used to it.

yuvaltassa:
Actually, even my "fix" is not doing anything, since A is anyway symmetric.
```
BEFORE  A = J * tmp'  = J * inv(M) * J'
AFTER   A = tmp * J'  = J * inv(M) * J'
```

sunilgora:
Thanks for the clarification. This makes perfect sense if I look at the `mj_solveM(m, d, tmp, jac, 6);` with size of matrices $M$ and $J$. 

Also, I am curious if only the sum of the diagonal elements of A are required (why is it so I am still not sure!) then the calculation of remaining elements is redundant I guess.
This `mju_mulMatMatT(A, jac, tmp, 6, nv, 6);` can be made efficient by calculating diagA instead of A matrix for all bodies.

Do you (dis)agree?

yuvaltassa:
This could be made more efficient yes, but there would be little benefit. The functions in engine_setconst are called once at the end of the compilation phase, they are not in the runtime code path.

PR #2660, opened 2025-06-01
Added ellipsoid fluid model for MJX (w/o implicitfast)

vishnusangli:
## Description
We have implemented Mujoco's ellipsoid fluid model in MJX for fast and parallelizable fluid model simulation. 

## Changes
- Implemented ellipsoid-based passive fluid forces at the geom level, including support for viscous drag, lift, and added mass effects.
- Added per-step filtering logic to ensure that a single body does not receive both box and ellipsoid forces—box forces are applied only to bodies without any geoms marked for ellipsoid interaction. (use_box_model is a (nbody, ) local variable declared in _fluid function, could become a mjx.Data attribute)
- Integrated the ellipsoid model into _fluid and updated passive to store and return qfrc_fluid separately.
- Extended unit tests to validate the full fluid force pipeline using dummy aerodynamic coefficients.

## Limitations
Ellipsoid model is not yet implemented for the mjx implicitfast integrator backend due to instability with fluid force derivatives. The implictfast integrator is stable but inexact without any force derivatives.

### DIFF
--- a/mjx/mujoco/mjx/_src/passive.py
+++ b/mjx/mujoco/mjx/_src/passive.py
@@ -29,7 +29,7 @@ from mujoco.mjx._src.types import JointType
 from mujoco.mjx._src.types import Model
 from mujoco.mjx._src.types import ModelJAX
 from mujoco.mjx._src.types import OptionJAX
-from mjx_fluid._src.types import GeomType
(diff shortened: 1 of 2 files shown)


saran-t:
Adding @btaba for MJX review and @quagla for maths review.

PR #2665, opened 2025-06-05
Declare runtime dependency on typing_extensions

layus:
It is not only a development dependency. It is needed by `python/mujoco/__init__.py` at runtime.

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -30,6 +30,7 @@ dependencies = [
     "glfw",
     "numpy",
     "pyopengl",
+    "typing_extensions",
 ]
 dynamic = ["readme", "scripts"]
 


saran-t:
@quagla the dependency is coming from TypeAlias in __init__.py, do we actually need it?

quagla:
It's required as long as we support Python 3.9. From Python 3.10 we can use `typing` (as we do internally).

yuvaltassa:
So this PR can be approved, right? @quagla 

(maybe add a comment in Init.py to delete this once Python 3.9 is no longer supported)

kwlzn:
@quagla is this something we could get merged/released soon?

SebasNorenaR:
Closing this PR since we have bumped our minimum Python requirement to 3.10, making this Python 3.9-specific fix no longer necessary. Thanks for your work on this @layus !

PR #2712, opened 2025-07-01
add MUJOCO_USE_SYSTEM_TINYXML2 CMake option

fferri:
This PR introduces a CMake option ~~TINYXML2_USE_SYSTEM_PACKAGE~~ MUJOCO_USE_SYSTEM_TINYXML2 (default OFF) to allow users to link against a system-provided installation (via find_package(tinyxml2 REQUIRED)) of tinyxml2 instead of always fetching and building MuJoCo’s bundled copy.

Motivation:

If the client program also uses tinyxml2 e.g. links to the system's version of tinyxml2, symbol clashes can occur, especially if the versions are slightly different, which leads to crashes.

### DIFF
--- a/cmake/MujocoDependencies.cmake
+++ b/cmake/MujocoDependencies.cmake
@@ -149,24 +149,30 @@ target_include_directories(
 target_compile_options(qhullstatic_r PRIVATE ${MUJOCO_MACOS_COMPILE_OPTIONS})
 target_link_options(qhullstatic_r PRIVATE ${MUJOCO_MACOS_LINK_OPTIONS})
 
-set(tinyxml2_BUILD_TESTING OFF)
-findorfetch(
-  USE_SYSTEM_PACKAGE
-  OFF
-  PACKAGE_NAME
-  tinyxml2
-  LIBRARY_NAME
-  tinyxml2
-  GIT_REPO
-  https://github.com/leethomason/tinyxml2.git
-  GIT_TAG
-  ${MUJOCO_DEP_VERSION_tinyxml2}
-  TARGETS
-  tinyxml2
-  EXCLUDE_FROM_ALL
-)
-target_compile_options(tinyxml2 PRIVATE ${MUJOCO_MACOS_COMPILE_OPTIONS})
-target_link_options(tinyxml2 PRIVATE ${MUJOCO_MACOS_LINK_OPTIONS})
+option(TINYXML2_USE_SYSTEM_PACKAGE "Use the system's tinyxml2" OFF)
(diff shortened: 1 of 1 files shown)


cmake/MujocoDependencies.cmake
```diff
@@ -149,24 +149,30 @@ target_include_directories(
...
-  https://github.com/leethomason/tinyxml2.git
-  GIT_TAG
-  ${MUJOCO_DEP_VERSION_tinyxml2}
-  TARGETS
-  tinyxml2
-  EXCLUDE_FROM_ALL
-)
-target_compile_options(tinyxml2 PRIVATE ${MUJOCO_MACOS_COMPILE_OPTIONS})
-target_link_options(tinyxml2 PRIVATE ${MUJOCO_MACOS_LINK_OPTIONS})
+option(TINYXML2_USE_SYSTEM_PACKAGE "Use the system's tinyxml2" OFF)
```
traversaro, line 152: For coherence with existing options (see https://github.com/google-deepmind/mujoco/blob/50de1b3c2c62a42b394167b853cba9813f6b99c7/simulate/cmake/SimulateDependencies.cmake#L29) could it make sense to call this `MUJOCO_USE_SYSTEM_TINYXML2` ?

cmake/MujocoDependencies.cmake
```diff
@@ -149,24 +149,30 @@ target_include_directories(
...
+if(TINYXML2_USE_SYSTEM_PACKAGE)
+  find_package(tinyxml2 REQUIRED)
+  add_library(tinyxml2 ALIAS tinyxml2::tinyxml2)
+else()
```
traversaro, line 156: Instead of having an if, could it make sense to just pass `${TINYXML2_USE_SYSTEM_PACKAGE}` as `USE_SYSTEM_PACKAGE` argument, call `target_compile_options` and `target_link_options` only if `NOT TINYXML2_USE_SYSTEM_PACKAGE`, and in the rest of the CMake code (so in https://github.com/google-deepmind/mujoco/blob/50de1b3c2c62a42b394167b853cba9813f6b99c7/CMakeLists.txt#L143, https://github.com/google-deepmind/mujoco/blob/50de1b3c2c62a42b394167b853cba9813f6b99c7/src/experimental/usd/CMakeLists.txt#L101) refer to `tinyxml2::tinyxml2` ?

traversaro (approved):
Thanks a lot! Eventually I think it make sense to use `tinyxml2::tinyxml2` everywhere, but that can be a different PR.

PR #2714, opened 2025-07-02
Add cmake install rules for engine plugins

DavidPL1:
Building from source requires manually creating the 'mujoco_plugin' directory and moving the plugin libraries from the build directory to their destination for them to be used (e.g. with simulate).

This PR simply adds install rules for all plugins and makes the target directory configurable (relative to CMAKE_INSTALL_PREFIX) by setting `MUJOCO_PLUGIN_INSTALL_DIR`. The target directory defaults to `${CMAKE_INSTALL_BINDIR}/mujoco_plugin` which is simulate's lookup dir.


traversaro:
See also https://github.com/google-deepmind/mujoco/pull/1515 .

DavidPL1:
Oh sorry, I did overlook that PR. Closing in favor of #1515

PR #2746, opened 2025-07-14
Makes busy wait in thread pool optional to prevent unnecessary high cpu loads 

DavidPL1:
Using a thread pool with $n > 1$ workers substantially raises the CPU usage. This is mainly due to the worker queue using `std::this_thread::yield` [right here](https://github.com/google-deepmind/mujoco/blob/7d5b360dea2d5a45eef1ae3e71536b5d8b5860a1/src/thread/thread_queue.h#L94C29-L94C34).
Especially in a nearly idle simulation state (e.g. when pausing in Simulate-like programs), this unnecessarily increases CPU load with $n$ cores at 100% load. 

My proposed solution is to introduce an option to toggle busy waiting, which would opt for `std::this_thread::sleep` if disabled. This would retain the option to use `yield` instead of `sleep` if speed is more important. Although, I think that sleeping should be the default fit for most use cases.

Let me know your thoughts on this

### DIFF
--- a/doc/APIreference/functions.rst
+++ b/doc/APIreference/functions.rst
@@ -2903,6 +2903,15 @@ Adds a thread pool to mjData and configures it for multi-threaded use.
 
 Enqueue a task in a thread pool.
 
+.. _mju_threadPoolSetBusyWait:
+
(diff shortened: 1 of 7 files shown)


DavidPL1:
This should be handled by the updated mjthread API

Issue #2727, opened 2025-07-08
Allow overriding already registered plugins


DavidPL1:
### The feature, motivation and pitch

I have locally modified the SDF plugin by adding more configuration options for a specific project. 
When using the python bindings, the plugin can not be loaded because a call to `mujoco.mj_loadPluginLibrary` with the modified plugin library will error due to the plugins of libsdf_plugin.so already being registered.

Using the modified plugin library actually requires removing libsdf_plugin.so from the python package's install path. (Or building and installing the python package from source, which is obviously overkill). 

I'd like the option to intentionally override an already registered plugin, maybe by using an `override_existing` bool as second argument to the `mj_loadPluginLibrary` call. WARN or INFO prints of overridden plugins could be helpful and defaulting the bool to `false` would retain the current behavior of exiting with an error in case an override is not intentional.

Rawdyrathaur:
Hi @DavidPL1 👋,

Thanks for reporting this issue! I'm Manish (new to contributing here), and I completely understand your situation. Needing to reinstall everything just to test a modified plugin sounds really frustrating!

Here's what I'm thinking for a solution:
We could add a simple `force=True` option when loading plugins:

```python
# Instead of this failing:  
mj.plugin.load_plugin_library("my_modified_plugin.so")  

# You could do:  
mj.plugin.load_plugin_library("my_modified_plugin.so", force=True)

saran-t:
@DavidPL1 can you not just register the modified plugin under a different identifier?

DavidPL1:
I guess it would, as modification of the model is necessary anyways to use the modified features

Rawdyrathaur:
@DavidPL1   I understand your point about using different identifiers.
 - using a different identifier is cleaner. I am going to implement 

```cpp
// Register modified plugin under new name
mjp_registerPluginAlias(&sdf_plugin, "my_custom_sdf");

PR #2752, opened 2025-07-16
Add override_existing flag for plugin loader

VaibhavBhagat665:
This PR adds an optional `override_existing` flag to:
- mj_loadPluginLibrary
- mj_loadAllPluginLibraries

It allows overriding plugins with the same name that are already registered.


DavidPL1:
Probably based on #2727

Please don't create PRs with LLM generated nonsense.

PR #2789, opened 2025-08-07
Add plugin install rules

joaocandre:
Default CMake configuration has no install rules for plugin libraries, a simple update is provided to install them under `lib/plugin`, and create a symbolic link to this directory under `bin` as `mujoco_plugin` to enable plugin loading by utility binaries such as `simulate` and `testspeed`.
This matches the structure in release tarball archives,  thus allows replacing a given release with a source-built alternative while preserving functionality.


traversaro:
Related PR: https://github.com/google-deepmind/mujoco/pull/1515 .

Issue #2275, opened 2024-12-06
use the python version of mujoco3.2.6 under win11, an OSError: [WinError 1114] The dynamic link library (DLL) initialization routine failed.


LJoson:
### Intro

Hi!

I am a Robotics Engineer I use MuJoCo for my project.

### My setup

mujoco 3.2.6

### What's happening? What did you expect?

When I use the python version of mujoco3.2.6 under win11, an OSError: [WinError 1114] The dynamic link library (DLL) initialization routine failed. What puzzles me is that it works normally under Linux. In the end, I can only use the 2.3.7 version of mujoco under win.

### Steps for reproduction

Just use pip install mujoco on the win11 platform, and then execute python -m mujoco.viewer to reproduce

### Confirmations

- [X] I searched the [latest documentation](https://mujoco.readthedocs.io/en/latest/overview.html) thoroughly before posting.
- [X] I searched previous [Issues](https://github.com/google-deepmind/mujoco/issues) and [Discussions](https://github.com/google-deepmind/mujoco/discussions), I am certain this has not been raised before.

LJoson:
After many attempts, I finally found that only python of mujoco3.2.6 will cause this error on the win11 platform.The only solution is to lower the version. Are there any other solutions?

yuvaltassa:
To clarify, are you reporting a regression in Windows support between version <= 3.2.5 and 3.2.6 ?

LanYangya:
win10 also appear

LJoson:
> To clarify, are you reporting a regression in Windows support between version <= 3.2.5 and 3.2.6 ?

yes

Kakcalu13:
Yes, I faced the same issue. I was developing the work on a Mac M2 and Linux using mujoco 3.2.6. However, when it came to Windows 11, I encountered the exact same issue described in the title.  

I was forced to use version 3.2.3 on Windows 11 since 3.2.3 also worked on Mac and Linux


I created a simple pip that allows anyone to use their own XML or my default XML to enable FEAGI to take control. This means a lot of people (hopefully) will try it out, including non-tech-savvy users. So this is likely to be an issue for Windows users only

saran-t:
Could you please confirm your Python version?

LJoson:
> Could you please confirm your Python version?

3.9.19

saran-t:
Thanks. Does anyone have a stack trace?

Kakcalu13:
> Could you please confirm your Python version?

[...]

PR #2819, opened 2025-08-26
Update plugin loading to specify winmode for CDLL call

shi-eric:
Currently in `__init__.py` no `winmode` is specified, which can result in DLL loading issues on Windows: https://stackoverflow.com/questions/59330863/cant-import-dll-module-in-python/64472088#64472088

Specifying `winmode=0` uses a traditional Windows DLL search order. We think this could fix issues like #2275


shi-eric:
Closing since the issue we're trying to fix persists.

PR #2850, opened 2025-09-15
Update cmake_minimum_required version

thowell:


okmatija:
Could you add comments explaining why we're doing this. I'm sure we'll forget and its quite a strange change to make so it will be confusing.

Maybe something like:

```
# tinyobjloader hasn't updated its minimum required CMake version in X years and recent versions of CMake appear to have broken backwards compatibility

traversaro:
The minimum required version can also be increased either by passing via the command line the [`CMAKE_MINIMUM_REQUIRED_VERSION`](https://cmake.org/cmake/help/latest/variable/CMAKE_MINIMUM_REQUIRED_VERSION.html) variable, perhaps that is more mantainable that adding a custom patch step?

thowell:
@okmatija added comments explaining the patches

thowell:
for context we are updating the cmake_minimum_required versions for tinyobjloader and ccd because with the ci change to macos ~~15~~ **runner with cmake version 4** doesn't seem to be compatible with the versions specified in these dependencies.

@traversaro will you submit a pr with the approach proposed in your comment? thanks!

traversaro:
> The minimum required version can also be increased either by passing via the command line the [`CMAKE_MINIMUM_REQUIRED_VERSION`](https://cmake.org/cmake/help/latest/variable/CMAKE_MINIMUM_REQUIRED_VERSION.html) variable, perhaps that is more mantainable that adding a custom patch step?

Sorry for the mixup, the correct variable is https://cmake.org/cmake/help/latest/variable/CMAKE_POLICY_VERSION_MINIMUM.html .

traversaro:
> @traversaro will you submit a pr with the approach proposed in your comment? thanks!

Done in https://github.com/google-deepmind/mujoco/pull/2851, but I did not explicitly tested it.

thowell:
updated my comment above to note that i think this is the result of macos runner using cmake version 4
https://github.com/actions/runner-images/issues/12934

thowell:
@traversaro thanks!

yuvaltassa (approved):

Issue #2707, opened 2025-06-29
Default class name missing when the spec attached to a parent without a suffix and prefix


vidurvij-apptronik:
### Intro

Hi!

I am Senior Software Engineer at Apptronik using Mujoco to support my team efforts.

### My setup

Linux/MacOS, Mujoco C++, Mujoco 3.3.4

### What's happening? What did you expect?

When the child spec has a default class and it is attached to a parent class without using any suffix and prefix, the default class doesn't have any name. This results in an incompatible xml file. 

### Steps for reproduction

1. Run the code below
2. Try loading the saved XML in the mujoco simulation application or using API.

### Minimal model for reproduction

This it the XML produced by my script

```XML

<mujoco model="MuJoCo Model">
  <compiler angle="radian"/>

  <default>
    <default class="lol"/>
    <default/>
  </default>

  <worldbody>
    <body/>
    <body/>
  </worldbody>
</mujoco>


```

### Code required for reproduction

```
#include<mujoco/mujoco.h>

using namespace std;

int main(){

    mjSpec* spec = mj_makeSpec();
    mjsBody* body = mjs_findBody(spec, "world");
    mjsBody* body2 = mjs_addBody(body, nullptr);
    mjsElement* frame = mjs_addFrame(mjs_findBody(spec, "world"), NULL)->element;

    mjsDefault* def = mjs_addDefault(spec, "lol", nullptr);

    mjSpec* specA = mj_makeSpec();
    mjsBody* bodyA = mjs_findBody(specA, "world");
    mjsBody* bodyA2 = mjs_addBody(bodyA, nullptr);

    mjsDefault* defA2 = mjs_addDefault(spec, "lol", nullptr);

    mjs_attach(frame, bodyA2->element, "", "");

    mj_compile(spec, nullptr);

    char error[1024];
    mj_saveXML(spec, "test.xml", error, sizeof(error));


    return 0;
}
```

### Confirmations

- [x] I searched the [latest documentation](https://mujoco.readthedocs.io/en/latest/overview.html) thoroughly before posting.
- [x] I searched previous [Issues](https://github.com/google-deepmind/mujoco/issues) and [Discussions](https://github.com/google-deepmind/mujoco/discussions), I am certain this has not been raised before.

ayushdebnath012:
I want to work on this issue please assign me this issue

discobot:
Traced this: the attach path copies the child's root default tree and only
[...]

PR #2890, opened 2025-10-04
Clarify mjs_addDefault usage and document correct default class naming

ayushdebnath012:
This PR updates the documentation and header comments for the mjs_addDefault function to clarify its correct usage. It now explicitly states that only one unnamed default is allowed at the top level, and all other defaults should be named by passing a non-null classname. Example usage is also provided.

This is in regard to issue https://github.com/google-deepmind/mujoco/issues/2707

Changes:

Added a descriptive comment and example usage for mjs_addDefault in:
references.h
mujoco.h
user_api.h
Clarifies that only one unnamed default is allowed at the top level.
Shows how to use named and unnamed defaults correctly.
Motivation:
This change helps prevent user errors and invalid XML generation by making the API contract explicit in the documentation and headers. It addresses confusion around unnamed defaults and ensures users follow MuJoCo’s requirements for default classes.

Checklist:

 Documentation updated
 Example usage provided
 No breaking changes to API

### DIFF
--- a/doc/includes/references.h
+++ b/doc/includes/references.h
@@ -3463,6 +3463,10 @@ mjsTuple* mjs_addTuple(mjSpec* s);
 mjsKey* mjs_addKey(mjSpec* s);
 mjsPlugin* mjs_addPlugin(mjSpec* s);
 mjsDefault* mjs_addDefault(mjSpec* s, const char* classname, const mjsDefault* parent);
+// Example usage:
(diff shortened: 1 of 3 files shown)


ayushdebnath012:
How is the working code? Anything else to be done?

ayushdebnath012:
@havess can we merge this pr plz

PR #2903, opened 2025-10-17
JavaScript Bindings WIP

manevimatiasekumen:


davidhozic:
Will this be available to use without Javascript as well? It'd be nice to have web demos for things written in other languages that support compilation to WebAssembly.

PR #2913, opened 2025-10-25
Fix LTO/IPO being forcefully enabled

### LINKED ISSUE #2904: LTO is forcefully enabled instead of by default in the build system
### Intro

Hi!

I am a student (in a few days researcher) at University of Ljubljana, I use MuJoCo for my research on RL.

### My setup

MuJoCo 3.3.7. Other info is irrelevant.

A few CMake files contain forceful set of the CMake variable `CMAKE_INTERPROCEDURAL_OPTIMIZATION` (LTO).

- https://github.com/google-deepmind/mujoco/blob/36bed9a9da8cbd42a8dddd8ab3d586a51abedd7d/cmake/MujocoOptions.cmake#L107-L109
- https://github.com/google-deepmind/mujoco/blob/36bed9a9da8cbd42a8dddd8ab3d586a51abedd7d/sample/cmake/SampleOptions.cmake#L107-L109
- https://github.com/google-deepmind/mujoco/blob/36bed9a9da8cbd42a8dddd8ab3d586a51abedd7d/simulate/cmake/SimulateOptions.cmake#L107-L109

This causes problems whenever the compiled (static) libraries may be used with a different linker, which is now the case for the Rust language on the default Linux target (they switched to rust-lld).

As a result of the forceful enable, the compiled library files contain additional information that is only compatible with the same linker. See the following issue: https://github.com/rust-lang/rust/issues/146952.

Based on the commit message that added the forceful enable of LTO, the plan was to enable it by default, not permanently.

Several CI files also contain commands where this is explicitly disabled (https://github.com/google-deepmind/mujoco/blob/36bed9a9da8cbd42a8dddd8ab3d586a51abedd7d/.github/workflows/build.yml#L178), but as the result of the given lines inside CMake option files, nothing changes.

### Steps for reproduction

1. Compile with `CMAKE_INTERPROCEDURAL_OPTIMIZATION` disabled from the console
2. CMAKE_INTERPROCEDURAL_OPTIMIZATION (LTO) is still enabled

Irrelevant

Irrelevant

### Confirmations

davidhozic:
Fixes #2904 .


davidhozic:
@saran-t This works as expected now.

I'm just wondering it it's okay to raise an error (FATAL_ERROR) when `CMAKE_INTERPROCEDURAL_OPTIMIZATION:BOOL=ON` is given instead off `MUJOCO_ENABLE_IPO:BOOL=ON`?
There's unfortunately no way to check if the user has set CMAKE_INTERPROCEDURAL_OPTIMIZATION to OFF however.

stevenjj:
Thanks for this. I also needed to disable this cmake flag to link my source build of mujoco v3.6.0  to an older version of Ubuntu. 

My particular issue was that v3.6.0 needs to compile with gcc/g++ 10 but since I need to link mujoco to code that has to compile with gcc/g++ 9, it couldn't resolve LTO errors. The fix is to simply disable LTO when compiling v3.6.0

This is definitely the clean way to do it, but I was able to hardcode `set(CMAKE_INTERPROCEDURAL_OPTIMIZATION OFF)` as part of my source build. I also had to remove the warnings as error flag in the same options file. 

Hopefully this helps someone else and I do hope this gets merged.

davidhozic:
This seemed to be fixed in a different PR

Issue #2849, opened 2025-09-15
Forces and torques induced by an external magnetic field


jplayan:
Hello!

I'm a PhD student using MuJoCo for my research. I'm working on a project that requires moving a robot within a magnetic field.

For now, I've already implemented a plugin that computes and applies the forces and torques on a body given a uniform magnetic field and its gradient.

The magnetic field I'm working on is quite complex and cannot be easily described by a simple mathematical function. For this reason, I was thinking of providing the option to read the magnetic field from a precomputed 3D grid (for a region in space).

My main question would be if this is something that you would be interested in having in MuJoCo as an official plugin. I could polish the code and provide some documentation if you think this is something that could be useful for the community. If you think some extra functionality is needed, I could also implement it depending on how much load it is.

Thank you for your time and for the amazing work you've done on MuJoCo!

giusenso:
Hi @jplayan, I am planning to simulate something similar. Did you do a plugin in the end? If not, would you mind sharing your code? Thanks!

jplayan:
Hello @giusenso! 

I've just now opened a pull request with the changes (#2916). It's still a work in progress, the basic simulation with a ferromagnetic material induced by a constant field and gradient is working but I still need to add more things (such as the field generated by permanent magnets).

v-r-a:
Hello @jplayan,

In continuation of #671, I tried to simulate the magnet-magnet interaction. Here is a [video](https://youtu.be/yPfmXFyQD1g?si=Fucd9rlWMp2bOGK0) of the trials. I modelled the magnets as point dipoles (ref for interaction [force](https://onlinelibrary.wiley.com/doi/10.1155/1998/79537) and [moments](https://onlinelibrary.wiley.com/doi/10.1155/1999/97902)). I am looking at magnets of size and strength roughly similar to fridge magnets (N35, say). 

I found that the main issue with modelling magnets as dipoles arises during collisions, where the field’s singularity at the origin causes large attractive forces. These forces are multiple orders higher than the collision restoration forces, leading to unreal penetration and eventual simulation instability (the above video is paused at such instances).

Next, we decided to implement [magnetic field due to a finite cylinder](https://www.sciencedirect.com/science/article/pii/S0304885317334662?via%3Dihub). The approach is to use a distributed source to avoid singularity at the origin. Analytical expressions for the field (as provided in the paper) and its gradient have been developed. I plan to test the code soon and make it public.

[...]

PR #2916, opened 2025-10-28
Creation of plugin to simulate magnetic forces

jplayan:
Hello!

I'm a PhD student working on robotics within magnetic fields. I originally created an issue (#2849) mentioning that I had created a plugin to simulate magnetic forces.

I've already implemented some points and some others I would still like to add. I create this pull request to get feedback or for the main maintainers to give feedback on specific parts if they have time.

The main points I would like to implement are:
- [x] Wrench induced in a ferromagnetic element when inside a magnetic field
- <s>Magnetic field generated by a permanent magnet</s> (probably to be added in another pull request)
- <s>Allow loading a predefined magnetic filed (from a file such as a ".csv")</s> (too niche and not as useful)
- [x] Visualization of the wrench in each magnet

Let me know if there any other open points that could be interesting for the community or if there's anything else I should take into account.

Note: For now this is a draft as I'm still working on the code and modifying it


yuvaltassa:
Please coordinate with @v-r-a who has been working on this feature for a while

jplayan:
I've cleaned the classes and added two XMLs as examples. For now it's only supporting induced magnets and not permanent magnets, perhaps that can be added here if ready or to another merge.

As we spoke with @v-r-a the idea of having a file from which you could read the magnetic field is probably too niche and not as useful so I've removed it.

I'm not entirely certain I've correctly registered the plugin within MuJoCo as my local version seems to be unable to find it unless I specify the full path. Not sure if there is some other place where I should specify how to find it or if I'm missing a step somewhere...

Anyway, other than that, where I would appreciate some advice, I think the merge is ready to be reviewed.

Issue #2960, opened 2025-11-29
wasm demo calling an error API


ShanningZhuang:
### Intro

Hi!

I am a PhD student at Tsinghua University, I use MuJoCo for my research on Reinforcement Learning in Robotics Control.

### My setup

M4 Macbook Pro
Environment settings as in https://github.com/google-deepmind/mujoco/blob/main/wasm/README.md

### What's happening? What did you expect?

Running the demo with error
`
Initialization error:  TypeError: mujoco.MjModel.loadFromXML is not a function
    at App.loadModel (app.ts:211:35)
    at main (app.ts:489:9)
`

### Steps for reproduction

Preparing environment as in the https://github.com/google-deepmind/mujoco/blob/main/wasm/README.md
`
git clone https://github.com/emscripten-core/emsdk.git
./emsdk/emsdk install 4.0.10
./emsdk/emsdk activate 4.0.10
source ./emsdk/emsdk_env.sh
npm install --prefix ./wasm
export PATH="$(pwd)/wasm/node_modules/.bin:$PATH"
python3 -m venv .venv
source .venv/bin/activate
pip install -r python/build_requirements.txt
npm run dev:demo --prefix ./wasm
`
And opening the port in broswer

### Code required for reproduction

https://github.com/google-deepmind/mujoco/blob/69179ead4844c850129e41a8be5d23cc0ea87cfd/wasm/demo_app/app.ts#L211
I search the whole mujoco source code folder and I don't find any API in documentation as **mujoco.MjModel.loadFromXML**. I think it is deprecated, the right one should be **mujoco.MjModel.mj_loadXML**.
Or there are something I miss when preparing environment?

### Confirmations

- [x] I searched the [latest documentation](https://mujoco.readthedocs.io/en/latest/overview.html) thoroughly before posting.
- [x] I searched previous [Issues](https://github.com/google-deepmind/mujoco/issues) and [Discussions](https://github.com/google-deepmind/mujoco/discussions), I am certain this has not been raised before.

ShanningZhuang:
When I change:
mujoco.MjModel.loadFromXML
to
mujoco.MjModel.mj_loadXML
and it work without error

ShanningZhuang:
Also I found when I load a model with error it won't print the error out in the web dev tools.
https://github.com/google-deepmind/mujoco/blob/e55fff5dea6f1d5dd7963ca52eecc41d05ad0922/wasm/codegen/generated/bindings.cc#L7806
[...]

PR #2961, opened 2025-11-29
Fix incorrect method name in WASM demo app #2960

ShanningZhuang:
https://github.com/google-deepmind/mujoco/issues/2960
Replace loadFromXML with mj_loadXML to match the correct API method name in MjModel. The previous method name was incorrect and would cause the model loading to fail.

This fixes model loading in the WASM demo application.


okmatija:
Thanks for the fix! I made a commit which is attributed to you.

Also closed #2960

PR #2962, opened 2025-12-01
Enable energy for warp

aaydinoglu:
I noticed that energy is implemented in MJW but can not be enabled due to missing type. This will fix that.

### DIFF
--- a/mjx/mujoco/mjx/_src/io.py
+++ b/mjx/mujoco/mjx/_src/io.py
@@ -234,7 +234,6 @@ def _put_option(
   for i in range(mujoco.mjtEnableBit.mjNENABLE):
     if o.enableflags & 2**i and 2 ** i not in set(types.EnableBit):
       raise NotImplementedError(f'{mujoco.mjtEnableBit(2**i)}')
-  
   if o.enableflags & types.EnableBit.ENERGY and impl == types.Impl.JAX:
     raise NotImplementedError('ENERGY enable flag not implemented for JAX backend.')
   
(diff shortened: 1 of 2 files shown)


thowell:
@aaydinoglu thank you for this contribution to mjx!

i think we will also need to add the field `energy` to `mjx.Data`, will you please add this field to the dataclass in types.py?

@btaba is there anything else we need to implement in this pr in order for energy computation to work with `impl="warp"`?

aaydinoglu:
> i think we will also need to add the field energy to mjx.Data, will you please add this field to the dataclass in types.py?

Apologies for the late response. Given that jax backend does not support energy, shouldn't the energy field only exist in `mjx.Data._impl`? I also locally made a simple test **[1]**.

<br><br>

**[1] Simple script to test that energy works. We initialize a single sphere and then check:**
```
potential_energy = (mass=1) * (gravity=1) * (height=100) = 100
```
For the `mjx.Impl.WARP`, test passes. For `mjx.Impl.JAX`, we get:
```
NotImplementedError: ENERGY enable flag not implemented for JAX backend.
```
as intended.

Here is the full script for reproducability.
```
import mujoco
import mujoco.mjx as mjx

sphere_xml = """
<mujoco>
  <option gravity="0 0 -1"/>
  <worldbody>    
    <body name="sphere" pos="0 0 100.0">
      <joint type="free"/>
      <geom type="sphere" size="0.1" rgba="0 0 .9 1" mass="1"/>
    </body>
  </worldbody>
</mujoco>
"""

model = mujoco.MjModel.from_xml_string(sphere_xml)
model.opt.enableflags |= mujoco.mjtEnableBit.mjENBL_ENERGY
data = mujoco.MjData(model)

impl = mjx.Impl.WARP
mjx_model = mjx.put_model(model, impl=impl)
mjx_data = mjx.make_data(model, impl=impl, nconmax=10, njmax=10)

mjx_data = mjx.forward(mjx_model, mjx_data)

assert mjx_data.energy[0] == 100
```
Below are the versions that are releveant for running the script:
```
mujoco                   3.5.1.dev875733943
mujoco-mjx               3.5.1 
warp-lang                1.11.1
jax                      0.9.0.1
jax-cuda12-pjrt          0.9.0.1
jax-cuda12-plugin        0.9.0.1
jaxlib                   0.9.0.1
```

PR #2968, opened 2025-12-06
Add missing option 'quadratic' to dof brackets

davidhozic:
The added 'quadratic' option was only added in the body of the [``body/flexcomp/dof``](https://mujoco.readthedocs.io/en/stable/XMLreference.html#body-flexcomp) documentation.

This also adds it to the brackets of available options.


quagla (approved):

davidhozic:
Seems this was added.

Issue #2980, opened 2025-12-11
MJX: Warp FFI sometimes deadlocks when `donate_argnums` is present in `pmap`


hartikainen:
### Intro

Hey,

The following bug report is a bit limited in its details and I'm aware that the Jax-Warp backend is still very experimental. But I still wanted to report this issue to see if anyone has experienced something similar, and hopefully also save some time from others.

### My setup

`mujoco==3.4.0`, `mujoco-mjx==3.4.0`, `warp-lang==1.10.1` on `linux_x86_64` with 8xA100 GPUs.

### What's happening? What did you expect?

We use [Brax](https://github.com/google/brax) [PPO](https://github.com/google/brax/blob/a6b0c6b67841ff89f53a0667a3afaa336c11b0b0/brax/training/agents/ppo/train.py) to train RL policies. We've recently noticed that `jax.pmap` of `training_epoch` in [`ppo/train.py`](https://github.com/google/brax/blob/a6b0c6b67841ff89f53a0667a3afaa336c11b0b0/brax/training/agents/ppo/train.py#L636-L639) hangs when `donate_argnums=(0,1)`. What's annoying about this hanging is that it only happens after O(10e6) training steps.

The reason I think this is a MJX-Warp FFI issue is that the training works fine with `impl=jax` and only fails when `impl=warp`. Additionally, I've recently observed similar hanging in MJX-Warp FFI in a completely independent setting. In that setting, I was trying to access Warp memory outside of the allocated range, which deadlocked silently in a very similar way.

A simple workaround for those who might experience similar deadlocks/hangs is to remove the `donate_argnums` variable from `jax.pmap`.

### Steps for reproduction

As mentioned above, I unfortunately don't yet have a good reproduction for this. Our mujoco model is proprietary and the training setup is a bit customized, which is why I haven't been able to reproduce this with open-source models or vanilla Brax.

### Confirmations

- [x] I searched the [latest documentation](https://mujoco.readthedocs.io/en/latest/overview.html) thoroughly before posting.
- [x] I searched previous [Issues](https://github.com/google-deepmind/mujoco/issues) and [Discussions](https://github.com/google-deepmind/mujoco/discussions), I am certain this has not been raised before.

btaba:
Thanks @hartikainen , sorry for the delay. See the related comment here https://github.com/google-deepmind/mujoco/pull/3017#issuecomment-3757469633

btaba:
[...]

PR #3017, opened 2026-01-12
Fix MJX Warp FFI Deadlock in Multi-GPU

Adityakk9031:
Fix:#2980
Issue
In multi-GPU environments (e.g. using jax.pmap), identical virtual addresses can be assigned to buffers on different devices. The MJX-Warp FFI graph cache used 
(call_id, buffer_pointers)
 as the cache key. This led to collisions where a CUDA graph captured on one device was incorrectly retrieved and launched on another device, causing deadlocks or crashes (Issue #2980).

Solution
Updated 
mujoco/mjx/third_party/warp/_src/jax_experimental/ffi.py
 to include the 
device_ordinal
 in the cache key hash.

Before: key = hash((call_id, pointers...))
After: key = hash((device_ordinal, call_id, pointers...))
This ensures that CUDA graphs are cached uniquely per device, preventing cross-device launch conflicts.


btaba:
This makes sense to me, @Adityakk9031 can you make a PR to [warp-lang](https://github.com/NVIDIA/warp)? We will be removing this vendor code from MJX (it is an exact copy of what's in warp-lang anyhow)

Adityakk9031:
@btaba https://github.com/NVIDIA/warp/pull/1181 created

Adityakk9031:
@btaba any update

btaba:
Hi @Adityakk9031 you'll have to run with this locally, I plan to delete these warp-lang vendor files

Adityakk9031:
@btaba ok

PR #3044, opened 2026-01-22
Fix GIL deadlock in Python viewer's set_figures/set_images methods

kevinzakka:
The set_figures, clear_figures, set_texts, clear_texts, set_images, and clear_images methods contain spin-wait loops that held the GIL. When a key callback is registered via launch_passive, the render thread's OnKey callback tries to acquire the GIL, causing deadlock.

Fix by releasing the GIL during spin-waits, then re-acquiring it for Python object access. For methods without Python object access, add py::call_guard<py::gil_scoped_release>() to bindings.

### DIFF
--- a/python/mujoco/CMakeLists.txt
+++ b/python/mujoco/CMakeLists.txt
@@ -26,7 +26,11 @@ if(APPLE)
   enable_language(OBJCXX)
 endif()
 
-set(CMAKE_CXX_STANDARD 20)
+if(MSVC AND MSVC_VERSION GREATER_EQUAL 1927)
+  set(CMAKE_CXX_STANDARD 20) # For forceinline lambdas.
+else()
+  set(CMAKE_CXX_STANDARD 17)
+endif()
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
--- a/python/mujoco/simulate.cc
+++ b/python/mujoco/simulate.cc
@@ -17,6 +17,7 @@
 #include <memory>
 #include <stdexcept>
 #include <string>
+#include <thread>  // NOLINT(build/c++11)
 #include <tuple>
 #include <utility>
 #include <vector>
@@ -73,11 +74,10 @@ class UIAdapterWithPyCallback : public Adapter {
 // Waits for an atomic value to become the expected value, releasing the GIL
 // during the wait to prevent deadlock with render thread's key callback which
 // needs to acquire the GIL.
-void AtomicWaitNoGil(std::atomic_int& atomic, int expected) {
+void WaitForAtomicNoGil(std::atomic_int& atomic, int expected) {
   py::gil_scoped_release no_gil;
-  int value;
-  while ((value = atomic.load()) != expected) {
-    atomic.wait(value);
+  while (atomic.load() != expected) {
+    std::this_thread::yield();
   }
 }
 
@@ -113,7 +113,7 @@ class SimulateWrapper {
 
   void WaitUntilExit() {
     if (simulate_) {
-      AtomicWaitNoGil(simulate_->exitrequest, 2);
+      WaitForAtomicNoGil(simulate_->exitrequest, 2);
     }
   }
 
@@ -148,7 +148,7 @@ class SimulateWrapper {
   void SetFigures(
       const std::vector<std::pair<mjrRect, py::object>>& viewports_figures) {
     if (simulate_) {
-      AtomicWaitNoGil(simulate_->newfigurerequest, 0);
+      WaitForAtomicNoGil(simulate_->newfigurerequest, 0);
     }
 
     // Pairs of [viewport, figure], where viewport corresponds to the location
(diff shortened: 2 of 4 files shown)


aftersomemath:
Looks good, when I implemented the original "fixes" for this I did not realize the render thread was acquiring the GIL.

There is a slight risk in releasing the GIL during the spin-lock and then using the py::object's, they could have changed. However, I believe this can only occur if the user has another thread which modifies those objects, which is unlikely and poor programming anyway. Mitigating that risk would require copying the figure data twice, which is expensive, and so this fix strikes a good balance.

okmatija:
Thanks for the PR! I will try to unblock you asap. Is there a reason you didn't make the same change in all the functions you listed (set_figures, clear_figures, set_texts, clear_texts, set_images, and clear_images methods)? Also we have [migrated to C++20](https://github.com/google-deepmind/mujoco/blob/main/cmake/MujocoOptions.cmake#L18) so we could fix this while also addressing the TODOs and getting rid of the spin locks, which make me nervous for the reason Levi mentioned. Could you try the following:

Add this helper above `SimulateWrapper`

```cpp
// Waits for an atomic value to become the expected value, releasing the GIL
// during the wait to prevent deadlock with render thread's key callback which
// needs to acquire the GIL.
void AtomicWaitNoGil(std::atomic_int& atomic, int expected) {
  py::gil_scoped_release no_gil;
  int value;
  while ((value = atomic.load()) != expected) {
    atomic.wait(value);
  }
}
```

Then apply these changes, you will also no longer need to #include chrono and thread.

```cpp
  void WaitUntilExit() {
    if (simulate_) {
      AtomicWaitNoGil(simulate_->exitrequest, 2);
    }
  }
  
  [...]
  
    void SetFigures(
      const std::vector<std::pair<mjrRect, py::object>>& viewports_figures) {

    if (simulate_) {
      AtomicWaitNoGil(simulate_->newfigurerequest, 0);
    }

    // Pairs of [viewport, figure], where viewport corresponds to the location
    // of the figure on the viewer window.
    for (const auto& [viewport, figure] : viewports_figures) {
      mjvFigure casted_figure = *figure.cast<MjvFigureWrapper&>().get();
      simulate_->user_figures_new_.push_back(std::make_pair(viewport, casted_figure));
    }

    int value = 0;
    simulate_->newfigurerequest.compare_exchange_strong(value, 1);
  }

  void ClearFigures() {
    if (simulate_) {
      AtomicWaitNoGil(simulate_->newfigurerequest, 0);
    }

    simulate_->user_figures_new_.clear();

    int value = 0;
    simulate_->newfigurerequest.compare_exchange_strong(value, 1);
  }

  void SetTexts(
      const std::vector<std::tuple<int, int, std::string, std::string>>&
          texts) {
    if (simulate_) {
      AtomicWaitNoGil(simulate_->newtextrequest, 0);
    }

    // Collection of [font, gridpos, text1, text2] tuples for overlay text
    for (const auto& [font, gridpos, text1, text2] : texts) {
      simulate_->user_texts_new_.push_back(std::make_tuple(font, gridpos, text1, text2));
    }

    int value = 0;
    simulate_->newtextrequest.compare_exchange_strong(value, 1);
  }

  void ClearTexts() {
    if (simulate_) {
      AtomicWaitNoGil(simulate_->newtextrequest, 0);
    }

    simulate_->user_texts_new_.clear();

    int value = 0;
    simulate_->newtextrequest.compare_exchange_strong(value, 1);
  }

  void SetImages(
    const std::vector<std::tuple<mjrRect, pybind11::array>> viewports_images
  ) {
    if (simulate_) {
      AtomicWaitNoGil(simulate_->newimagerequest, 0);
    }

    for (const auto& [viewport, image] : viewports_images) {
      auto buf = image.request();
      if (buf.ndim != 3) {
        throw std::invalid_argument("image must have 3 dimensions (H, W, C)");
      }
      if (static_cast<int>(buf.shape[2]) != 3) {
        throw std::invalid_argument("image must have 3 channels");
      }
      if (buf.itemsize != sizeof(unsigned char)) {
        throw std::invalid_argument("image must be uint8 format");
      }

      // Calculate size of the image data
      size_t height = buf.shape[0];
      size_t width = buf.shape[1];
      size_t size = height * width * 3;

      // Make a copy of the image data since Python is
      // not required to keep it
      std::unique_ptr<unsigned char[]> image_copy(new unsigned char[size]());
      std::memcpy(image_copy.get(), buf.ptr, size);

      simulate_->user_images_new_.push_back(std::make_tuple(viewport, std::move(image_copy)));
    }

    int value = 0;
    simulate_->newimagerequest.compare_exchange_strong(value, 1);
  }

  void ClearImages() {
    if (simulate_) {
      AtomicWaitNoGil(simulate_->newimagerequest, 0);
    }

    simulate_->user_images_new_.clear();

    int value = 0;
    simulate_->newimagerequest.compare_exchange_strong(value, 1);
  }

```

kevinzakka:
Thanks @okmatija, I've addressed your comment.

kevinzakka:
Closing, done in e977b0d6609b36f1ede721d5809cc1789132357d.

Issue #3038, opened 2026-01-22
Compiling mujoco with MinGW fails because of redefined stat


maximilian-maisel-bl:
### Intro

I'm compiling Mujoco for different targets in a Linux-based CI.

### My setup

- Ubuntu 24.04 amd64
- Mujoco 3.4.0

### What's happening? What did you expect?

Building Mujoco with MinGW fails because `stat` is redefined.
```
[  9%] Building CXX object CMakeFiles/mujoco.dir/src/user/user_resource.cc.obj
/home/user/mujoco/src/user/user_resource.cc:36: error: "stat" redefined [-Werror]
   36 |   #define stat _stat
      | 
In file included from /home/user/mujoco/src/user/user_resource.cc:19:
/usr/share/mingw-w64/include/sys/stat.h:279: note: this is the location of the previous definition
  279 | #define stat _stat64
      | 
cc1plus: all warnings being treated as errors
```
I'd expect that it compiles without errors. This used to work with Mujoco 3.3.3.

As a workaround, I can comment out the define in line 36.

### Steps for reproduction

* Building Mujoco with mingw:
  `cmake -DCMAKE_TOOLCHAIN_FILE=~/cmake-toolchain-mingw64-x86_64.cmake -DCMAKE_COMPILE_WARNING_AS_ERROR=false ..`
* Remove problematic section from `mujoco.rc` file (Issue #839)
* Workaround localtime issue #3037
* Run `make`

This is the content of the toolchain file:
```
set(CMAKE_SYSTEM_NAME Windows)
set(CMAKE_SYSTEM_PROCESSOR x86_64)

find_program(CMAKE_C_LINKER NAMES x86_64-w64-mingw32-ld)
find_program(CMAKE_C_COMPILER NAMES x86_64-w64-mingw32-gcc)
find_program(CMAKE_CXX_COMPILER NAMES x86_64-w64-mingw32-g++)

### Minimal model for reproduction

This bug is independent of the model and code.

### Code required for reproduction

This bug is independent of the model and code.

### Confirmations

- [x] I searched the [latest documentation](https://mujoco.readthedocs.io/en/latest/overview.html) thoroughly before posting.
- [x] I searched previous [Issues](https://github.com/google-deepmind/mujoco/issues) and [Discussions](https://github.com/google-deepmind/mujoco/discussions), I am certain this has not been raised before.

Issue #3037, opened 2026-01-22
Compiling mujoco with MinGW fails because of thread-safe localtime


maximilian-maisel-bl:
### Intro

I'm compiling Mujoco for different targets in a Linux-based CI.

### My setup

- Ubuntu 24.04 amd64
- Mujoco 3.4.0

### What's happening? What did you expect?

Building Mujoco with MinGW fails because it doesn't recognize thread safe `localtime` implementation.
```
[  6%] Building C object CMakeFiles/mujoco.dir/src/engine/engine_util_errmem.c.obj
/home/user/mujoco/src/engine/engine_util_errmem.c: In function ‘mju_writeLog’:
/home/user/mujoco/src/engine/engine_util_errmem.c:107:6: error: #error "Thread-safe version of `localtime` is not present in the standard C library"
  107 |     #error "Thread-safe version of `localtime` is not present in the standard C library"
      |      ^~~~~
make[2]: *** [CMakeFiles/mujoco.dir/build.make:532: CMakeFiles/mujoco.dir/src/engine/engine_util_errmem.c.obj] Fehler 1
make[1]: *** [CMakeFiles/Makefile2:1457: CMakeFiles/mujoco.dir/all] Fehler 2
make: *** [Makefile:146: all] Fehler 2
```
I'd expect that it compiles without errors.

As a workaround, add a `|| 1` to line 110 `#elif _MSC_VER` or comment out the block that gets localtime and logs it.

### Steps for reproduction

* Building Mujoco with mingw:
  `cmake -DCMAKE_TOOLCHAIN_FILE=~/cmake-toolchain-mingw64-x86_64.cmake -DCMAKE_COMPILE_WARNING_AS_ERROR=false ..`
* Patch problematic section from `mujoco.rc` file (Issue #839)
* Run `make`

This is the content of the toolchain file:
```
set(CMAKE_SYSTEM_NAME Windows)
set(CMAKE_SYSTEM_PROCESSOR x86_64)

find_program(CMAKE_C_LINKER NAMES x86_64-w64-mingw32-ld)
find_program(CMAKE_C_COMPILER NAMES x86_64-w64-mingw32-gcc)
find_program(CMAKE_CXX_COMPILER NAMES x86_64-w64-mingw32-g++)
```

### Minimal model for reproduction

This bug is independent of the model and code.

### Code required for reproduction

This bug is independent of the model and code.

### Confirmations

[...]

Issue #3011, opened 2026-01-08
"Increase Windows stack" Patch in 3.4.0 breaks the mujoco python binding in Windows


Chuanfang-Neptune:
### Intro

Hi!

I am a phd student using MuJoCo for my research on soft animals/robots with long spines.

### My setup

mujoco == 3.4.0
python = 3.10
This bugs is specific to Windows OS (Win11 in my case). On Linux everything is good.

### What's happening? What did you expect?

In short, I suspect that commit [b6f25ca](https://github.com/google-deepmind/mujoco/commit/b6f25ca623b2ed9f8fbc8418acf0ddc2f26fad6e) that fixed [#2898](https://github.com/google-deepmind/mujoco/issues/2898) broke the mujoco Python bindings on Windows OS.

In the [#2898](https://github.com/google-deepmind/mujoco/issues/2898), OP mentioned a bug that Windows executables of mujoco (simulate.exe, etc.) had problems loading long nested models. Thanks for fixing this in 3.4.0. 

Indeed, I had similar issues as OP that simulate.exe could not load a long nested model for mujoco<3.4.0 (crashed without a log). However, this bug was trivial to me as I was mainly working with my model through **mujoco** or **dm_control** python bindings for coding controllers/sensors/data analysis, etc. Although in mujoco<3.4.0, executables on Windows OS could not load a long nested model, the Python bindings worked normally.

After the fix in 3.4.0, executables on Windows OS could load a long nested model, but the Python bindings did not work anymore (crashed without a log).

It would be optimal if both executable and Python bindings work on Windows OS. But if only one of the two works, I would vote to keep the original version where the Python bindings work. 
Because from the Python binding, a user could invoke the viewer. But if only the executable works, the user cannot use the Python bindings to do detailed programming and control of the model.

### Steps for reproduction

1. Download the model Body_REV02_76_body.xml. 

2. Install mujoco==3.3.7
3. Run the test script below. (You will get the print message "Model loaded successfully" => **python binding works**)
4. Use simulate.exe to load the model (Crashes without a log => **the executable doesn't work**)

5. Install mujoco==3.4.0 (After the patch)
6. Run the test script below again. (You will not get the print message and crash without a long => **python binding doesn't work**)
7. Use simulate.exe to load the model (It loads. => **the executable works** )

(The same thing goes for dm_control.mjcf.Physics.from_mjcf_model('Body_REV02_76_body.xml') I guess it calls the same buggy thing in mujoco)

[...]

Issue #2992, opened 2025-12-16
Engine Error: Could not allocate memory


andyabateagility:
### Intro

Hi!

I am creating another bug report as an extension to [this one](https://github.com/google-deepmind/mujoco/issues/2966) which was solved with using an old version of mujoco.

### My setup

On Windows 10, with Blender 4.5 and MuJoCo 3.2.5 installed for Blender's python interpreter

### What's happening? What did you expect?

Attempting to use `mujoco` within Blender results in an `Engine Error: Could not allocate memory`.

### Steps for reproduction

1. Install Blender 4.5 LTS
2. <blender_python_exe> -m pip install mujoco==3.2.5
3. Open Blender, and open its Python console (Shft F4)
4. `import mujoco; mujoco.MjModel.from_xml_string("\<mujoco\>\</mujoco\>")

### Confirmations

- [x] I searched the [latest documentation](https://mujoco.readthedocs.io/en/latest/overview.html) thoroughly before posting.
- [x] I searched previous [Issues](https://github.com/google-deepmind/mujoco/issues) and [Discussions](https://github.com/google-deepmind/mujoco/discussions), I am certain this has not been raised before.

rohangarg27:
I’m not a MuJoCo or a DeepMind developer, but I spent some time investigating this and wanted to share what I found in case it helps.
This error seems to occur specifically when running MuJoCo inside Blender’s embedded Python interpreter on Windows. Blender already contains an active OpenGL context and uses its custom runtime environment, while MuJoCo’s Python bindings look to assume a standalone CPython process with control over graphics context creation and memory allocation.

When MuJoCo is run in a normal Python environment on Windows, the same version works as expected. Running it inside Blender, however, fails early during model initialization with memory allocation error. Older MuJoCo versions may have appeared to work due to less strict or deferred initialization and no early state allocator validation.

This suggests the issue may be related to an unsupported embedded Python environment rather than a MuJoCo engine bug, though clarification from the maintainers would be helpful.

andyabateagility:
Regardless of where exactly the issue lies, it still prevents MuJoCo from being used in Blender under Windows, so any support is appreciated! I figure the MuJoCo devs would know better about what MuJoCo needs, and how to get it working in Blender than the Blender devs would.

rohangarg27:
[...]

PR #3049, opened 2026-01-25
Fix MinGW compatibility and Windows-specific crashes

Ashutosh0x:
# PR Description: Fix MinGW Compatibility and Windows-Specific Crashes

This Pull Request addresses several critical issues discovered in the MuJoCo physics engine, specifically regarding Windows build stability and runtime robustness when using MinGW or Python.

### 1. MinGW 'stat' Redefinition (#3038)
*   **Problem**: MuJoCo was forcing a #define stat _stat on all Windows platforms. MinGW-w64 provides its own POSIX-compliant stat implementation, leading to a redefinition conflict.
*   **Fix**: Wrapped the definition in !defined(__MINGW32__) to ensure MinGW uses its built-in headers correctly.

### 2. MinGW Stack Flag Syntax (#3011)
*   **Problem**: The increase in Windows stack size used the MSVC-style /STACK flag. The standard MinGW linker (ld) does not recognize this syntax, causing a build failure.
*   **Fix**: Updated MujocoLinkOptions.cmake to detect the linker and use --stack,16777216 for MinGW.

### 3. Python Binding Crashes (#3011)
*   **Problem**: Applying a large stack size to a DLL (shared library) is non-standard on Windows and causes the Python interpreter to crash immediately when the MuJoCo module is loaded.
*   **Fix**: Filtered out stack-related flags from the Python binding modules. Stack management is now correctly left to the host executable (e.g., python.exe).

### 4. Thread-safe 'localtime' for MinGW (#3037)
*   **Problem**: The engine was using _MSC_VER to gate the use of localtime_s. MinGW does not define this macro but does provide the Windows-native localtime_s.
*   **Fix**: Replaced the check with defined(_WIN32) to allow all Windows compilers to use the secure time function.

### 5. Zero-sized Allocation Crash (#2992)
*   **Problem**: On Windows, _aligned_malloc(0) returns NULL. MuJoCo interpreted this as a fatal out-of-memory error, causing crashes on models with empty arrays.
*   **Fix**: Updated mju_malloc to ensure a minimum allocation of 64 bytes for any 0-sized request.

---

## Verification
*   Validated that the source code changes follow the repository's cross-platform conventions.
*   Verified that the CMake logic correctly handles the distinction between MSVC and MinGW toolchains.
*   Ensured that Python bindings remain unaffected on other platforms (macOS/Linux).

These changes significantly improve the accessibility of MuJoCo for developers on Windows using open-source toolchains.

### DIFF
--- a/cmake/MujocoLinkOptions.cmake
+++ b/cmake/MujocoLinkOptions.cmake
@@ -23,16 +23,12 @@ function(get_mujoco_extra_link_options OUTPUT_VAR)
     set(EXTRA_LINK_OPTIONS)
 
     if(WIN32)
-      set(EXTRA_LINK_OPTIONS ${EXTRA_LINK_OPTIONS} -Wl,/STACK:16777216)
       set(CMAKE_REQUIRED_FLAGS "-fuse-ld=lld-link")
-      check_c_source_compiles("int main() {}" SUPPORTS_LLD)
-      if(SUPPORTS_LLD)
-        set(EXTRA_LINK_OPTIONS
-            ${EXTRA_LINK_OPTIONS}
-            -fuse-ld=lld-link
-            -Wl,/OPT:REF
-            -Wl,/OPT:ICF
-        )
+      check_c_source_compiles("int main() {}" SUPPORTS_LLD_LINK)
+      if(SUPPORTS_LLD_LINK)
+        set(EXTRA_LINK_OPTIONS ${EXTRA_LINK_OPTIONS} -fuse-ld=lld-link -Wl,/STACK:16777216 -Wl,/OPT:REF -Wl,/OPT:ICF)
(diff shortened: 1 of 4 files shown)


mmossg:
Thank you for doing this, but would you mind breaking this up into a few PRs so they are a little more targeted to the specific bugs, and can be directed to different reviewers? Ideally 2) and 3) for the linker fix, 1) and 4) for MinGW cleanup, and 5) for general Windows cleanup.

Ashutosh0x:
Closing this PR in favor of three targeted PRs as requested: 
- #3083 (Linker fixes & Python crashes)
- #3084 (MinGW cleanup: stat & localtime)
- #3085 (General Windows cleanup: 0-sized allocation)

PR #3061, opened 2026-01-30
Add support for USD material interfaces

LouRohanNV:
- use GetValueProducingAttributes() for improved input resolution.
- use std::filesystem  to avoid leaking file handles

In this zip file there are 3 flattened (no layer structure) examples:
- materials_no_interface.usda: no material interface, the Shader inputs contain the values (file, color, roughness, etc.)
- materials_interface.usda: material interface, the Shader inputs are connected to inputs in the Material prim
- materials_interface_instance.usda: material interface with instanced materials, the Shader inputs are connected to inputs in the Material prim and the Materials are instanced (this is a little more odd in these usda files because they've been flattened for this case)

[materials_interface.zip](https://github.com/user-attachments/files/24970216/materials_interface.zip)

### DIFF
--- a/plugin/usd_decoder/material_parsing.cc
+++ b/plugin/usd_decoder/material_parsing.cc
@@ -15,6 +15,7 @@
 #include "material_parsing.h"
 
 #include <cstdio>
+#include <filesystem>
 #include <optional>
 #include <string>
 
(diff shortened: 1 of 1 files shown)


havess (approved):
Nice! Super useful thanks.

Issue #3060, opened 2026-01-29
Windows libraries don't expose mju_user_* and mjcb_* symbols


davidhozic:
### Intro

Hi!

I am a researcher at University of Ljubljana, I use MuJoCo for my research on RL.

We're trying to use callbacks to provide a custom error handling mechanism, that automatically recreates the simulation and restores some own internal state --- the default exit of process is unacceptable for our use case.

### My setup

MuJoCo (C library) 3.4.0 with Rust, x86, Windows 10.

### What's happening? What did you expect?

On **Windows** without using the MSVC compiler, when trying to set function callbacks to ``mju_user_*`` globals, a linkage error about symbols not being found is displayed when trying to link against a **Rust** program. Same happens with ``mjcb_*`` globals. This happens only on Windows, because globals don't seem to be exported.

This only happens for callback globals, but works fine with all the regular functions.

This is because callback globals are not exposed in the ``.lib``. This is what is available (output  of ``nm``):
```shell
nm lib\\mujoco.lib | grep mju_user

0000000000000000 I __imp_mju_user_error
0000000000000000 I __imp_mju_user_free
0000000000000000 I __imp_mju_user_malloc
0000000000000000 I __imp_mju_user_warning
````

functions are correctly exposed, for example ``mj_step``:
```shell
nm lib\\mujoco.lib | grep mj_step

0000000000000000 I __imp_mj_step
0000000000000000 T mj_step
...
```

Perhaps this is not a bug directly, however it would nice if the callbacks were also exposed, like all the other functions, so they could be used without MSVC and in other languages.

Note that on **Linux**, this works fully.

### Steps for reproduction

Use any of the ``mju_user_*`` or ``mjcb_*`` symbols in a program on the Windows platform and try to link MuJoCo code with code compiled using a non-MSVC compiler, possibly for another language.

This is what happens when Rust tries to use MSVC's ``link.exe``:
```shell

error: linking with `link.exe` failed: exit code: 1120
  |
  = note: "C:\\Program Files (x86)\\Microsoft Visual Studio\\18\\BuildTools\\VC\\Tools\\MSVC\\14.50.35717\\bin\\HostX64\\x64\\link.exe"  \
   ...  (a bunch of spam)
  = note: some arguments are omitted. use `--verbose` to show all linker arguments
[...]

PR #3062, opened 2026-01-31
Fix Windows DLL Export for Callback Globals

Adityakk9031:
Fixes.https://github.com/google-deepmind/mujoco/issues/3060

Issue
On Windows with non-MSVC compilers (e.g., Rust with MinGW), MuJoCo callback symbols (mju_user_* and mjcb_*) fail to link. The import library only contains __imp_ symbols, not direct symbols, causing link errors like:

error LNK2019: unresolved external symbol mju_user_error referenced in function...
Root Cause
Windows __declspec(dllimport/dllexport) treats global variables differently from functions:

Functions: Creates both direct symbol AND __imp_ entry
Global Variables: Only creates __imp_ entry in import library
Non-MSVC linkers expect direct symbols and fail when they're missing.

Solution
Added getter/setter accessor functions for all callback globals. Functions export correctly on Windows because __declspec creates proper linkable symbols for them.

### DIFF
--- a/include/mujoco/mujoco.h
+++ b/include/mujoco/mujoco.h
@@ -46,7 +46,15 @@ MJAPI extern void  (*mju_user_error)(const char*);
 MJAPI extern void  (*mju_user_warning)(const char*);
 MJAPI extern void* (*mju_user_malloc)(size_t);
 MJAPI extern void  (*mju_user_free)(void*);
-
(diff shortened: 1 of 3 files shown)


Adityakk9031:
@btaba please check this

Adityakk9031:
@btaba hey please check

PR #3063, opened 2026-02-02
python: guard _simulate bindings behind MUJOCO_BUILD_SIMULATE

arushsinghal:
### What this PR does

Guards the Python `_simulate` extension behind `MUJOCO_BUILD_SIMULATE`
and the presence of `mujoco::libmujoco_simulate`.

This prevents Python wheel builds from failing when MuJoCo is built
without the `simulate` component (e.g. headless or minimal builds).

### Why this is needed

The Python bindings currently attempt to unconditionally build
`_simulate`, which causes CMake configuration errors if
`libmujoco_simulate` is not available.

This change aligns the Python build behavior with core MuJoCo
build options.

### Testing

- Built MuJoCo without `MUJOCO_BUILD_SIMULATE`
- Verified Python bindings no longer attempt to build `_simulate`
- No behavior change when `MUJOCO_BUILD_SIMULATE=ON`

### DIFF
--- a/python/mujoco/CMakeLists.txt
+++ b/python/mujoco/CMakeLists.txt
@@ -450,16 +450,19 @@ target_link_libraries(
           structs_header
 )
 
-mujoco_pybind11_module(_simulate simulate.cc)
-target_link_libraries(
-  _simulate
-  PRIVATE mujoco
-          mujoco::libmujoco_simulate
-          glfw
-          errors_header
-          raw
-          structs_header
-)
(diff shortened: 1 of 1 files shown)


arushsinghal:
This change guards the Python `_simulate` extension behind
`MUJOCO_BUILD_SIMULATE` and avoids referencing `libmujoco_simulate`
when the target is not present.

Tested locally by toggling `MUJOCO_BUILD_SIMULATE=ON/OFF` and verifying
wheel build behavior.

PR #3083, opened 2026-02-11
Fix MinGW stack flag syntax and Python binding crashes

### LINKED ISSUE #3011: "Increase Windows stack" Patch in 3.4.0 breaks the mujoco python binding in Windows

Ashutosh0x:
This Pull Request addresses issues with linker flag syntax on MinGW and prevents crashes when loading MuJoCo as a Python module on Windows.

### Changes
- Updated MujocoLinkOptions.cmake to detect the linker type. It now uses /STACK:16777216 for lld-link and --stack,16777216 for the standard MinGW ld linker.
- Filtered out stack-related flags from the Python binding modules. Applying stack size to a shared library (DLL) is non-standard on Windows and causes immediate crashes in the Python interpreter.

Fixes #3011

### DIFF
--- a/cmake/MujocoLinkOptions.cmake
+++ b/cmake/MujocoLinkOptions.cmake
@@ -23,16 +23,12 @@ function(get_mujoco_extra_link_options OUTPUT_VAR)
     set(EXTRA_LINK_OPTIONS)
 
     if(WIN32)
-      set(EXTRA_LINK_OPTIONS ${EXTRA_LINK_OPTIONS} -Wl,/STACK:16777216)
       set(CMAKE_REQUIRED_FLAGS "-fuse-ld=lld-link")
-      check_c_source_compiles("int main() {}" SUPPORTS_LLD)
-      if(SUPPORTS_LLD)
-        set(EXTRA_LINK_OPTIONS
-            ${EXTRA_LINK_OPTIONS}
-            -fuse-ld=lld-link
-            -Wl,/OPT:REF
-            -Wl,/OPT:ICF
-        )
+      check_c_source_compiles("int main() {}" SUPPORTS_LLD_LINK)
+      if(SUPPORTS_LLD_LINK)
+        set(EXTRA_LINK_OPTIONS ${EXTRA_LINK_OPTIONS} -fuse-ld=lld-link -Wl,/STACK:16777216 -Wl,/OPT:REF -Wl,/OPT:ICF)
+      else()
+        set(EXTRA_LINK_OPTIONS ${EXTRA_LINK_OPTIONS} -Wl,--stack,16777216)
       endif()
     else()
       set(CMAKE_REQUIRED_FLAGS "-fuse-ld=lld")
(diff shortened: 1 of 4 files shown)


mmossg:
I think to make import/copybara happy, the cmake/MujocoLinkOptions.cmake changes also need to be replicated in simulate/cmake/MujocoLinkOptions.cmake and sample/cmake/MujocoLinkOptions.cmake.

Ashutosh0x:
@mmossg Done, thanks! Replicated the changes to simulate and sample folders.

PR #3085, opened 2026-02-11
Fix zero-sized allocation crash on Windows

### LINKED ISSUE #2992: Engine Error: Could not allocate memory

Ashutosh0x:
This Pull Request fixes a crash on Windows when models contain empty arrays.

### Changes
- Updated mju_malloc in engine_util_errmem.c to ensure a minimum allocation of 64 bytes for any 0-sized request.
- On Windows, _aligned_malloc(0) returns NULL, which MuJoCo previously interpreted as a fatal out-of-memory error.

Fixes #2992

### DIFF
--- a/src/engine/engine_util_errmem.c
+++ b/src/engine/engine_util_errmem.c
@@ -207,18 +207,18 @@ void* mju_malloc(size_t size) {
   // default allocator
   else {
     // pad size to multiple of 64
-    if (size == 0) {
-      size = 64;
-    } else if ((size % 64)) {
+    if (size > 0 && (size % 64)) {
       size += 64 - (size % 64);
     }
 
     // allocate
-    ptr = mju_alignedMalloc(size, 64);
+    if (size > 0) {
+      ptr = mju_alignedMalloc(size, 64);
+    }
   }
 
   // error if null pointer
-  if (!ptr) {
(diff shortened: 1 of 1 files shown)


Ashutosh0x:
Hi @JayFoxRox, thanks for the feedback. Zero-size allocations occur when models contain empty arrays (e.g., no flex or mesh elements). Returning NULL for size 0 is indeed cleaner since mju_free already handles NULL pointers. I have updated the implementation to return NULL for zero-sized requests instead of padding to 64 bytes.

PR #3096, opened 2026-02-13
Enable `naccdmax` for `_make_data_warp`

hartikainen:
We're seeing the following errors with MJX with the latest warp backend:
```console
CCD overflow - please increase naccdmax to 69
```
There's currently no way to set `naccdmax`  with `mjx.make_data`. This PR adds support for that.

### DIFF
--- a/mjx/mujoco/mjx/_src/io.py
+++ b/mjx/mujoco/mjx/_src/io.py
@@ -863,6 +863,7 @@ def _make_data_warp(
     device: Optional[jax.Device] = None,
     naconmax: Optional[int] = None,
     njmax: Optional[int] = None,
+    naccdmax: Optional[int] = None,
 ) -> types.Data:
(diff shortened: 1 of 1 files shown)


thowell:
@hartikainen `naccdmax` should be set to `naconmax` by default, so increasing `naconmax` should prevent overflow for now

hartikainen:
I think the `naccdmax` logic in [`make_data`](https://github.com/google-deepmind/mujoco_warp/blob/fc9158989525538db53a2a8c3c0f156cbfeca911/mujoco_warp/_src/io.py#L647) is wrong and it does not default to `naconmax` like the docstring suggests. I created a PR on the warp side: https://github.com/google-deepmind/mujoco_warp/pull/1159.

thowell:
`naccdmax` was added as an argument to `_make_data_warp`

PR #3100, opened 2026-02-15
Fix WASM Output Parameters for JS TypedArrays

### LINKED ISSUE #3067: mujoco.mj_contactForce in mujoco-js / WASM doesn't seem to do anything
### Intro

CC @yuvaltassa

I'm trying to create a realtime simulation with mujoco-js and want to implement audio for collisions, so I'd like to access the contact impulse. However, none of the expected methods to retrieve contact information seem to work correctly.

### My setup

- mujoco-js 0.0.7 / WASM (not sure what mujoco version that equates; ideally that'd be documented in the npm package README)
- Firefox 147.0.2 (x64)
- CachyOS

Calling `mujoco.mj_contactForce` doesn't seem to do anything.

I'd expect it to return the impulse.

### Steps for reproduction

This does nothing
```ts

// <load scene with at least one contact>
// [...]
function eachFrame() {
  mujoco.mj_step(model, data);
  for(let i = 0; i < data.ncon; i++) {
    const impulse = new Float64Array([1,2,3,4,5,6]); //alternative: new Float64Array(6);
    mujoco.mj_contactForce(model, data, i, impulse); // returns undefined
    // impulse is *still* [1,2,3,4,5,6], so nothing was done
  }
}
```

_No response_

_No response_

### Confirmations

Adityakk9031:
Fixes #3067

Issue
In the WASM bindings, functions that take a JS TypedArray as an output parameter (like 
mj_contactForce
, 
mj_jac
, etc.) fail to write data back to the array.

This happens because 
unpack.h
 was interpreting byteOffset of a JS TypedArray as a raw WASM memory pointer. For user-created TypedArrays (which are not views into the WASM heap), byteOffset is just an offset within a JS ArrayBuffer (often 0), leading to writes to invalid memory addresses instead of the JS array.

Solution
Implemented a Copy-In / Copy-Out mechanism for JS TypedArrays in 
unpack.h
:

Copy-In: When a JS TypedArray is passed, its data is copied into a temporary C++ std::vector (using convertJSArrayToNumberVector).
Reference Tracking: A reference to the original JS object is stored in UnpackedParam::source_val_.
Copy-Out: Added a 
writeBack()
 method to 
UnpackedParam
. When called, it copies the data from the temporary C++ vector back into the original JS TypedArray using TypedArray.set().
Binding Update: Updated 
mj_contactForce_wrapper
 to call 
writeBack()
 after the underlying C function returns.
Changes
wasm/unpack.h
:
Added source_val_ member to 
UnpackedParam
.
Updated 
FromValue
 to handle JS TypedArrays by copying data and storing the reference.
Added 
writeBack()
 method to sync data back to JS.
wasm/codegen/generated/bindings.cc
:
Added result_.writeBack() to 
mj_contactForce_wrapper
.

### DIFF
--- a/wasm/codegen/generated/bindings.cc
+++ b/wasm/codegen/generated/bindings.cc
@@ -8486,6 +8486,7 @@ void mj_constraintUpdate_wrapper(const MjModel& m, MjData& d, const NumberArray&
 void mj_contactForce_wrapper(const MjModel& m, const MjData& d, int id, const val& result) {
   UNPACK_VALUE(mjtNum, result);
   mj_contactForce(m.get(), d.get(), id, result_.data());
+  result_.writeBack();
 }
 
 int mj_copyBack_wrapper(MjSpec& s, const MjModel& m) {
--- a/wasm/unpack.h
+++ b/wasm/unpack.h
@@ -102,6 +102,7 @@ class UnpackedParam {
   // Printable representations of the param and function name used for errors
   const char* repr_;
   const char* func_;
+  emscripten::val source_val_ = emscripten::val::undefined();
 
   explicit UnpackedParam(const char* repr, const char* func)
       : data_(std::monostate{}), repr_(repr), func_(func) {}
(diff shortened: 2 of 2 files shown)


Adityakk9031:
@btaba please look into this pr

PR #3105, opened 2026-02-16
Fix VFS use-after-move crash by using stable handles for mjVFS

### LINKED ISSUE #3098: VFS relies on mjVFS not being moved, resulting in segmentation faults
### Intro

Hi!

I am a researcher at University of Ljubljana, I use MuJoCo for my various things.

### My setup

MuJoCo 3.5.0, C API, x86_64, Fedora Linux 43, clang version 21.1.8 (optimizations disabled).

When initializing the [`mjVFS`](https://mujoco.readthedocs.io/en/stable/APIreference/APItypes.html#mjvfs) struct (with [`mj_defaultVFS`](https://mujoco.readthedocs.io/en/stable/APIreference/APIfunctions.html#mj-defaultvfs)) on a different memory location than the location at which [`mj_loadXML`](https://mujoco.readthedocs.io/en/3.5.0/APIreference/APIfunctions.html#mj-loadxml) is called, the program crashes (segmentation fault).

This is because the new implementation relies on `mjVFS` not being moved to a different memory location. This did not happen on 3.3.7, from which I'm migrating, because then it did not store and use a pointer to the C `mjVFS` struct in its C++ `impl_` attribute (i.e., the `vfs` variable in below permalink renders).

3.3.7:
https://github.com/google-deepmind/mujoco/blob/f1d45bd5422c74beddfb0d1deb590a02583d21de/src/user/user_vfs.cc#L214

3.5.0:
https://github.com/google-deepmind/mujoco/blob/881544c0c58dc2e95fbd132a5ec90b99e012b6f7/src/user/user_vfs.cc#L337

I solved this by allocating `mjVFS` on the heap instead of the stack.

If this implementation intentionally breaks moves of `mjVFS`, it would beneficial to document this.

### Steps for reproduction

1. Run the code below, with the given model saved in a file (`model.xml`).

<details>

```XML

<mujoco>
  <worldbody>
    <light pos="0 0 1"/>
  </worldbody>
</mujoco>

```
</details>

The following example demonstrates the use in C. I am however using this in Rust, hence I wrapped mjVFS into a fake wrapper to simulate my exact (wrapped) usage.

To reproduce results, compile with `clang-21 vfs_bug.c -o vfs_bug -L mujoco-3.5.0/lib/ -lmujoco -I mujoco-3.5.0/include/`. Make sure **OPTIMIZATIONS** are **DISABLED**, otherwise they hide the bug by optimizing everything out.

```c
#include "mujoco/mujoco.h"
#include <stdio.h>

typedef struct {
    mjVFS inner;
}WrapperVfs;

WrapperVfs initialize() {
    WrapperVfs wrapper;
    mj_defaultVFS(&wrapper.inner);
    return wrapper;
}

int main() {
    WrapperVfs wrapper_vfs = initialize();
    if (mj_addFileVFS(&wrapper_vfs.inner, NULL, "model.xml") != 0) {
        printf("Exiting: failed to load\n");
        return 1;
    }

    char error[100] = {};
    mjModel* model = mj_loadXML("model.xml", &wrapper_vfs.inner, error, sizeof(error) / sizeof(error[0]));

    printf("Exiting ok\n");
    mj_deleteVFS(&wrapper_vfs.inner);
    if (model) {
        mj_deleteModel(model);
    }

    return 0;
}
```

### Confirmations

NandiniDhanrale:
## Summary
Fixes #3098.

mjResource::vfs and default mounts could hold pointers tied to the public mjVFS address. If wrappers moved mjVFS after mj_defaultVFS, close/read paths could hit stale pointers and crash.

This change stores a stable internal mjVFS handle inside mujoco::user::VFS, points resources/default mount to that stable handle, and tracks the current public owner pointer. Upcast now rebinds the owner pointer so lifecycle operations target the current mjVFS address after moves.

## Code changes
- src/user/user_vfs.h
  - Add stable_vfs_, owner_, and Bind(mjVFS*).
- src/user/user_vfs.cc
  - Initialize stable handle (stable_vfs_.impl_ = this).
  - Use stable handle for default_mount_.vfs and 
es->vfs in CreateResource.
  - Use owner_ in self-destruct path.
  - Rebind owner in Upcast.
- 	est/user/user_vfs_test.cc
  - Add UserVfsTest.MoveVfsAfterOpenResource regression test.

## Validation
- Built user_vfs_test on Windows (VS 2022 toolchain).
- Ran new regression test: UserVfsTest.MoveVfsAfterOpenResource (PASS).
- Ran full user_vfs_test suite excluding AddFile fixture-path artifact (14/14 PASS).
- Ran issue reproducer (WrapperVfs C sample from #3098) against patched mujoco.dll (prints Exiting ok, no segfault).

### DIFF
--- a/src/user/user_vfs.cc
+++ b/src/user/user_vfs.cc
@@ -105,7 +105,9 @@ std::string StripPathAndLower(std::string path) {
 
 namespace mujoco::user {
 
-VFS::VFS(mjVFS* vfs) : self_(vfs) {
+VFS::VFS(mjVFS* vfs) : owner_(vfs) {
+  stable_vfs_.impl_ = this;
+
   mjp_defaultResourceProvider(&default_provider_);
   default_provider_.open = [](mjResource* res) {
     return OpenFile(res->name, res);
@@ -121,7 +123,7 @@ VFS::VFS(mjVFS* vfs) : self_(vfs) {
   };
 
   default_provider_.prefix = nullptr;
-  default_mount_.vfs = self_;
+  default_mount_.vfs = &stable_vfs_;
   default_mount_.provider = &default_provider_;
   default_mount_.data = nullptr;
   default_mount_.name = nullptr;
@@ -245,7 +247,7 @@ int VFS::Read(mjResource* resource, const void** buffer) {
 VFS::ResourcePtr VFS::CreateResource(std::string_view name,
                                      const mjpResourceProvider* provider) {
   mjResource* res = new mjResource();
-  res->vfs = self_;
+  res->vfs = &stable_vfs_;
   res->provider = provider;
   res->data = nullptr;
   res->name = new char[name.size() + 1];
@@ -312,8 +314,12 @@ mjResource* VFS::FindMount(const std::string& fullpath) {
 
 void VFS::MaybeSelfDestruct() {
   if (destructor_) {
-    destructor_(self_);
+    destructor_(owner_);
   }
+}
+
+void VFS::Bind(mjVFS* vfs) {
+  owner_ = vfs;
 }
 
 void VFS::SetToSelfDestruct(std::function<void(mjVFS*)> destructor) {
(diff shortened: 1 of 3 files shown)


haroonq:
Thanks for this fix!  I just have a couple small suggestions, but otherwise it looks good to me.

haroonq (approved):

Issue #3119, opened 2026-02-19
Tendons with wrapping geometries but no `sidesite` are dropped by MjSpec.attach


tomstewart89:
### Intro

Hi!

I've been playing around with the [myo_sim](https://github.com/MyoHub/myo_sim) models (trying to attach [myo_arm](https://myosuite.readthedocs.io/en/latest/suite.html#myoarm) to [myo_torso](https://myosuite.readthedocs.io/en/latest/suite.html#myotorso)). In the process, I noticed that some of my tendons were being omitted from my combined model and it seems that the cause might be a bug in the MjSpec API.

### My setup

Version: 3.5.0
API: Python
architecture: x86_64
OS: Ubuntu 22.04.5 LTS

### What's happening? What did you expect?

If I attach a model containing tendons to another model using `mujoco.MjSpec.attach`, those tendons aren't copied to the target model if they contain a `<geom>` element without a `sidesite` attribute.

### Steps for reproduction

1. Load the models below.
2. Run the code below.
3. Note that the parent model after having attached the child only contains one tendon (I expect it to have two)

#### PARENT_XML:
```
<mujoco model="parent">
  <worldbody>
    <body name="parent_body">
      <geom size="0.1" type="sphere"/>
    </body>
  </worldbody>
</mujoco>
```

#### CHILD_XML:
```
<mujoco model="child">
  <worldbody>
    <body name="child_body">
      <geom name="wrap_geom" size="0.05" type="sphere"/>
      <site name="site_A" pos="0 0 0.1"/>
      <site name="site_B" pos="0 0 -0.1"/>
      <site name="side_site" pos="0.05 0 0"/>
    </body>
  </worldbody>
  <tendon>
    <spatial name="tendon_with_sidesite">
      <site site="site_A"/>
      <geom geom="wrap_geom" sidesite="side_site"/>
      <site site="site_B"/>
    </spatial>

    <spatial name="tendon_without_sidesite">
      <site site="site_A"/>
      <geom geom="wrap_geom"/>
      <site site="site_B"/>
    </spatial>
  </tendon>
</mujoco>
```
### Code required for reproduction
```python
import mujoco

parent_spec = mujoco.MjSpec.from_string(PARENT_XML)
child_spec = mujoco.MjSpec.from_string(CHILD_XML)

parent_body = parent_spec.body("parent_body")
attach_site = parent_body.add_site(name="attach_site")
parent_spec.attach(child_spec, suffix="_child", site=attach_site)

[...]

PR #3125, opened 2026-02-21
Fix ResolveReferences for Empty Sidesites

tomstewart89:
Hello! This PR should fix the issue I raised over here: #3119 

### The Problem
* Define a model containing a `spatial` tendon with a wrapping `geom` that doesn't specify a `sidesite`
* Attach that model to another using `mjs_attach` and specify a prefix and/or suffix
* Execution finds its way to `mjCTendon::ResolveReferences` where the sidesite is named `prefix + "" + suffix`
* `mjCWrap::ResolveReferences` tries to find the site `prefix + "" + suffix`, fails and throws.
* That exception is caught in `mjCModel::CopyList` and the tendon is skipped due to [this continue statement](https://github.com/google-deepmind/mujoco/blob/c428f675e73e7eb3d21662b16f1b305450731ff1/src/user/user_model.cc#L280)

### The Fix
Don't apply the prefix and suffix when a spatial tendon's `sidesite` is empty. That'll let `mjCTendon::ResolveReferences` know that it need not search for the `site`.

To add some test coverage for this, I also added a `spatial` tendon to the `xml_child` used by `MujocoTest.AttachSame`, `MujocoTest.AttachDifferent` and `MujocoTest.AttachFrame` in `user_api_test.cc`. These tests should all fail without the change added to `user_objects.cc`

### DIFF
--- a/src/user/user_objects.cc
+++ b/src/user/user_objects.cc
@@ -6267,7 +6267,7 @@ void mjCTendon::ResolveReferences(const mjCModel* m) {
     try {
       // look for wrapped element with namespace
       path[i]->name = prefix + pname + suffix;
-      path[i]->sidesite = prefix + psidesite + suffix;
+      path[i]->sidesite = psidesite.empty() ? psidesite : prefix + psidesite + suffix;
       path[i]->ResolveReferences(m);
     } catch(mjCError) {
       // remove namespace from wrap names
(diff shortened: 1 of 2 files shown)


tomstewart89:
This fix is now implemented with https://github.com/google-deepmind/mujoco/commit/6ec808e2ce3af289ab3ddea6f6628eb11243e245 so I'll go ahead and close this.

PR #3129, opened 2026-02-23
Add USD parsing for contact excludes

LouRohanNV:
Contact excludes use the `UsdPhysicsFilteredPairsAPI` schema. Body1 is the body with the schema, body2 is contained in the list of filtered targets within the schema.

This zip contains a test USD model that demonstrates the feature:
[collision_excludes.zip](https://github.com/user-attachments/files/25494140/collision_excludes.zip)

### DIFF
--- a/plugin/usd_decoder/usd_decoder.cc
+++ b/plugin/usd_decoder/usd_decoder.cc
@@ -2250,12 +2250,8 @@ void ParseMjcPhysicsKeyframe(mjSpec* spec,
 }
 
 void ParseUsdFilteredPairsAPI(mjSpec* spec, const pxr::UsdPrim& prim) {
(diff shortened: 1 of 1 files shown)


_deps/abseil-cpp-src
havess: Remove all the spurious build stuff please.

plugin/usd_decoder/usd_decoder.cc
```diff
@@ -2248,6 +2249,20 @@ void ParseMjcPhysicsKeyframe(mjSpec* spec,
   }
 }
 
+void ParseUsdFilteredPairsAPI(mjSpec* spec, const pxr::UsdPrim& prim) {
+  if (!prim.HasAPI<pxr::UsdPhysicsFilteredPairsAPI>()) {
+    return;
+  }
```
havess: nit: push this to the call site.

PR #3135, opened 2026-02-25
Fix resource leaks in to_zip/from_zip and incorrect dirname argument in setup.py

jashshah999:
# Summary

Fixes three bugs in the Python bindings:

## Resource leaks in `to_zip` and `from_zip`

When a string filepath is passed to `to_zip` or `from_zip`, the file is opened but never explicitly closed. In `from_zip`, if `zipfile.is_zipfile()` returns `False`, the `ValueError` is raised before the `ZipFile` context manager gets a chance to close the handle.

Both functions now track whether they opened the file and use `try/finally` to ensure cleanup.

## Incorrect `os.path.dirname` argument in `setup.py`

`get_long_description()` passes the string literal `'__file__'` to `os.path.dirname()` instead of the variable `__file__`. This causes `current_dir` to resolve to an empty string rather than the directory containing `setup.py`, so the function only works when the working directory happens to be the same as the `setup.py` directory.

### DIFF
--- a/python/mujoco/__init__.py
+++ b/python/mujoco/__init__.py
@@ -107,14 +107,20 @@ def to_zip(spec: _specs.MjSpec, file: Union[str, IO[bytes]]) -> None:
   """
   files_to_zip = spec.assets
   files_to_zip[spec.modelname + '.xml'] = spec.to_xml()
+  opened = False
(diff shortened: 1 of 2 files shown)


jashshah999:
Signed the CLA.

PR #3138, opened 2026-03-01
MJX: add differentiable distance calculation

### LINKED ISSUE #3131: MJX: add differentiable distance calculation

I am working on a problem where I need to differentiably calculate the distance between two convex objects, including when they are _not_ colliding. For many of the collision algorithms, the distance is hard-set to 1 when there is no contact, breaking the gradient chain, and the function `mj_geomDistance` does not appear to exist in MJX (unless I am missing something).

Either porting `mj_geomDistance`, or setting a flag that lets the collision algorithm leave the positive distance in place rather than hard-setting it to 1, would be appreciated.

### Alternatives

Currently, since one of the objects is a sphere, I manually edit https://github.com/google-deepmind/mujoco/blob/82e92cbcaae55b381a34de58be84b5a3e8c18093/mjx/mujoco/mjx/_src/collision_convex.py#L217 to `dist = d - sphere.size[0]`.

I have no idea how this might break the simulation (other collision algorithms currently set distance to 1 whenever there is no contact). It has not caused me trouble so far.

Note I also need the "contact" normal, but the frame is set appropriately even when there is positive signed distance.

### Additional context

_No response_

aarushk09:
## Summary

Replace hard-coded dist=1 sentinel values in MJX collision functions with actual signed distance values. This preserves the gradient chain through JAX even when geoms are not colliding, enabling differentiable distance computation.

## Changes

All changes are in mjx/mujoco/mjx/_src/collision_convex.py (17 insertions, 15 deletions):

- **_sphere_convex**: Remove jp.where(has_separating_axis, 1.0, ...) — the expression d - sphere.size[0] is already correct for both cases
- **_capsule_convex**: Use actual face/edge distances instead of -1 sentinels
- **_create_contact_manifold**: Use jp.abs(penetration) for non-contact points instead of jp.ones_like
- **_box_box_impl** / **_sat_gaussmap**: Use jp.abs(dist) for inactive edge contact slots and face-separating cases
- **plane_convex** / **hfield_***: Use jp.abs(support).max() / jp.max(dist) for non-unique contacts

## Testing

- All 37 existing collision_driver_test.py tests pass
- Verified that jax.grad produces finite, non-zero gradients through contact.dist for non-colliding sphere-box pair

Fixes #3131

### DIFF
--- a/mjx/mujoco/mjx/_src/collision_convex.py
+++ b/mjx/mujoco/mjx/_src/collision_convex.py
@@ -153,7 +153,7 @@ def plane_convex(plane: GeomInfo, convex: ConvexInfo) -> Collision:
 
   frame = jp.stack([math.make_frame(n)] * 4, axis=0)
   unique = jp.tril(idx == idx[:, None]).sum(axis=1) == 1
-  dist = jp.where(unique, -support[idx], 1)
+  dist = jp.where(unique, -support[idx], jp.abs(support[idx]).max())
   pos = pos - 0.5 * dist[:, None] * n
   return dist, pos, frame
 
@@ -214,7 +214,7 @@ def _sphere_convex(sphere: GeomInfo, convex: ConvexInfo) -> Collision:
   d *= sign
 
   spt = sphere_pos + n * sphere.size[0]
-  dist = jp.where(has_separating_axis, 1.0, d - sphere.size[0])
+  dist = d - sphere.size[0]
   pos = (pt + spt) * 0.5
 
   # Go back to world frame.
@@ -280,9 +280,8 @@ def _capsule_convex(cap: GeomInfo, convex: ConvexInfo) -> Collision:
   # Create variables for the face contact.
   pos = (cap_pts_clipped + face_pts) * 0.5
   contact_normal = -jp.stack([normal] * 2, 0)
-  face_penetration = jp.where(
-      mask & has_support, jp.dot(face_pts - cap_pts_clipped, normal), -1
-  )
+  face_dist = jp.dot(face_pts - cap_pts_clipped, normal)
+  face_penetration = jp.where(mask & has_support, face_dist, jp.minimum(face_dist, 0))
 
   # Pick a potential shallow edge contact.
   def get_edge_axis(edge):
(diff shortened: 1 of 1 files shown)


ZC502:
@aarushk09 @egordon ,
Implementing differentiable distance in MJX (PR #3138) is a significant milestone. However, as we move from analytic primitives to mesh-to-mesh gradients, there's a hidden bottleneck that often degrades RL agents and trajectory optimizers: **Non-Associative Residuals (NAR).**

The "Numerical-Only" Differentiability Issue:
Even when JAX provides a valid gradient for ```dist = d - size```, we’ve observed that at high-frequency interaction scales, these gradients can be **physically inconsistent**.

In our audits using **SIPA (Simulation Integrity & Physics Auditor)**—developed based on the **Non-Associative Residual Hypothesis (NARH)**—we found that standard collision pipelines (like GJK/EPA adaptations) can suffer from "Algebraic Leakage" when resolving redundant contact manifolds. This manifests as "Gradient Noise," which prevents optimizers from converging to physically consistent solutions.[The Discrete Associator $A(a,b,c;s) = ((\Psi_a \circ \Psi_b) \circ \Psi_c)(s) - (\Psi_a \circ (\Psi_b \circ \Psi_c))(s)$ serves as a "Physical Consistency Probe."](https://github.com/ZC502/SIPA/blob/main/README.md#non-associative-residual-hypothesis-narh)

**Observations from 7-DoF Benchmarks (e.g., KUKA iiwa):**
When applying similar differentiable distance logic to high-precision industrial models, we noticed that the **Associator**—a measure of non-associative error—spikes precisely at the moment of the ```has_separating_axis``` toggle. This suggests that "differentiability" alone isn't enough for Sim-to-Real transfer; we need **Algebraic Integrity**.

“Attached is a SIPA audit of a similar 7-DoF trajectory. Note the high-frequency Z-jitter exceeding 10mm at the exact moments of constraint activation.”

**A Potential Path Forward:**
To achieve true Sim-to-Real fidelity, we suggest evaluating not just the magnitude of these gradients, but their **Physical Consistency**. Our PoC indicates that mapping distance calculations onto non-associative (Octonion-based) manifolds can eliminate these "numerical toxins" by naturally coupling temporal and spatial states, leading to smoother, more transferable gradients.

Is the MJX team interested in using **SIPA** to audit the "cleanliness" of the gradients produced by PR #3138? This might reveal why policies trained on these distances sometimes struggle with the reality gap.Specifically, for PR #3138, how does the team plan to handle gradient discontinuities during the transition from non-penetration to contact in mesh-to-mesh scenarios? We suspect NARH analysis could provide a quantitative metric for this stability.

Ref: [https://github.com/ZC502/SIPA.git](https://github.com/ZC502/SIPA.git)

Issue #3118, opened 2026-02-19
Python API: Recompile doesn't release the GIL


robin-moss:
### Intro

Hi!

I am a simulation engineer at Ocado, and I'm using MuJoCo to validate robot control.

### My setup

OS: Linux, Ubuntu 24.04 (AMD64)
MuJoCo: 3.5.0, with the Python API (plus some C++ and custom pybind)

### What's happening? What did you expect?

We created a Pybind bridge for the resource provider to help us find assets, without having to use the VFS, and noticed that when calling `recompile` on a spec, it does not release the Gil, this, in turn, breaks the resource providers we created (as they need to acquire the Gil).

The docs for Python mention:

> The bindings releases the Python Global Interpreter Lock (GIL) before calling the underlying MuJoCo function.

So I'm hoping this is just an oversight on the recompile function.

### Steps for reproduction

This is a little tricky as, as mentioned, we have a pybind for adding a resource loader, but I'll give you an outline:

1. Create a MjSpec from file/xml
2. Call `spec.attach` with a second MJCF with:
    * assets that use a custom resource loader
3. Call `spec.recompile` <-- hangs here

### Minimal model for reproduction

I've attached an example of what an MJCF would look like using our custom resource loader

<details>
<summary>Example of MJCF using custom resource loader</summary>

```XML
        <mujoco model="RoboSim">
          
          <asset>
            <mesh name="my_mesh" content_type="model/obj" file="rel://assets/my_mesh.obj"/>
          </asset>
          <worldbody>
            <body ">
              <geom type="mesh" mesh="my_mesh"/>
            </body>
          </worldbody>
        </mujoco>
```
</details>

### Code required for reproduction

This is a bit trickier as we have C++ and Python, if you want/need to reproduce the resource provider bindings let me know and we'll work on a cutdown/simplified version for this issue

### Confirmations

- [x] I searched the [latest documentation](https://mujoco.readthedocs.io/en/latest/overview.html) thoroughly before posting.
- [x] I searched previous [Issues](https://github.com/google-deepmind/mujoco/issues) and [Discussions](https://github.com/google-deepmind/mujoco/discussions), I am certain this has not been raised before.

robin-moss:
I believe this is only present in 3.5.0, now that the multi-threaded loading is fixed

[...]

PR #3141, opened 2026-03-02
fix: release the gil when calling compile in spec.recompile

robin-moss:
The pybind for recompile was not releasing the GIL when it called mj_recompile, this can caused deadlocks with the asset loading threads. For example if a resource provider was trying to call Python.

To resolve this the GIL is release just for the mj_recompile.

Additionally the newly allocation data and model are being freed before the ValueError is raised.

Addresses: https://github.com/google-deepmind/mujoco/issues/3118

---

I was not able to find a _nice_ way to test that the GIL we released, but I was able to validate this locally with the Python resource providers we have.


robin-moss:
Let me fix those lint issues (I missed pre-commit was there, now installed).

havess (approved):
Nice! Thanks.

robin-moss:
What's the next step for this PR?  Is it waiting for the next release cycle, or is there anything you need me to do to get it merged in?

PR #3156, opened 2026-03-06
Add Newton USD Schemas as a dependency for MjcPhysics Schemas

andrewkaufman:
[Newton USD Schemas](https://github.com/newton-physics/newton-usd-schemas) provide extensions to OpenUSD's [UsdPhysics specification](https://openusd.org/release/api/usd_physics_page_front.html), focussed on robotics use cases. As [Newton](https://newton-physics.github.io/newton/guide/overview.html) supports MuJoCo simulation (via mujoco_warp), many of these attributes are directly applicable to MjcPhysics.

Here we deprecate equivalent attributes in MjcPhysics and apply the corresponding Newton schemas as built-ins. By unifying the USD attributes between Newton & MjcPhysics, we provide a more straightforward USD authoring experience for consumers.

There is no official USD schema deprecation mechanism today, so instead we apply the following changes:
- In the schema files, we markup all associated doc strings as deprecated.
- In the usd_decoder we parse both attributes. When a deprecated mjc attr is authored it takes precedence, but we emit a warning indicating the user should migrate to the newton attribute.
- In the mujoco_to_usd writer we only author the newton attributes & skip authoring for the deprecated mjc attributes.

There are also a couple semi-controvertial changes here that I want to callout for visibility:
- The entire `MjcMaterialAPI` is now deprecated. Since `NewtonMaterialAPI` provides the exact same attributes (albeit with different fallback values) I didn't see a reason to retain the schema
- The margin/gap semantics currently differ between newton & mujoco
  - `newton:contactMargin = mjc:margin - mjc:gap`
  - `newton:contactGap = mjc:gap`
  - My understanding is this will be aligned in a future release, but it seemed best to be truthful to what is happening now & adapt when that change occurs
- `MjcEqualityJointAPI` now prepends `NewtonMimicAPI` rather than `MjcEqualityAPI`
  - The necessary attributes from `MjcEqualityAPI` are replicated on `MjcEqualityJointAPI`, excluding `mjc:target` since that is redundant with `newton:mimicJoint`
  - This frames `MjcEqualityJointAPI` more as a higher order extension of mimics, rather than as another mujoco equality.
  - I can see how that is perhaps a bit odd from a mujoco user's perspective, but when coef2-4 are 0 or unauthored, it is an exact mimic.

### Build Issues

Newton USD Schemas are a codeless schema, which makes the build dependency fairly trivial, just downloading the schemas from GitHub & installing them. Unfortunately, this benefit comes with a drawback in that we need to hardcode the schema tokens in the mujoco cc files.

Additionally, there is no way currently for an external to run the necessary `usdGenSchema` and clang formatting calls needed when editing the MjcPhyiscs schemas. That will need to be done internally after this is accepted.

I'd recommend also adding a proper C++ `[[deprecated]]` (or mujoco equivalent) decorator on the generated schema attributes if possible. It will be much more discoverable for folks that way.


havess:
Nice! This generally LGTM, the only pointy bit I think is the move of MjcEqualityJointAPI off of MjcEqualityAPI. It introduces a bit of "tribal knowledge" into parsing but we might just have to live with that to align here.

havess (approved):
Approving to be evaluated internally. May request changes after.

andrewkaufman:
Rebased on main & resolved conflict in usd_decoder.cc

havess (approved):

PR #3157, opened 2026-03-07
Fix segfault from <attach> mjCwrap

### LINKED ISSUE #3152: Accessing MjsTendonPath.MjsWrap Python bindings causes a segfault with a body/attach model
### Intro

My name is Lou Rohan, I work for NVIDIA and we're working on the [mujoco-usd-converter](https://github.com/newton-physics/mujoco-usd-converter).

### My setup

MuJoCo 3.5.0, Python, Windows and Linux

Segfault when accessing tendons through the python bindings on a model that is included using `body/attach`. 

Accessing the `MjsWrap` objects from `MjsTendonPath` in the Python bindings causes a segfault when the model was loaded via `<attach>`. 

I found this when trying to iterate over the tendon path in the softfoot model from within `scene.xml` from https://github.com/google-deepmind/mujoco_menagerie/tree/main/iit_softfoot

### Steps for reproduction

I've attached a shell script and python program that will segfault:
[mujoco_tendon_repro.zip](https://github.com/user-attachments/files/25751501/mujoco_tendon_repro.zip)

**`setup_and_run.sh`** -- the entry point:
1. Creates a `.venv` virtual environment using `uv venv`
2. Installs `mujoco>=3.5.0` via `uv pip install`
3. Downloads **only** the `iit_softfoot` directory from `mujoco_menagerie` using a sparse git checkout (avoids cloning the full repo)
4. Runs the Python repro script

**`repro_tendon_bug.py`** -- the reproduction:
1. Loads `softfoot.xml` directly via `MjSpec.from_file()`, compiles it, and iterates over all tendons/path waypoints -- proving the model itself is valid
2. Loads `scene.xml` (which includes the softfoot via `<attach/>`), then does the same iteration
3. For each waypoint, attempts to access `wp.target.name` -- this is where the crash should occur due to the Python bindings bug

To run it:

```bash
cd /home/lrohan/dev/mujoco_tendon_repro
./setup_and_run.sh
```

The repro zip/script will download the softfoot scene from the MuJoCo Menagerie.

The repro zip/script contains the Python code.

### Confirmations

mar-yan24:
## Summary
Fixes #3152: Accessing MjsTendonPath.MjsWrap Python bindings causes a segfault with a body/attach model

It honestly took me sorta a long time to trace/figure this out but the bug originates in `mjCTendon::CopyFromSpec()` in `src/user/user_objects.cc`. 

When the model is loaded using the <attach> block, MuJoCo runs CopyList() for each of the tendons:
`candidate->model = this;`
`candidate->CopyFromSpec();`
`candidate->ResolveReferences(this);`

This shows you how the model gets updated, but `CopyFromSpec()` never propagated that to the `mjCWrap` objects in `path[]`. This means that each wrap had like a null or old pointer from the child.

This solution also wasn't exactly new since the `mjCTendon::SetMode()` function already does this; I'm not sure why `CopyFromSpec` wasn't doing the same thing.

I lowkey decided on omitting the test-cases for this since it is such a small fix if that's ok.

### DIFF
--- a/src/user/user_objects.cc
+++ b/src/user/user_objects.cc
@@ -6441,8 +6441,9 @@ void mjCTendon::CopyFromSpec() {
   material_ = spec_material_;
   userdata_ = spec_userdata_;
 
-  // clear precompiled
+  // propagate model pointer to wraps and clear precompiled
   for (int i=0; i < path.size(); i++) {
+    path[i]->model = model;
     if (path[i]->Type() == mjWRAP_CYLINDER) {
       path[i]->spec.type = mjWRAP_SPHERE;
     }


quagla (approved):

LouRohanNV:
@havess @quagla, do you guys need anything else to merge this fix?

quagla:
Sorry I was out of office and I missed this, I am merging the clone #3406 which also has tests. Thanks for catching the fix!

Issue #2145, opened 2024-10-16
Adhesion actuator `dyntype`


vaxenburg:
### The feature, motivation and pitch

Is there a good reason the `dyntype` of adhesion actuators can only be `none`? Can't we expose `dyntype` as another custom attribute of the adhesion actuators, like e.g. gain? For example, having a `filter` dynamics for adhesion could be quite useful (and realistic in some cases).

yuvaltassa:
This is a reasonable ask, however this should be possible now by saving an adhesion actuator as <general> and then adding a dyntype. Yes, this is not as elegant as <adhesion>, but should just work.

So unless you're saying that you tried this and it failed, this request is basically about nicer MJCF cosmetics.

If you tried it and it failed then that's a bug 🙂

vaxenburg:
For now, I do something like this, but it's not very nice:
```python
adhesion_actuator.dclass.general.dyntype = 'filter'
adhesion_actuator.dclass.general.dynprm = (0.01,)
```
Yeah, this is not a bug report, just an enhancement request

vaxenburg:
@yuvaltassa is it me or is your comment missing a bit or two? What do you mean by "saving an adhesion actuator as"?

yuvaltassa:
I mean that if you have an XML with `<adhesion>` and you save the XML then you get a `<general>`.

But the trick you're using with default classes is even better, if you have procedural code.

That said, your enhancement request is valid and would not be hard to add. There are two options really:
1. Expose `dyntype`
2. If we think only filters are useful, expose `timeconst`, like in `<position>`.

WDYT?

vaxenburg:
Thanks Yuval, that sounds good. For all I know, for the `<adhesion>` shortcut we don't need more than `filterexact`, so that'd be my vote probably. And if you want more, you do it with `<general>`. How about that?

mar-yan24:
@yuvaltassa it has been a while since this enhancement was opened but this enhancement but it caught my eye and I am quite interested on implementing it. If you haven't done too much work on it would it be possible for me to get started on a PR for this feature? I did some preliminary tracing and I think exposing `timeconst` just requires some changes to allow for a new parameter in `mjs_setToAdhesion()`.

mar-yan24:
Got a little ahead of myself and went and implemented the feature. Hopefully nothing looks too bad and the feature is still somewhat desired 😄.

yuvaltassa:
[...]

PR #3162, opened 2026-03-10
full adhesion timeconst feature update

mar-yan24:
## Summary
This pull request covers a basic implementation of enhancement suggestion #2145, the addition of a `timeconst` for adhesion actuation. This addition should allow for easy control over the control signal for "sticky forces". 

The basic reason why this feature is helpful is because the introduction of a time constant adds a simple low-pass filter between control signal and actual force, as @yuvaltassa and @vaxenburg alluded to earlier. 

## Main Changes
I added the `timeconst` parameter for `mjs_setToAdhesion()`. If its positive, the actuator gets a filter, `dyntype=mjDYN_FILTER`, if its 0, behavior is unchanged. This was the main change, all others following this are supporting changes.

I changed the XML schema and parser to add `timeconst` to the list of allowed attributes so you can write it within the model file. I also added python and wasm bindings to expose the parameter so python users can call `actuator.set_to_adhesion(gain=100, timeconst=0.03)`. This should keep the api in sync with xml.

I also described the change in the documentation and added some basic tests.

This change also follows the same existing schema within `<cylinder>` and `<position>` actuators that have the `timeconst` attribute, so this was an existing feature, just extended it to one more actuator type.

Let me know if anything looks awry and I'll take a look.

### DIFF
--- a/doc/XMLreference.rst
+++ b/doc/XMLreference.rst
@@ -6151,14 +6151,16 @@ applying the adhesive force. In the video above, such inactive contacts are blue
 An adhesion actuator's length is always 0. :at:`ctrlrange` is required and must also be nonnegative (no repulsive forces
 are allowed). The underlying :el:`general` attributes are set as follows:
 
-=========== ======= =========== ========
-Attribute   Setting Attribute   Setting
-=========== ======= =========== ========
-dyntype     none    dynprm      1 0 0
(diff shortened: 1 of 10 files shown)


yuvaltassa (approved):

yuvaltassa:
this needs to change to conform to the new schema, let me know if you still want to do this

PR #3168, opened 2026-03-11
Add manual workflow trigger for WASM publishing

MatiasManevi:
This PR adds support for manually triggering the `publish-wasm` workflow via `workflow_dispatch`.

### Changes

* Added a `workflow_dispatch` trigger with a required `tag_version` input.
* Introduced a `VERSION` environment variable in the workflow that resolves to:
  * the provided `tag_version` when triggered manually, or
  * `github.ref_name` when triggered by a tag push.
* Updated `package_wasm` in `build_steps.sh` to use the `VERSION` environment variable instead of deriving the version directly from `GITHUB_REF`.
* Kept a fallback to `${GITHUB_REF#refs/tags/}` for backward compatibility if `VERSION` is not set.

### Motivation

Previously the workflow could only run when a tag was pushed. Adding `workflow_dispatch` allows maintainers to:

* manually publish a release when needed,
* retry publishing without creating a new tag,
* test the release pipeline more easily.

Using a dedicated `VERSION` variable also avoids overriding GitHub-provided environment variables such as `GITHUB_REF`, which keeps the workflow behavior clearer and more predictable.

### Behavior

| Trigger            | Version source      |
| ------------------ | ------------------- |
| Tag push (`1.2.3`) | `github.ref_name`   |
| Manual dispatch    | `tag_version` input |

No changes to the resulting npm package contents or publish process.

### DIFF
--- a/.github/workflows/build_steps.sh
+++ b/.github/workflows/build_steps.sh
@@ -222,7 +222,7 @@ package_wasm() {
     cp wasm/package.npm.json wasm/dist/package.json
     cp wasm/README.md wasm/dist/README.md
 
-    VERSION=${GITHUB_REF#refs/tags/}
+    VERSION="${VERSION:-${GITHUB_REF#refs/tags/}}"
     npm --prefix wasm/dist version "${VERSION}" --no-git-tag-version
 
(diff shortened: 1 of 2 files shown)


mmossg (approved):
LGTM, thanks!

PR #3200, opened 2026-03-27
Update mjwarp render doc after flex rendering improvements

StafaH:
https://mujoco--3200.org.readthedocs.build/en/3200/mjwarp/index.html


thowell:
thanks @StafaH!

PR #3202, opened 2026-03-30
[Unity plugin] Update to URP for materials

Balint-H:
The built-in pipeline is beginning to be deprecated, projects will use URP by default. This PR swaps the materials to use the URP lit materials, so users don't have to manually update/reassign each material to their imported MJCFs.

A couple other housekeeping updates are also included:
- STL/OBJ file name parsing is now no longer case sensitive (many menagerie repos use uppercase extensions)
- The mouse spring utility was throwing errors due to GUI functionality becoming more restricted to specific contexts. We now store the result of the ID call while we are in the appropriate context.

### DIFF
--- a/unity/Editor/Components/MjMouseSpring.cs
+++ b/unity/Editor/Components/MjMouseSpring.cs
@@ -31,13 +31,17 @@ namespace Mujoco {
 
     private Color _translucentRed = new Color(1, 0, 0, 0.1f);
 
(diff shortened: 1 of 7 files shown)


Balint-H:
Could I ask for approval @erez-tom?

erez-tom (approved):

yuvaltassa (approved):

PR #3205, opened 2026-03-31
Fix GCC 15 compatibility warnings

totrytakeoff:
- engine_print.c: change 'char* c' to 'const char* c' since strchr() returns const char*; fixes -Werror=discarded-qualifiers
- engine_plugin.cc: add bounds check for nattribute to prevent alloc-size overflow; fixes -Walloc-size-larger-than


totrytakeoff:
Rebased this PR onto the latest main and cleaned up the history to a single commit. The engine_print.c const fix is already present upstream now, so this PR now only keeps the remaining engine_plugin.cc bounds check for the GCC 15 alloc-size warning. CLA and check-changes are passing.

PR #3225, opened 2026-04-15
Fix AttributeError in Renderer.__del__ on partial construction

### LINKED ISSUE #3213: Renderer.__del__ raises AttributeError: '_mjr_context' if __init__ fails partway through
## Bug

`mujoco.rendering.classic.renderer.Renderer.__init__` assigns `self._gl_context` and then constructs `self._mjr_context = mujoco.MjrContext(model, font_scale.value)`. If anything between those two lines raises (e.g. an invalid model, GPU context creation failure, EGL/OSMesa init issue, segfault during `MjrContext` construction), the `_mjr_context` attribute is never assigned.

The destructor `__del__` then calls `close()`, which accesses `self._mjr_context` unconditionally:

```python
def close(self) -> None:
    ...
    if self._gl_context:
        self._gl_context.free()
    self._gl_context = None
    if self._mjr_context:                # ← AttributeError here
        self._mjr_context.free()
    self._mjr_context = None
```

The result is a noisy `AttributeError` during garbage collection that masks the *real* `__init__` failure.

## Reproduction

Tested on `mujoco==3.6.0`, Python 3.12, Windows 11. The bug is also present in the current `main` branch ([renderer.py L70-L92 + L324-L329](https://github.com/google-deepmind/mujoco/blob/main/python/mujoco/rendering/classic/renderer.py)).

```python
from mujoco.rendering.classic.renderer import Renderer

class _Repro(Renderer):
    def __init__(self):
        # Simulate __init__ failing AFTER `self._gl_context = None`
        # but BEFORE `self._mjr_context = MjrContext(...)`.
        # In real code this happens when MjrContext construction
        # raises (invalid model, GPU init failure, etc.).
        self._gl_context = None
        # _mjr_context intentionally NOT assigned

r = _Repro()
del r
```

Output:
```
Exception ignored in: <function Renderer.__del__ at 0x...>
Traceback (most recent call last):
  File ".../mujoco/rendering/classic/renderer.py", line 339, in __del__
    self.close()
  File ".../mujoco/rendering/classic/renderer.py", line 327, in close
    if self._mjr_context:
       ^^^^^^^^^^^^^^^^^
AttributeError: 'Renderer' object has no attribute '_mjr_context'
```

## Expected behavior

`Renderer.__del__` should be safe to call on a partially-constructed instance and should never raise. The original `__init__` exception should be the only thing the user sees.

## Suggested fix

Initialize both attributes to `None` at the very top of `__init__`, before anything that can fail:

```python
def __init__(self, model, ...):
    # Defensive: assign None first so __del__ is safe even if construction fails
    self._gl_context = None
    self._mjr_context = None

    # ... existing __init__ body ...
```

That's a one-line fix. Alternatively, `close()` could use `getattr(self, "_mjr_context", None)` and `getattr(self, "_gl_context", None)`, but pre-assigning in `__init__` is cleaner because it documents the invariant explicitly.

## Real-world impact

When this triggers (typically on a CI agent without a GPU, or right after a driver/CUDA upgrade), the user sees the `AttributeError` before they see the *actual* underlying failure (`MjrContext` rejected the model, EGL couldn't initialize, etc.), making the real problem invisible.

We hit it intermittently in our test suite on a Windows machine after some unrelated DLL loading order changes — chasing the AttributeError took about 30 minutes before we realized the destructor was hiding the real exception. We monkey-patched `Renderer.close()` defensively in our `conftest.py`, but the upstream fix is trivial enough that it would be nice to land in the next mujoco release.

## Environment

- mujoco: 3.6.0 (also confirmed in `main` branch)
- Python: 3.12 (also reproducible on 3.10, 3.11)
- OS: Windows 11
- The bug is platform-independent — the attribute initialization order is the same in all build flavors.

devshahofficial:
## Summary
- Pre-initialize `self._gl_context` and `self._mjr_context` to `None` at the top of `Renderer.__init__` so `close()` (called from `__del__`) is safe on a partially-constructed instance.
- Add a regression test that uses `sys.unraisablehook` to assert `__del__` raises nothing when `__init__` fails early.

## Why
If `Renderer.__init__` raises before the rendering contexts are assigned — e.g. `width > model.vis.global_.offwidth` triggers a `ValueError` near the top of `__init__`, or `MjrContext` construction fails later — `__del__` calls `close()` which unconditionally accesses `self._gl_context` and `self._mjr_context`, raising `AttributeError`. This surfaces as a noisy `Exception ignored in <function Renderer.__del__ ...>` traceback during garbage collection and masks the real `__init__` exception from the user.

Reproduction on `mujoco==3.6.0`:

```python
import mujoco

xml = """
<mujoco>
  <visual>
    <global offwidth="50" offheight="50"/>
  </visual>
  <worldbody/>
</mujoco>
"""
try:
    mujoco.Renderer(model, height=50, width=200)
except ValueError:
    pass
# Exception ignored in: <function Renderer.__del__ at 0x...>
# AttributeError: 'Renderer' object has no attribute '_gl_context'
```

The fix pre-initializes both attributes to `None` at the very top of `__init__`, before any code that can raise. The new regression test exercises the `width > offwidth` path — it fails on `main` with `['AttributeError']` captured via `sys.unraisablehook` and passes with the fix applied.

Fixes #3213.

### DIFF
--- a/python/mujoco/renderer_test.py
+++ b/python/mujoco/renderer_test.py
@@ -14,6 +14,9 @@
 # ==============================================================================
 """Tests for the MuJoCo renderer."""
 
+import gc
+import sys
+
 from absl.testing import absltest
 from absl.testing import parameterized
 import mujoco
@@ -156,6 +159,47 @@ class MuJoCoRendererTest(parameterized.TestCase):
       self.assertNotEqual(failing_render_size, render_size)
       with self.assertRaises(ValueError):
         renderer.render(out=np.zeros((*failing_render_size, 3), np.uint8))
+
+  def test_renderer_del_safe_when_init_fails_early(self):
+    """Regression test for #3213.
+
+    Renderer.__del__ must be safe on a partially-constructed instance when
+    __init__ raises before the rendering contexts are assigned. Previously,
+    AttributeError from __del__ masked the real __init__ exception.
+    """
+    xml = """
+<mujoco>
+  <visual>
+    <global offwidth="50" offheight="50"/>
+  </visual>
+  <worldbody/>
+</mujoco>
+"""
+    model = mujoco.MjModel.from_xml_string(xml)
+
+    # Capture any exception raised from __del__ on the partially-constructed
+    # Renderer. Without the fix, __del__ raises AttributeError, which is
+    # funneled through sys.unraisablehook.
+    unraisable = []
+    old_hook = sys.unraisablehook
+    sys.unraisablehook = lambda args: unraisable.append(args)
+    try:
+      # width > offwidth raises ValueError in __init__ before
+      # self._gl_context is assigned.
+      with self.assertRaises(ValueError):
+        mujoco.Renderer(model, height=50, width=200)
+      gc.collect()
+    finally:
+      sys.unraisablehook = old_hook
+
+    self.assertEqual(
+        [u.exc_type.__name__ for u in unraisable],
+        [],
+        msg=(
+            'Renderer.__del__ raised on a partially-constructed instance; '
+            'see #3213.'
+        ),
+    )
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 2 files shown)


devshahofficial:
@haroonq @yuvaltassa fixed this. happy to iterate!

haroonq (approved):

haroonq (approved):

PR #3226, opened 2026-04-16
Fix float comparision in tests for non-x86 CPUs

hannesbraun:
On non-x86 CPUs (64-bit ARM, 64-bit RISC-V, ...), this assertion fails (tested with Alpine Linux). Increasing the tolerance in this case helps for the tests to pass on those platforms as well.


yuvaltassa:
This was fixed a while back

PR #3235, opened 2026-04-19
MJX WARP Segmentation

tkelestemur:
## Summary

Expose MuJoCo Warp segmentation rendering through the `mujoco.mjx` interface.

This PR adds an MJX-facing segmentation render path so users can request segmentation from the Warp renderer and recover it from the packed render outputs, alongside the existing MJX render APIs.

## What Changed

This PR:
- adds `mjx.render_with_segmentation(...)`
- adds `mjx.get_segmentation(...)`
- exports the new APIs from `mujoco.mjx`
- threads segmentation buffer metadata through the MJX/Warp render context
- adds tests for segmentation unpacking and Warp-side integration
- updates the MJX rendering docs with segmentation usage and output semantics

It also includes a small cleanup pass around the new API:
- factors shared `RenderContextPytree` validation into one helper
- fixes the public render return annotations so they match the actual tuple outputs
- adds an early error when `render_with_segmentation(...)` is called with a render context that has no segmentation-enabled cameras
- clarifies that segmentation returns geom IDs, with `-1` for background and `-2` for flex bodies

## Why

MuJoCo Warp already supports segmentation internally, but MJX did not expose that output through its JAX-facing rendering interface.

As a result, MJX users could render RGB and depth with the Warp backend, but could not access per-pixel segmentation labels from the same path. This change closes that gap and makes Warp segmentation available directly from MJX.

## User-Facing Behavior

After this change, MJX users can:
- create a render context with segmentation enabled
- call `mjx.render_with_segmentation(...)`
- recover per-camera segmentation images with `mjx.get_segmentation(...)`

The returned segmentation image contains MuJoCo geom IDs. Background pixels are `-1`, and flex body pixels are `-2`.

## Implementation Note

`mjx/mujoco/mjx/warp/render.py` is normally autogenerated. I edited that file manually in this PR because I do not have access to the shim code generator used to produce it.

The manual edits were kept narrowly scoped to the segmentation plumbing and follow the existing generated structure as closely as possible.

## Validation

Locally, I verified:
- `git diff --check`
- `pre-commit` on the PR diff
- MJX packaging via `python -m build`
- targeted MJX render utility tests for the new segmentation unpacking path

This PR is intentionally scoped to exposing Warp segmentation through MJX only.


devshahofficial:
Nice addition. I can help with an outside validation pass focused on user-facing behavior: multi-camera segmentation shape/packing, background/flex sentinel values, and whether `get_segmentation` is ergonomic from a JAX caller perspective. One thing I would check carefully is the manually edited generated Warp file; if maintainers expect that file to stay generated, it may help to isolate the intended generator delta or document the temporary manual patch clearly.

btaba (approved):

PR #3237, opened 2026-04-20
Generating Mujoco MJCF XSD schema from the parser

### LINKED ISSUE #6: Publish XML schema for MJCF
MJCF is [well-documented](http://mujoco.org/book/XMLreference.html), but a formal schema is not available. A schema based on e.g. [XSD](https://www.w3schools.com/xml/schema_intro.asp), [relaxNG](https://relaxng.org/), or [schematron](https://www.schematron.com/) would help 3rd parties write their own validators. For example DeepMind's `mjcf` package required its own [custom schema](https://github.com/deepmind/dm_control/blob/master/dm_control/mjcf/schema.xml) in order to properly verify and manipulate MJCF elements in Python.

julien-blanchon:
Adds an auto-generated MJCF XSD, built from the parser tables in `src/xml/` and enriched with defaults and docs from `doc/XMLreference.rst`. 

This is intended for IDE/LLM editing support (elements, attributes, enums), not as a replacement for compiler validation.

## What’s included

* XSD generator (`mj_printSchemaXSD`) based on `MJCF[]` + attribute-type table
* CLI tool to dump the raw schema (`sample/xmlschema.cc`)
* Enrichment script to pull defaults/docs from RST and detect drift (`--strict`)
* Docs + contributing updates (regen + editor integration)

Closes #6. Refs julien-blanchon/mujoco-schema#1.

## Regeneration

```bash
# Build the xmlschema cli
cmake --build build --target xmlschema
# Build the base schema from the parser data
./build/bin/xmlschema /tmp/raw.xsd
# Eventually enrich the schema with documentation using the `XMLreference.rst`
python doc/mjcf_schema_enrich.py --in /tmp/raw.xsd \
    --rst doc/XMLreference.rst --out doc/mjcf.xsd --strict --report
```

## Test plan

- [x] Builds on Linux / macOS / Windows
- [x] The generated schema (`doc/mjcf.xsd`) validates against XMLSchema (valid schema)
- [x] `./model/**/*.xml` validate against the schema
- [x] Output is deterministic

## Decisions

**Where to serve `mjcf.xsd` ?** 
Should we keep store it in the repo at `doc/mjcf.xsd`, or as a Github release artifact ? Or something else ?

**Adding CI ?**
Should we add a lightweight CI job to:
  * build `xmlschema`
  * run enricher with `--strict`
  * diff against committed XSD
  * validate schema + sample models (and eventually `google-deepmind/mujoco_menagerie` model too)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -126,10 +126,15 @@ regenerate it:
 ```bash
 cmake --build build --target xmlschema
 ./build/bin/xmlschema /tmp/raw.xsd
-python3 doc/mjcf_schema_enrich.py \
+./doc/mjcf_schema_enrich.py \
     --in /tmp/raw.xsd --rst doc/XMLreference.rst --out doc/mjcf.xsd \
     --strict --report
 ```
+
+`doc/mjcf_schema_enrich.py` is a [uv](https://docs.astral.sh/uv/) PEP 723
+script; its shebang provisions a matching Python interpreter, so the only
+prerequisite is `uv` on `PATH`. Run it as `./doc/mjcf_schema_enrich.py ...`
+or `uv run doc/mjcf_schema_enrich.py ...`.
 
 `--strict` exits non-zero on drift between the parser, the attribute table
 (`src/xml/xml_native_schema.cc`), and the RST — fix warnings before
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -114,3 +114,24 @@ Thanks for improving MuJoCo!
 
 This project follows Google's
 [Open Source Community Guidelines](https://opensource.google/conduct/).
+
+
+## Regenerating the MJCF XSD
+
+The MJCF XSD shipped with releases lives at `doc/mjcf.xsd` and is
+auto-generated from `src/xml/` and `doc/XMLreference.rst`. If your change
+touches MJCF parsing, adds an attribute, or edits the reference docs,
+regenerate it:
+
+```bash
+cmake --build build --target xmlschema
+./build/bin/xmlschema /tmp/raw.xsd
+python3 doc/mjcf_schema_enrich.py \
+    --in /tmp/raw.xsd --rst doc/XMLreference.rst --out doc/mjcf.xsd \
+    --strict --report
+```
+
+`--strict` exits non-zero on drift between the parser, the attribute table
+(`src/xml/xml_native_schema.cc`), and the RST — fix warnings before
+committing. See the header comment in `xml_native_schema.cc` for which files
+to edit when adding a new attribute.
--- a/doc/XMLreference.rst
+++ b/doc/XMLreference.rst
@@ -7,6 +7,31 @@ Introduction
 
 This chapter is the reference manual for the MJCF modeling language used in MuJoCo.
 
+.. _XSDSchema:
+
+Editor integration (XSD)
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+The MJCF grammar is also published as an `XSD <https://www.w3.org/XML/Schema>`_
+schema, generated directly from the parser source and the prose below. Pointing
+your editor at it gives you autocompletion, inline documentation, and validation
+while writing MJCF.
+
+For VS Code, install the
+`Red Hat XML extension <https://marketplace.visualstudio.com/items?itemName=redhat.vscode-xml>`_
+and reference the schema from the root element:
+
+.. code-block:: xml
+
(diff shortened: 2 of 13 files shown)


julien-blanchon:
You can found the resulting schema here if you want to test it in your IDE directly: https://raw.githubusercontent.com/julien-blanchon/mujoco/refs/heads/schema-autogen-xsd/doc/mjcf.xsd

For vscode you need https://github.com/redhat-developer/vscode-xml + 
```
<mujoco xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/julien-blanchon/mujoco/refs/heads/schema-autogen-xsd/doc/mjcf.xsd">
  ```
In the header

devshahofficial:
This is a useful direction. I can help test the generated schema against a broader set of MJCFs, especially `model/` plus a subset of Menagerie models. My instinct would be to keep `doc/mjcf.xsd` committed if deterministic generation is enforced in CI; otherwise users would not have a stable URL for editor integration. A lightweight CI check that regenerates + diffs the schema feels like the right guardrail. Happy to run a validation pass and report concrete schema misses if that would help review.

julien-blanchon:
Broader testing will be greatly appreciate ! 
But maybe let's wait for GDM people inputs on this before.

julien-blanchon:
Any follow up on this ? 

@devshahofficial do not hesitate to test, expecially on some edge case / minor mujoco feature. I've done some testing already on all the ./model mujoco models from this repo, and some of the Menagerie models.

About commiting the `doc/mjcf.xsd`, I was also thinking releasing the `mjcf.xsd` along side the mujoco build inside of the released (see https://github.com/google-deepmind/mujoco/releases). So the schema CI doesn't have to trigger at every commit (just on release), we get both a versioned url with the version tag in it (`https://github.com/google-deepmind/mujoco/releases/download/<version_tag>/mjcf.xsd`), and a always up to date `latest` version (`https://github.com/google-deepmind/mujoco/releases/latest/download/mjcf.xsd`).

julien-blanchon:
Any follow up on this ?

devshahofficial:
Sorry for the slow follow-up. I tested the generated `mjcf.xsd` from your `schema-autogen-xsd` branch with `xmllint --schema`.

Results:
- `model/**/*.xml`: 75/77 pass.
- local `mujoco_menagerie_go2` subset: 4/4 pass.
- all repo XML files with `<mujoco>` root: 386/395 pass.

The useful misses I found:
- `flexcomp dof="2d"` is rejected, but it is documented and present in `fdof_map`.
- `composite size="1"` is rejected as not length 3; the reader uses `ReadAttr(..., exact=false)`, so shorter lists are accepted there.
- `replicate prefix="..."` is rejected, but `prefix` is parsed for replicate.
- `compiler conflict="merge|warning|error"` is rejected, but it is documented and parsed.

There are also a couple of test-data files with non-whitespace text nodes that the schema rejects; I’d treat those separately from the attribute coverage gaps above.

devshahofficial:
I took another pass on the concrete schema misses from my earlier validation and opened a stacked draft here: #3410.

The final patch adds `compiler conflict`, `replicate prefix`, and makes top-level `composite size` up-to-3 instead of exact-3. After rebasing onto current `main`, `flexcomp dof="2d"` is already covered by the updated `fdof_map`.

Validated the generated raw XSD with `xmllint` against a snippet covering all four cases, and the enrichment script still runs in non-strict/report mode. Strict mode is still blocked by the existing doc-drift warnings, so I left that out of scope.

julien-blanchon:
Hey @devshahofficial sorry I missed your first follow up. I just had a look at https://github.com/google-deepmind/mujoco/pull/3410 it look good to me. 

For the strict mode I don't think it's a big issue.

About the `mjcf.xsd` hosting what would be the best option for the mujoco team ? Having the schema commited inside the repo with a pre-commit ? Or a CI workflow ? Or eventually having the CI releasing the schema as release artefact in the Github release bucket ? 

I would lean more on hosting it on github release personally

devshahofficial:
My preference is to keep `doc/mjcf.xsd` in the repo and have CI regenerate + diff it. That gives editor users a stable raw URL on every tagged commit, keeps schema changes reviewable beside parser changes, and catches drift before release. A release asset could still be added as a convenience, but I would treat it as distribution rather than the source of truth.

That said, the hosting/update contract is a maintainer call. Would the MuJoCo team prefer committed + CI-checked, or release-only? I’m happy to adapt #3410 once that direction is clear.

yuvaltassa:
This is now fixed, I did it the hard way 🙂 

Please let me know what you think

PR #3240, opened 2026-04-22
Add config on msvc for mujoco with usd

jingkaimori:
This pr add compile configurations to build mujoco with `MUJOCO_WITH_USD=on` on msvc.

- expand cmake configurations
- solve compile error reported by openUSD `TF_REGISTRY_FUNCTION` macros
- explode WriteSpecToData as api of mujoco

### DIFF
--- a/src/experimental/usd/plugins/mjcf/mujoco_to_usd.h
+++ b/src/experimental/usd/plugins/mjcf/mujoco_to_usd.h
@@ -27,7 +27,7 @@ namespace usd {
 //   layer: SdfLayerRefPtr that will be written to.
 //   skip_elems_from_usd: If true, skips over elements in the spec that have a usd_prim_path
 //   custom attribute set.
-bool WriteSpecToData(mjSpec* spec, pxr::SdfLayerRefPtr layer, bool skip_elems_from_usd = false);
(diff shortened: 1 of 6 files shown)


src/experimental/usd/plugins/mjcf/mujoco_to_usd.h
```diff
@@ -27,7 +27,7 @@ namespace usd {
 //   layer: SdfLayerRefPtr that will be written to.
 //   skip_elems_from_usd: If true, skips over elements in the spec that have a usd_prim_path
 //   custom attribute set.
-bool WriteSpecToData(mjSpec* spec, pxr::SdfLayerRefPtr layer, bool skip_elems_from_usd = false);
+MJAPI bool WriteSpecToData(mjSpec* spec, pxr::SdfLayerRefPtr layer, bool skip_elems_from_usd = false);
```
havess, line 30: Why do we need to add MJAPI here?

jingkaimori: Because this file is required by not only mjcf plugin, but usd writer in `mujoco` module also needs this api.
MJCF plugin must treat mujoco as its dependency, rather than mujoco relies on mjcf plugin.
So this api is added into sources of `mujoco` instead of mjcf plugin.
To declare this symbol correctly and make it visible to mjcf plugin, `MJAPI` macro is added to provide proper dll export declaration.

PR #3247, opened 2026-04-28
Handle rollout MuJoCo errors as warnings

### LINKED ISSUE #2704: [rollout] Handle exceptions as warnings

Dear MuJoCo-team,

thank you for providing and maintaining such a great simulator. Currently, I am working as an SDE intern at Amazon robotics, using MuJoCo for large-scale simulations. Within our internal software stack, we are using the ``rollout()`` function for large-scale batched workflows.

**Problem:**
In the early stages of our simulations, the computed trajectories are often unstable, occasionally leading to computationally infeasible states (e.g., singular Hessian) in one of the batches, resulting in an exception. Consequently, when the simulation crashes, all other batches are lost.

**Potential Solution:**
To address this issue, could the ``rollout()`` function be adjusted to handle exceptions in a similar way as warnings? Specifically, when an exception occurs in a batch, the remainder of that trajectory could be filled with the current state and sensor values.

Thank you very much in advance, looking forward to the discussion.

Best,
Florian

### Alternatives

_No response_

### Additional context

_No response_

devshahofficial:
## Summary
- add `raise_on_error` to rollout so existing strict error behavior remains the default
- when `raise_on_error=False`, intercept MuJoCo fatal errors during a rollout, fill the failed trajectory remainder from the current state and sensor data, and continue other batches
- document the new tolerant path and cover both single-threaded and threaded rollout behavior

Fixes #2704

## Validation
- `python3 -m py_compile python/mujoco/rollout.py python/mujoco/rollout_test.py`
- built local MuJoCo, generated the Python sdist, built and installed a wheel from this checkout
- `MUJOCO_GL=disable /tmp/mujoco-rollout-venv/bin/python -m pytest -q --pyargs mujoco.rollout_test` (`67 passed, 1 skipped`)

### DIFF
--- a/python/mujoco/rollout.cc
+++ b/python/mujoco/rollout.cc
@@ -14,10 +14,8 @@
 
 #include <algorithm>
 #include <cstddef>
-#include <exception>
 #include <iostream>
 #include <memory>
-#include <mutex>
 #include <optional>
 #include <sstream>
 #include <vector>
@@ -135,6 +133,7 @@ void _unsafe_rollout(std::vector<const mjModel*>& m, mjData* d, int start_roll,
     }
 
     // roll out trajectory
+    bool nerror = false;
     for (size_t t = 0; t < nstep; t++) {
       // check for warnings
       bool nwarning = false;
@@ -143,37 +142,6 @@ void _unsafe_rollout(std::vector<const mjModel*>& m, mjData* d, int start_roll,
           nwarning = true;
           break;
         }
-      }
-
-      bool nerror = false;
-
-      size_t step = r*static_cast<size_t>(nstep) + t;
-
-      // controls
-      if (control && !nwarning) {
-        mj_setState(m[r], d, control + step*ncontrol, control_spec);
-      }
-
-      // step
-      if (!nwarning) {
-        try {
-          InterceptMjErrors(mj_step)(m[r], d);
-        } catch (const FatalError&) {
-          if (raise_on_error) {
-            throw;
-          }
-          nerror = true;
-        }
-      }
-
-      // copy out new state
-      if (state && !nwarning && !nerror) {
-        mj_getState(m[r], d, state + step*nstate, mjSTATE_FULLPHYSICS);
-      }
-
-      // copy out sensor values
-      if (sensordata && !nwarning && !nerror) {
-        mju_copy(sensordata + step*nsensordata, d->sensordata, nsensordata);
       }
 
       // if any warnings or handled errors, fill remaining outputs with current outputs, break
@@ -188,6 +156,41 @@ void _unsafe_rollout(std::vector<const mjModel*>& m, mjData* d, int start_roll,
           }
         }
         break;
+      }
+
+      size_t step = r*static_cast<size_t>(nstep) + t;
+
+      // controls
+      if (control) {
+        mj_setState(m[r], d, control + step*ncontrol, control_spec);
+      }
+
+      // step
+      try {
+        InterceptMjErrors(mj_step)(m[r], d);
+      } catch (const FatalError&) {
+        if (raise_on_error) {
+          throw;
+        }
+        // Write this step before continuing to the fill path.
+        if (state) {
+          mj_getState(m[r], d, state + step*nstate, mjSTATE_FULLPHYSICS);
+        }
+        if (sensordata) {
+          mju_copy(sensordata + step*nsensordata, d->sensordata, nsensordata);
+        }
+        nerror = true;
+        continue;
+      }
+
+      // copy out new state
+      if (state) {
+        mj_getState(m[r], d, state + step*nstate, mjSTATE_FULLPHYSICS);
+      }
+
+      // copy out sensor values
+      if (sensordata) {
+        mju_copy(sensordata + step*nsensordata, d->sensordata, nsensordata);
       }
     }
   }
@@ -204,53 +207,34 @@ void _unsafe_rollout_threaded(std::vector<const mjModel*>& m,
   int nfulljobs = nbatch / chunk_size;
   int chunk_remainder = nbatch % chunk_size;
   int njobs = (chunk_remainder > 0) ? nfulljobs + 1 : nfulljobs;
-  std::exception_ptr rollout_error;
-  std::mutex rollout_error_mutex;
 
   // Reset the pool counter
   pool->ResetCount();
 
   // schedule all jobs of full (chunk) size
   for (int j = 0; j < nfulljobs; j++) {
-    auto task = [=, &m, &d, &rollout_error, &rollout_error_mutex](void) {
+    auto task = [=, &m, &d](void) {
       int id = pool->WorkerId();
-      try {
-        _unsafe_rollout(m, d[id], j*chunk_size, (j+1)*chunk_size,
-          nstep, control_spec, state0, warmstart0, control, state, sensordata,
-          raise_on_error);
-      } catch (...) {
-        std::lock_guard<std::mutex> lock(rollout_error_mutex);
-        if (!rollout_error) {
-          rollout_error = std::current_exception();
-        }
-      }
+      _unsafe_rollout(m, d[id], j*chunk_size, (j+1)*chunk_size,
+        nstep, control_spec, state0, warmstart0, control, state, sensordata,
+        raise_on_error);
     };
     pool->Schedule(task);
   }
 
   // schedule any remaining jobs of size < chunk_size
   if (chunk_remainder > 0) {
-    auto task = [=, &m, &d, &rollout_error, &rollout_error_mutex](void) {
-      try {
-        _unsafe_rollout(m, d[pool->WorkerId()], nfulljobs*chunk_size,
-          nfulljobs*chunk_size+chunk_remainder,
-          nstep, control_spec, state0, warmstart0, control, state, sensordata,
-          raise_on_error);
-      } catch (...) {
-        std::lock_guard<std::mutex> lock(rollout_error_mutex);
-        if (!rollout_error) {
-          rollout_error = std::current_exception();
-        }
-      }
+    auto task = [=, &m, &d](void) {
+      _unsafe_rollout(m, d[pool->WorkerId()], nfulljobs*chunk_size,
+        nfulljobs*chunk_size+chunk_remainder,
+        nstep, control_spec, state0, warmstart0, control, state, sensordata,
+        raise_on_error);
     };
     pool->Schedule(task);
   }
 
   // wait for counter to increment up to number of jobs submitted by this thread
   pool->WaitCount(njobs);
-  if (rollout_error) {
-    std::rethrow_exception(rollout_error);
-  }
 }
 
 // NOLINTEND(whitespace/line_length)
@@ -351,12 +335,12 @@ class Rollout {
         } else {
           chunk_size_final = *chunk_size;
         }
-        _unsafe_rollout_threaded(
+        InterceptMjErrors(_unsafe_rollout_threaded)(
             model_ptrs, data_ptrs, nbatch, nstep, control_spec, state0_ptr,
             warmstart0_ptr, control_ptr, state_ptr, sensordata_ptr,
             this->pool_.get(), chunk_size_final, raise_on_error);
       } else {
-        _unsafe_rollout(
+        InterceptMjErrors(_unsafe_rollout)(
             model_ptrs, data_ptrs[0], 0, nbatch, nstep, control_spec,
             state0_ptr, warmstart0_ptr, control_ptr, state_ptr,
             sensordata_ptr, raise_on_error);
--- a/python/mujoco/rollout.cc
+++ b/python/mujoco/rollout.cc
@@ -172,15 +172,7 @@ void _unsafe_rollout(std::vector<const mjModel*>& m, mjData* d, int start_roll,
         if (raise_on_error) {
           throw;
         }
-        // Write this step before continuing to the fill path.
-        if (state) {
-          mj_getState(m[r], d, state + step*nstate, mjSTATE_FULLPHYSICS);
-        }
-        if (sensordata) {
-          mju_copy(sensordata + step*nsensordata, d->sensordata, nsensordata);
-        }
         nerror = true;
-        continue;
       }
 
       // copy out new state
--- a/python/mujoco/rollout.cc
+++ b/python/mujoco/rollout.cc
@@ -18,6 +18,7 @@
 #include <memory>
 #include <optional>
 #include <sstream>
+#include <utility>
 #include <vector>
 
 #include <mujoco/mujoco.h>
@@ -66,7 +67,33 @@ Roll out batch of trajectories from initial states, get resulting states and sen
     sensordata         (nbatch x nstep x nsendordata) nbatch trajectories of nstep sensordata arrays
     chunk_size         integer, determines threadpool chunk size. If unspecified, the default is
                            chunk_size = max(1, nbatch / (nthread * 10))
+    raise_on_error     boolean, whether to raise MuJoCo errors. If false, errors fill the
+                           remaining trajectory outputs like warnings.
 )";
+
+void FillRemainingOutputs(const mjModel* m, const mjData* d, size_t r, size_t t,
+                          int nstep, size_t nstate, size_t nsensordata,
+                          mjtNum* state, mjtNum* sensordata) {
+  for (; t < nstep; t++) {
+    size_t step = r*static_cast<size_t>(nstep) + t;
+    if (state) {
+      mj_getState(m, d, state + step*nstate, mjSTATE_FULLPHYSICS);
+    }
+    if (sensordata) {
+      mju_copy(sensordata + step*nsensordata, d->sensordata, nsensordata);
+    }
+  }
+}
+
+template <typename Callable>
+bool TryMjCall(Callable&& callable) {
+  try {
+    InterceptMjErrors(std::forward<Callable>(callable))();
+  } catch (const FatalError&) {
+    return false;
+  }
+  return true;
+}
 
 // C-style rollout function, assumes all arguments are valid
 // all input fields of d are initialised, contents at call time do not matter
@@ -74,7 +101,8 @@ Roll out batch of trajectories from initial states, get resulting states and sen
 void _unsafe_rollout(std::vector<const mjModel*>& m, mjData* d, int start_roll,
                      int end_roll, int nstep, unsigned int control_spec,
                      const mjtNum* state0, const mjtNum* warmstart0,
-                     const mjtNum* control, mjtNum* state, mjtNum* sensordata) {
+                     const mjtNum* control, mjtNum* state, mjtNum* sensordata,
+                     bool handle_mj_errors) {
   // sizes
   size_t nstate = static_cast<size_t>(mj_stateSize(m[0], mjSTATE_FULLPHYSICS));
   size_t ncontrol = static_cast<size_t>(mj_stateSize(m[0], control_spec));
@@ -115,7 +143,17 @@ void _unsafe_rollout(std::vector<const mjModel*>& m, mjData* d, int start_roll,
     }
 
     // set initial state
-    mj_setState(m[r], d, state0 + r*nstate, mjSTATE_FULLPHYSICS);
+    if (handle_mj_errors) {
+      if (!TryMjCall([&]() {
+        mj_setState(m[r], d, state0 + r*nstate, mjSTATE_FULLPHYSICS);
+      })) {
+        FillRemainingOutputs(
+            m[r], d, r, 0, nstep, nstate, nsensordata, state, sensordata);
+        continue;
+      }
+    } else {
+      mj_setState(m[r], d, state0 + r*nstate, mjSTATE_FULLPHYSICS);
+    }
 
     // set warmstart accelerations
     if (warmstart0) {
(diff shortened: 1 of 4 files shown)


devshahofficial:
@haroonq @yuvaltassa this is now marked ready for review. Would one of you, or another active maintainer/contributor, be able to take a look when you have bandwidth? Happy to follow up on any requested changes or adjust the approach if there is a better direction.

yuvaltassa:
@aftersomemath WDYT?

python/mujoco/rollout_test.py
```diff
@@ -711,6 +711,43 @@ def test_intercept_mj_errors(self):
...
+        initial_state,
+        ctrl,
+        raise_on_error=False,
+    )
+
+    expected_state, expected_sensordata = rollout.rollout(
+        model, mujoco.MjData(model), initial_state[1], ctrl[1]
+    )
+    np.testing.assert_array_equal(
+        state[0], np.tile(state[0, 0], (nstep, 1))
```
aftersomemath: The test would be stronger if the exception occurred in the middle of a rollout instead of the first step. If this can be done, I recommend it.

python/mujoco/rollout.py
```diff
@@ -61,6 +61,7 @@ def rollout(
       state: Optional[npt.ArrayLike] = None,
       sensordata: Optional[npt.ArrayLike] = None,
       chunk_size: Optional[int] = None,
+      raise_on_error: bool = True,
```
aftersomemath, line 64: It seems that rollout does not raise any errors when this flag is true. It just fills the buffer with the last state. So a better name is probably `fill_on_true`.

python/mujoco/rollout.cc
```diff
@@ -311,14 +362,29 @@ class Rollout {
         } else {
           chunk_size_final = *chunk_size;
         }
-        InterceptMjErrors(_unsafe_rollout_threaded)(
-            model_ptrs, data_ptrs, nbatch, nstep, control_spec, state0_ptr,
-            warmstart0_ptr, control_ptr, state_ptr, sensordata_ptr,
-            this->pool_.get(), chunk_size_final);
+        if (raise_on_error) {
+          InterceptMjErrors(_unsafe_rollout_threaded)(
```
aftersomemath: It seems that there is no need to use `InterceptMjErrors` on `_unsafe_rollout_threaded` because `_unsafe_rollout_threaded` is already catching all the MuJoCo exceptions, and the return value of `InterceptMjErrors` is not used.

python/mujoco/rollout.cc
```diff
@@ -311,14 +362,29 @@ class Rollout {
...
+              model_ptrs, data_ptrs, nbatch, nstep, control_spec, state0_ptr,
+              warmstart0_ptr, control_ptr, state_ptr, sensordata_ptr,
+              this->pool_.get(), chunk_size_final, true);
+        }
       } else {
-        InterceptMjErrors(_unsafe_rollout)(
-            model_ptrs, data_ptrs[0], 0, nbatch, nstep, control_spec,
-            state0_ptr, warmstart0_ptr, control_ptr, state_ptr, sensordata_ptr);
+        if (raise_on_error) {
+          InterceptMjErrors(_unsafe_rollout)(
```
aftersomemath: Same here.

python/mujoco/rollout.cc
```diff
@@ -114,7 +142,17 @@ void _unsafe_rollout(std::vector<const mjModel*>& m, mjData* d, int start_roll,
     }
 
     // set initial state
-    mj_setState(m[r], d, state0 + r*nstate, mjSTATE_FULLPHYSICS);
+    if (handle_mj_errors) {
+      if (!TryMjCall([&]() {
+        mj_setState(m[r], d, state0 + r*nstate, mjSTATE_FULLPHYSICS);
```
aftersomemath: I don't think it is necessary to catch exception in `mj_setState`. If there is an exception, it is probably our fault as the only thing the call's success depends on is the dimensions, which were already validated. Also, it makes the code verbose.

python/mujoco/rollout.cc
```diff
@@ -141,27 +179,38 @@ void _unsafe_rollout(std::vector<const mjModel*>& m, mjData* d, int start_roll,
...
       }
 
       size_t step = r*static_cast<size_t>(nstep) + t;
 
       // controls
       if (control) {
-        mj_setState(m[r], d, control + step*ncontrol, control_spec);
+        if (handle_mj_errors) {
+          if (!TryMjCall([&]() {
+            mj_setState(m[r], d, control + step*ncontrol, control_spec);
```
aftersomemath, line 173: Same here.

python/mujoco/rollout.cc
```diff
@@ -141,27 +179,38 @@ void _unsafe_rollout(std::vector<const mjModel*>& m, mjData* d, int start_roll,
...
+          }
+        } else {
+          mj_setState(m[r], d, control + step*ncontrol, control_spec);
+        }
       }
 
       // step
-      mj_step(m[r], d);
+      if (handle_mj_errors) {
+        if (!TryMjCall([&]() { mj_step(m[r], d); })) {
```
aftersomemath: I believe this call to `mj_step` is the only place where exceptions need to be caught. If so, then `TryMjCall` can be inlined, and the `FillRemainingOutputs` function can be removed.

Instead of `FillRemainingOutputs` just set a new flag (maybe `nerror`?) and trigger the existing block of code for filling the buffer.

python/mujoco/rollout.cc
```diff
@@ -141,27 +179,38 @@ void _unsafe_rollout(std::vector<const mjModel*>& m, mjData* d, int start_roll,
...
-        for (; t < nstep; t++) {
-          size_t step = r*static_cast<size_t>(nstep) + t;
-          if (state) {
-            mj_getState(m[r], d, state + step*nstate, mjSTATE_FULLPHYSICS);
-          }
-          if (sensordata) {
-            mju_copy(sensordata + step*nsensordata, d->sensordata, nsensordata);
-          }
-        }
+        FillRemainingOutputs(
```
aftersomemath: I recommend keeping this code as is, and add a flag `nerror` to the if statement.

aftersomemath (commented):
#2704 makes a good point, that exceptions should trigger the same behavior as warnings. I think we can simplify this PR considerably, see detailed comments.

devshahofficial:
@aftersomemath @yuvaltassa thanks, this was a helpful pass. I pushed a cleanup commit and the rollout code is a lot closer to the shape suggested in the review now.

Main changes:
- errors now go through the same fill-the-rest path as warnings, using an `nerror` flag
- removed the extra fill helper/template wrapper and stopped catching `mj_setState`
- removed the outer `InterceptMjErrors` calls around the rollout wrappers
- tightened the regression test so the error happens mid-rollout instead of on the first step, with both main-thread and threaded coverage

I kept the public kwarg as `raise_on_error` for now because the default path still raises, while `False` opts into the fill behavior. Happy to rename it if you prefer something like `fill_on_error`.

Validated locally with:
- `python3 -m py_compile python/mujoco/rollout.py python/mujoco/rollout_test.py`
- built the Python sdist and wheel from this branch
- `MUJOCO_GL=disable /private/tmp/mujoco-rollout-fix-venv/bin/python -m pytest -q --pyargs mujoco.rollout_test` (`67 passed, 1 skipped`)

python/mujoco/rollout.cc
```diff
@@ -139,39 +144,50 @@ void _unsafe_rollout(std::vector<const mjModel*>& m, mjData* d, int start_roll,
...
       }
 
       // copy out sensor values
-      if (sensordata) {
+      if (sensordata && !nwarning && !nerror) {
         mju_copy(sensordata + step*nsensordata, d->sensordata, nsensordata);
       }
+
+      // if any warnings or handled errors, fill remaining outputs with current outputs, break
+      if (nwarning || nerror) {
```
aftersomemath: This block of code that fills the buffer should remain at the beginning of the function.

python/mujoco/rollout.cc
```diff
@@ -139,39 +144,50 @@ void _unsafe_rollout(std::vector<const mjModel*>& m, mjData* d, int start_roll,
...
+          if (raise_on_error) {
+            throw;
+          }
+          nerror = true;
+        }
+      }
 
       // copy out new state
-      if (state) {
+      if (state && !nwarning && !nerror) {
```
aftersomemath: These additional checks will not be needed when the fill code block is moved back to the beginning.

python/mujoco/rollout.cc
```diff
@@ -139,39 +144,50 @@ void _unsafe_rollout(std::vector<const mjModel*>& m, mjData* d, int start_roll,
...
       // step
-      mj_step(m[r], d);
+      if (!nwarning) {
+        try {
+          InterceptMjErrors(mj_step)(m[r], d);
+        } catch (const FatalError&) {
+          if (raise_on_error) {
+            throw;
+          }
+          nerror = true;
```
aftersomemath: You can use the `continue` keyword here so that the remaining code is skipped, and the fill code will execute.

python/mujoco/rollout.py
aftersomemath: The name is correct as is, my previous comment was incorrect.

python/mujoco/rollout.cc
```diff
@@ -352,10 +390,10 @@ PYBIND11_MODULE(_rollout, pymodule) {
...
         py::arg("sensordata") = py::none(),
         py::arg("chunk_size") = py::none(),
+        py::arg("raise_on_error") = true,
         py::doc(rollout_doc));
 }
 
 }  // namespace
 
 }  // namespace mujoco::python
-
```
aftersomemath: nit: whitespace change should be removed

python/mujoco/rollout.cc
```diff
@@ -311,14 +348,15 @@ class Rollout {
         } else {
           chunk_size_final = *chunk_size;
         }
-        InterceptMjErrors(_unsafe_rollout_threaded)(
+        _unsafe_rollout_threaded(
```
aftersomemath, line 375: On second thought, we should still use `InterceptMjErrors` here. In the original PR I thought the usage was `InterceptMjErrors` was newly introduced, but it was already there.

python/mujoco/rollout.cc
```diff
@@ -311,14 +348,15 @@ class Rollout {
...
             model_ptrs, data_ptrs, nbatch, nstep, control_spec, state0_ptr,
             warmstart0_ptr, control_ptr, state_ptr, sensordata_ptr,
-            this->pool_.get(), chunk_size_final);
+            this->pool_.get(), chunk_size_final, raise_on_error);
       } else {
-        InterceptMjErrors(_unsafe_rollout)(
+        _unsafe_rollout(
```
aftersomemath, line 380: Same here

python/mujoco/rollout.cc
```diff
@@ -181,36 +197,57 @@ void _unsafe_rollout_threaded(std::vector<const mjModel*>& m, std::vector<mjData
...
   pool->ResetCount();
 
   // schedule all jobs of full (chunk) size
   for (int j = 0; j < nfulljobs; j++) {
-    auto task = [=, &m, &d](void) {
+    auto task = [=, &m, &d, &rollout_error, &rollout_error_mutex](void) {
       int id = pool->WorkerId();
-      _unsafe_rollout(m, d[id], j*chunk_size, (j+1)*chunk_size,
-        nstep, control_spec, state0, warmstart0, control, state, sensordata);
+      try {
```
aftersomemath: I believe we do not need this new code if we go back to using `InterceptMjErrors` with `_unsafe_rollout_threaded`.

python/mujoco/rollout.cc
```diff
@@ -181,36 +197,57 @@ void _unsafe_rollout_threaded(std::vector<const mjModel*>& m, std::vector<mjData
...
 
   // schedule any remaining jobs of size < chunk_size
   if (chunk_remainder > 0) {
-    auto task = [=, &m, &d](void) {
-      _unsafe_rollout(m, d[pool->WorkerId()], nfulljobs*chunk_size,
-        nfulljobs*chunk_size+chunk_remainder,
-        nstep, control_spec, state0, warmstart0, control, state, sensordata);
+    auto task = [=, &m, &d, &rollout_error, &rollout_error_mutex](void) {
+      try {
+        _unsafe_rollout(m, d[pool->WorkerId()], nfulljobs*chunk_size,
```
aftersomemath, line 258: Same here.

aftersomemath (commented):
Thanks for making the changes! I have to request a few more to further simplify things, but this looks almost done.

devshahofficial:
@aftersomemath @yuvaltassa thanks for the review pass. I pushed `8f482ac4` to address the latest rollout threads:

- moved the fill-remaining-outputs block back to the beginning of the per-step loop
- use `continue` after a handled `mj_step` error so the existing fill path handles the rest of the trajectory
- removed the extra `nwarning` / `nerror` checks that became unnecessary once the fill block moved back up
- restored `InterceptMjErrors` around both rollout wrapper calls
- removed the extra threaded exception-forwarding code and unused headers
- restored the whitespace-only EOF change

I kept `raise_on_error` unchanged since the follow-up said the name is correct. Local validation: Python syntax checks and `git diff origin/main --check` for the rollout files. I could not run the compiled rollout tests in this environment because `cmake` is not installed here.

python/mujoco/rollout.cc
```diff
@@ -161,7 +165,22 @@ void _unsafe_rollout(std::vector<const mjModel*>& m, mjData* d, int start_roll,
...
+        }
+        // Write this step before continuing to the fill path.
+        if (state) {
+          mj_getState(m[r], d, state + step*nstate, mjSTATE_FULLPHYSICS);
+        }
+        if (sensordata) {
+          mju_copy(sensordata + step*nsensordata, d->sensordata, nsensordata);
+        }
+        nerror = true;
+        continue;
```
aftersomemath: Now that the code is simplified, it seems clear that the continue statement and the `mj_getState` and `mju_copy` calls can be removed.

aftersomemath (commented):
Looks good to me except for a minor comment.

With regards to the comment about not running the new changes, it is important to run and test changes yourself before submitting them for review. I see most of the CI jobs passed, except for one on Mac OS which is complaining about the rollout.cc file. Not sure if that is related.

python/mujoco/rollout.cc
devshahofficial: @aftersomemath thanks, agreed. I pushed `7dffd601` to remove the redundant handled-error copy-out and `continue`. After `nerror` is set, the existing copy-out path records the current state/sensordata and the next loop iteration fills the remaining outputs. Local validation: `python3 -m py_compile python/mujoco/rollout.py python/mujoco/rollout_test.py`, `git diff --check -- python/mujoco/rollout.cc python/mujoco/rollout.py python/mujoco/rollout_test.py doc/python.rst`, and the same whitespace check against `origin/main`.

devshahofficial:
Rebased this onto current `main` and fixed the EOF/pre-commit issue. The handled `mj_step` error path still uses the warning-style fill behavior, with the redundant copy-out removed. Local `git diff --check` and Python syntax checks pass; fresh CI has already cleared pre-commit and the Studio jobs, with the rest still running.

devshahofficial:
Rebased this onto current `main` again and pushed the refreshed branch.

I also reran the local validation from an installed package this time:
- `git diff --check`
- `python3 -m py_compile python/mujoco/rollout.py python/mujoco/rollout_test.py`
- generated the Python sdist with `python/make_sdist.sh`
- installed the generated `mujoco-3.10.1.tar.gz` into a fresh venv against a local MuJoCo install
- `MUJOCO_GL=disable python -m pytest -q --pyargs mujoco.rollout_test -k "intercept_mj_errors or mj_errors_can_fill_remaining_trajectory_from_mid_rollout"` (`3 passed, 65 deselected`)
- `MUJOCO_GL=disable python -m pytest -q --pyargs mujoco.rollout_test -k "single_step"` (`4 passed, 64 deselected`)

Fresh CI is running now.

devshahofficial:
Rebased onto current main and pushed a clean refresh. Fresh CI should be running now.

Local validation:
- `git diff --check`
- `python3 -m py_compile python/mujoco/rollout.py python/mujoco/rollout_test.py`
- built a fresh minimal MuJoCo install
- generated the Python sdist
- installed the sdist into a fresh venv
- `MUJOCO_GL=disable python -m pytest -q --pyargs mujoco.rollout_test -k "intercept_mj_errors or mj_errors_can_fill_remaining_trajectory_from_mid_rollout"` (`3 passed, 65 deselected`)
- `MUJOCO_GL=disable python -m pytest -q --pyargs mujoco.rollout_test -k "single_step"` (`4 passed, 64 deselected`)

@yuvaltassa @aftersomemath I think this is in a good spot again. Could you take one more look / submit if the fresh matrix stays green?

devshahofficial:
CI rerun was blocked on GitHub side (`workflow file may be broken`), so I refreshed the branch to kick a new matrix on the current-base head `a45fe525`. No source changes beyond the validated rebase; the focused rollout slice still passes locally (`3 passed, 65 deselected`). I will let this run unless the new checks surface something real.

aftersomemath:
Hi, sorry for the delay. I have been very busy. I will look it over again soon.

devshahofficial:
Quick CI follow-up: the previous run died on hosted-runner comms in `macos-15-x86_64`, and the core Linux/macOS jobs were cancelled later. GitHub also refused a failed-job rerun with the same `workflow file may be broken` error, so I rebased onto current `main` to kick a clean matrix at `6fac912a`.

No source changes beyond the rebase. Local validation on the rebased head:
- `git diff --check`
- `python3 -m py_compile python/mujoco/rollout.py python/mujoco/rollout_test.py`
- minimal MuJoCo configure/build/install
- Python sdist build + wheel install from that sdist
- `MUJOCO_GL=disable python -m pytest -q --pyargs mujoco.rollout_test -k "intercept_mj_errors or mj_errors_can_fill_remaining_trajectory_from_mid_rollout"` (`3 passed, 65 deselected`)

Fresh CI is running now.

devshahofficial:
CI follow-up: the fresh matrix at `6fac912a` ended with the long Python binding jobs cancelled around the 6h mark, and GitHub still refuses a failed-job rerun with `workflow file may be broken`.

I do not see a public assertion/build failure to patch from here; the focused rollout slice was already passing locally on this head. @yuvaltassa / @aftersomemath, could one of you kick/check this from the maintainer side, or point me at a public log if there is a real hang I should chase?

aftersomemath (approved):
Looks good to me. Thank you for running the tests locally. Since main has passing CI, I would rebase on main and force push again to try and get it to pass.

yuvaltassa (approved):

yuvaltassa (changes requested):
This fails under both ASAN and TSAN. There are two distinct issues here:

A. High `setjmp`/`longjmp` Interception Overhead (Timeout in `test_detect_divergence`)
Wrapping `mj_step` inside the trajectory loop causes `InterceptMjErrors` to be evaluated for every step:
```cpp
    for (size_t t = 0; t < nstep; t++) {
      try {
        InterceptMjErrors(mj_step)(m[r], d);
      } catch (const FatalError&) {
        // ...
      }
```
*   **The Issue:** `InterceptMjErrors` calls `setjmp` and configures thread-local error handlers on every iteration. For `test_detect_divergence` (which runs `nstep = 10000`, `nbatch = 4`), this executes `setjmp` 40,000 times.
*   **The Impact:** Sanitizers (ASAN/TSAN) intercept `setjmp`/`longjmp` to track active stack pointer changes and thread states. Calling this so frequently adds massive instrumentation overhead, slowing down the test run and causing it to time out.
*   **Suggested Fix:** Instead of invoking `InterceptMjErrors` on every single step, we should invoke it once for the entire rollout loop (or per Batch Element `r`). For example:
    ```cpp
    bool nerror = false;
    volatile int t = 0;
    try {
      InterceptMjErrors([&]() {
        for (; t < nstep; ++t) {
          // ...
          mj_step(m[r], d);
          // ...
        }
      })();
    } catch (const FatalError&) {
      if (raise_on_error) {
        throw;
      }
      nerror = true;
    }
    if (nerror) {
      // Fill the remaining steps of the trajectory loop starting from t...
    }
    ```
    *(Note: Declaring `t` as `volatile` ensures the outer loop progress tracking isn't optimized into registers that might get clobbered by standard `longjmp` stack unwinding).*

B. Uncaught Python Exceptions on Worker Threads (Thread Pool Deadlock)
If a Python callback (such as a control callback) raises a Python exception:
1.  `CallPyCallback` in `callbacks.cc` captures the exception and calls `e.restore()`.
2.  `WrapFunc` checks `PyErr_Occurred()` after `longjmp` and throws `pybind11::error_already_set`.
3.  However, `_unsafe_rollout` only catches `FatalError`:
    ```cpp
    try {
      InterceptMjErrors(mj_step)(m[r], d);
    } catch (const FatalError&) {
      // ... pybind11::error_already_set is NOT caught here!
    }
    ```
4.  This means the C++ `pybind11::error_already_set` exception propagates uncaught out of the worker thread task lambda.
5.  In `ThreadPool::WorkerThread`, there is no try-catch around `task()`. The uncaught exception terminates the worker thread immediately, skipping task counter increments and condition variable notifications (`++ctr_; cv_ext_.notify_one()`).
6.  The calling thread then waits indefinitely in `WaitCount` for completed tasks, causing a permanent deadlock/hang.

*   **Suggested Fix:** The thread pool task execution or `_unsafe_rollout` should ensure that any exception (not just `FatalError`, but any `std::exception`) is safely caught so that thread pool counters are consistently incremented under all conditions, preventing hangs when thread panics occur.

devshahofficial:
@yuvaltassa fixed both sanitizer issues in `8a21079d` and rebased onto current `main`.

- `InterceptMjErrors` now wraps each trajectory once, with a volatile progress counter for the `longjmp` path.
- threaded jobs capture/rethrow the first exception only after every task reaches the pool completion counter, so a callback failure no longer kills a worker or hangs `WaitCount`.
- I removed the outer nested interceptors too; on current `main`, nested TLS warning forwarding can loop back into the same handler.

Validation: fresh MuJoCo install + Python sdist/wheel build, the focused handled-error slice (`3 passed`), threaded callback-error/pool-reuse smoke, `test_detect_divergence` (`1 passed`), `git diff --check`, and pre-commit. Fresh CI is running now.

yuvaltassa:
Thanks, will TAL soon

PR #3249, opened 2026-04-29
Set the USD schema multiccd scene flag default to True

LouRohanNV:
`multiccd` was enabled by default for 3.8.0, but the USD schema was not updated. Ideally there should be a test to detect this mismatch.

### DIFF
--- a/src/experimental/usd/mjcPhysics/generatedSchema.usda
+++ b/src/experimental/usd/mjcPhysics/generatedSchema.usda
@@ -138,7 +138,7 @@ class "MjcSceneAPI" (
         displayName = "Mid-Phase Collision Filtering Toggle"
         doc = "Enables mid-phase collision filtering using a static AABB bounding volume hierarchy (BVH)."
     )
-    uniform bool mjc:flag:multiccd = 0 (
+    uniform bool mjc:flag:multiccd = 1 (
         displayName = "Multiple Contact Collision Detection (CCD) Toggle"
         doc = "Enables multiple-contact collision detection for geom pairs using a general-purpose convex-convex collider."
     )
@@ -727,4 +727,3 @@ class MjcTendon "MjcTendon" (
         doc = "Radius of the cross-section area of the spatial tendon, used for rendering in MuJoCo. Parts of the tendon that wrap around geom obstacles are rendered with reduced width."
     )
 }
-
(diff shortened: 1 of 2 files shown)


LouRohanNV:
I believe the proper test for this new change is to add these lines in `mjc_physics_scene_test.cc` around line 182:

```
  EXPECT_DISABLE_FLAG_USD_FALLBACK_EQ_MODEL_DEFAULT(IslandFlag, mjDSBL_ISLAND);
  EXPECT_DISABLE_FLAG_USD_FALLBACK_EQ_MODEL_DEFAULT(MultiCCDFlag,
                                                    mjDSBL_MULTICCD);
```

Also, I'm getting this from a review of the file:

The assertion line that was deleted in the "enable multiccd" commit was *already broken* on its own terms: the macro compared a `bool` to an `int` bitmask via `EXPECT_EQ`/`EXPECT_NE`, so `1 == (1<<19)` is `false` for high bit positions. With `mjENBL_MULTICCD = 1<<…`, the assertion would never have actually matched the model default — it had to be replaced with the disable-flag form (`mjDSBL_MULTICCD`) anyway, but with the bool coercion fix it would have been a real check.

I'm not 100% certain if this is correct, but it suggests that these macros be adjusted to:

```
#define EXPECT_DISABLE_FLAG_USD_FALLBACK_EQ_MODEL_DEFAULT(usd_attr, mjc_flag) \
  {                                                                          \
    bool flag;                                                               \
    mjc_phys_scene.Get##usd_attr##Attr().Get(&flag);                         \
    EXPECT_NE(flag,                                                          \
              static_cast<bool>(default_model->opt.disableflags & (mjc_flag))); \
  }

#define EXPECT_ENABLE_FLAG_USD_FALLBACK_EQ_MODEL_DEFAULT(usd_attr, mjc_flag) \
  {                                                                          \
    bool flag;                                                               \
    mjc_phys_scene.Get##usd_attr##Attr().Get(&flag);                         \
    EXPECT_EQ(flag,                                                          \
              static_cast<bool>(default_model->opt.enableflags & (mjc_flag))); \
  }
```

havess:
Closing, this was fixed upstream in https://github.com/google-deepmind/mujoco/commit/a501731089a7563abf49337d1049dd6269cbcf9d

Issue #2982, opened 2025-12-11
Frame issue in `mjCBody::AccumulateInertia` causing issues with `fusestatic`


dicarlo236:
### Intro

Hi!

I'm using mujoco to simulate robots. I ran into an issue with loading URDFs containing fixed joints that I think is a bug in adding together inertia.

### My setup

I'm using mujoco built from master on commit e07adc387a09d08d456a8e23336875216c0b5059 on Ubuntu Linux 22.04 x86_64. I've also observed the issue on release 3.3.7. I think the issue will happen on all platforms.

### What's happening? What did you expect?

I noticed the inertias are not added together correctly when `fusestatic` is enabled and the child body is rotated. Translation is handled as expected.

I believe the issue is introduced in this commit: https://github.com/google-deepmind/mujoco/commit/7932b4b202289a98d5bc9b8041d576a091ccddd3

In particular, line 2266 should use `other_iquat` (which includes the pose of the child body) instead of `other->iquat`. 
https://github.com/google-deepmind/mujoco/blob/e07adc387a09d08d456a8e23336875216c0b5059/src/user/user_objects.cc#L2244-L2267

### Steps for reproduction

1. Load the model below, which is a cross and has `fusestatic` set to false.
2. Load in `simulate`. Press `space` to pause, `backspace` to reset the model to the origin, then press `i` to visualize inertias. They look correct.
3. Modify the model to set `fusestatic` to `true`
4. Load in `simulate` and observe the wrong inertia

Here is the model with `fusestatic` set to false:

<img width="476" height="432" alt="Image" src="https://github.com/user-attachments/assets/5f8a24a4-68a3-4ac2-b88b-d0ca175464f1" />

Here is the model with fusestatic set to true. The inertia of the child body (blue) is added to the inertia of the parent body (red) incorrectly. The expected behavior is pizza-box-shaped inertia that goes around the cross.

<img width="592" height="382" alt="Image" src="https://github.com/user-attachments/assets/b8844ad4-ede4-421e-a6bf-02be0b3485cb" />

Next, change line 2266 in `user_objects.cc` to 
```cpp
    {other_iquat[0], other_iquat[1], other_iquat[2], other_iquat[3]}
```
rebuild mujoco, load the model, view inertia and see that the inertia looks correct:

<img width="445" height="450" alt="Image" src="https://github.com/user-attachments/assets/8222ea5b-bd5a-4fdb-b260-3187fe3963c5" />

### Minimal model for reproduction

[...]

PR #3250, opened 2026-04-29
Fix COM frame composition in mjCBody::AccumulateInertia

gholmes829:
`mjCBody::AccumulateInertia` calls `mjuu_frameaccum` with the body pose and inertial offset arguments swapped, composing `body_ipose ∘ body_pose` instead of the documented `body_pose ∘ body_ipose`. The merged center-of-mass is wrong whenever the body pose and inertial offset don't commute as transforms.

Most commonly triggered by URDF fixed-joint children with degenerate-eigenvalue inertia (typical of capsule/cylinder approximations), where `mjuu_fullInertia` returns a non-trivial principal-axes rotation.

Distinct from #2982, which patched the inertia-tensor rotation in the same function. #2982's reproduction (child with `pos = 0`) collapses both arg orders to the same answer, so it could not have caught this.

Added test fails before the fix with merged `body_ipos = (0, 0, 0.5)` and passes after with `(0.5, 0, 0)`.

### DIFF
--- a/test/user/user_objects_test.cc
+++ b/test/user/user_objects_test.cc
@@ -2666,6 +2666,38 @@ TEST_F(UserObjectsTest, Inertial) {
   EXPECT_THAT(error, HasSubstr("fullinertia and inertial orientation cannot"));
 }
 
+// Merged COM must be correct when a fused-static child has a non-identity
+// inertial-frame orientation (e.g. from a degenerate diaginertia).
+TEST_F(UserObjectsTest, FuseStaticWithRotatedInertial) {
+  static constexpr char xml[] = R"(
+  <mujoco>
+    <compiler fusestatic="true"/>
+    <worldbody>
+      <body name="parent">
+        <freejoint/>
+        <inertial pos="0 0 0" mass="1" diaginertia="1 1 1"/>
+        <body name="child" pos="1 0 0">
+          <inertial pos="0 0 0" quat="0.5 0.5 -0.5 0.5"
+                    mass="1" diaginertia="0.01 0.01 0.02"/>
+        </body>
+      </body>
+    </worldbody>
+  </mujoco>
+  )";
+
+  char error[1024];
+  mjModel* m = LoadModelFromString(xml, error, sizeof(error));
+  ASSERT_THAT(m, NotNull()) << error;
+  // Child fuses into parent; only world + parent remain.
+  EXPECT_EQ(m->nbody, 2);
+  EXPECT_EQ(m->body_mass[1], 2);
+  // Mass-weighted COM in parent frame: (1*(0,0,0) + 1*(1,0,0)) / 2.
+  const mjtNum expected_com[3] = {0.5, 0, 0};
+  EXPECT_THAT(AsVector(m->body_ipos + 3, 3),
+              Pointwise(MjNear(1e-8, 1e-6), AsVector(expected_com, 3)));
+  mj_deleteModel(m);
+}
+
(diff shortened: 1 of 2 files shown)


test/user/user_objects_test.cc
```diff
@@ -2676,6 +2676,38 @@ TEST_F(UserObjectsTest, Inertial) {
   EXPECT_THAT(error, HasSubstr("fullinertia and inertial orientation cannot"));
 }
 
+// Merged COM must be correct when a fused-static child has a non-identity
+// inertial-frame orientation (e.g. from a degenerate diaginertia).
+TEST_F(UserObjectsTest, FuseStaticWithRotatedInertial) {
```
quagla, line 2671: Hi I tried to run this test and it passes even before the bugfix, could we please modify the test such that it fails before the fix and it passes afterwards? Thanks!

quagla (changes requested):

test/user/user_objects_test.cc
gholmes829: Hi @quagla thanks for the feedback.

I've re-checked on a clean worktree of current `main` (2b67fe3c) with only the test cherry-picked in (no source change) -- test fails for me with `body_ipos = {0, 0, 0.5}` vs expected `{0.5, 0, 0}`. Cherry-picking the source change from this PR makes it pass.

Could the source change have still been applied when you tested? Or is there a different testing approach I should take?

For reference, I'm building and running with:
```bash
cmake -B build -DMUJOCO_BUILD_TESTS=ON -DCMAKE_BUILD_TYPE=Release .
cmake --build build --target user_objects_test -j$(nproc)
./build/bin/user_objects_test --gtest_filter="UserObjectsTest.FuseStaticWithRotatedInertial"
```

Platform:
- Linux x86_64 (WSL2), gcc default, Release mode, googletest from the repo's `_deps/`

quagla: Something probably went wrong the previous time I tested it. I confirm it fails for me too!

quagla (approved):

test/user/user_objects_test.cc
gholmes829: Awesome -- glad we're on same page, and thanks again for the timely review and approval.

At this point are we just waiting for review from @yuvaltassa? Just checking in to make sure y'all aren't waiting on me for something that I'm not realizing.

yuvaltassa (approved):

gholmes829:
Hi @yuvaltassa @quagla, thanks again for the reviews. I am noticing the `feedback/copybara` failed despite all the normal tests and public repo checks passing. Just want to check in -- is there anything you need from me to remedy these, or is this something that is being handled internally?

gholmes829:
Hi @yuvaltassa  -- just following up. `feedback/copybara` has been red and I can't see the internal details. Anything needed from my end or is it being handled internally? Happy to rebase or adjust if needed.

gholmes829:
Hi @saran-t, sorry to escalate. This PR was approved by @quagla and @yuvaltassa, but `feedback/copybara` has been failing since the 5/18 run despite all public checks passing. My attempts to follow up on 5/22 and 6/4 haven't moved things forward, and I don't have visibility into the internal details.

Is there someone on the import side who could help, or anything I should adjust from here? I'd be happy to rebase if useful. It's a small but meaningful correctness fix and would love to see it land!

yuvaltassa:
We still haven't merged this? 🙄

Apologies

yuvaltassa (approved):

Issue #3175, opened 2026-03-12
[Unity plugin] 2 tests stochastically fail in the plugin test suite.


Balint-H:
### Intro

During the maintenance of the Unity plugin I ran into strange behaviour in the test suite.

### My setup

This was observed on Windows 11, Unity 6000.3.10f1, and using #3164 version of the plugin.

### What's happening? What did you expect?

Two of the tests of the Unity plugin, namely `MjcfHierarchyMirrorsTheHierarchyOfUnityObjects` and `SceneRecreatedAfterDeletion` fail stochastically (i.e. rerunning the tests causes them to pass sometimes). 

It is unclear at the moment whether it is a limitation of the tests or if it is a bug, but warrants some attention.

The first suspicion is that it might be due to the ordering of components not being deterministic, but in an initial scan I could not find a step that is known to have non-defined orders.

### Steps for reproduction

Add the plugin from disk to a Unity project, then add `"testables": ["org.mujoco"]` to the project manifest. Afterwards, you can run the tests from Window>General>Test runner. Running the tests multiple times will cause the tests to sometimes fail or succeed, seemingly randomly.

### Confirmations

- [x] I searched the [latest documentation](https://mujoco.readthedocs.io/en/latest/overview.html) thoroughly before posting.
- [x] I searched previous [Issues](https://github.com/google-deepmind/mujoco/issues) and [Discussions](https://github.com/google-deepmind/mujoco/discussions), I am certain this has not been raised before.

njfletcher215:
Hello!

Your first suspicion is correct, the ordering of components is not deterministic.

In `unity/Runtime/Components/MjScene.cs:138-150`:
```
    var hierarchyRoots = FindObjectsByType<MjComponent>(FindObjectsSortMode.None)
        .Where(component => MjHierarchyTool.FindParentComponent(component) == null)
        .Select(component => component.transform)
        .Distinct();
    _orderedComponents = new List<MjComponent>();
    foreach (var root in hierarchyRoots) {
      _orderedComponents.AddRange(MjHierarchyTool.LinearizeHierarchyBFS(root));
    }

    XmlDocument sceneMjcf = null;
    try {
      _generationContext = new MjcfGenerationContext();
      sceneMjcf = GenerateSceneMjcf(_orderedComponents);
```
[...]

PR #3267, opened 2026-05-09
[Unity] Switched from arbitrary to ID-based ordering when creating the scene

njfletcher215:


njfletcher215:
Closes #3175

Balint-H:
Heya! Thanks for looking into the issue! Have you ran the tests 10+ times with your fix? Weird request, I know, but in this issue the fail only happens on some occasions

njfletcher215:
No problem! Yes, each test passed 20/20 iterations.

For reference, before the fix both tests were about 50/50:
- MjcfHierarchyMirrorsTheHierarchy passed 11/20 iterations
- SceneRecreatedAfterDeletion passed 9/20 iterations.

Balint-H (approved):
Thanks for double checking! Looks good to me. There is a very minor extra computation cost with the flag but it is a non-issue since CreateScene is called sparsely.

Balint-H:
LGTM, @erez-tom

SebasNorenaR (approved):

Issue #505, opened 2022-09-27
Regarding the new boxQP implementation


nrkumar93:
Glad to see the new [boxQP](https://github.com/deepmind/mujoco/blob/1e2a9a53bcca2e2bf6758b350db32346d132098d/include/mujoco/mujoco.h#L1036) implementation and looking forward to the next release. 

Just want to ask if it 
1. implements the Projected Newton Method as it is not explicitly stated anywhere yet.
2. is ready to use or is a work in progress?

Thanks

yuvaltassa:
1. As stated in the changelog, it implements the algorithm described in the appendix of [this paper](https://doi.org/10.1109/ICRA.2014.6907001). Yes, it is a projected Newton method.

2. It is ready to use, but there is a an upcoming small modification. Currently it assumes you are giving it a symmetric Hessian. If you give it an asymmetric one it will fail to converge. After the upcoming change it only reads from the lower triangle which will effectively symmetrise the Hessian, making the above issue moot.

yuvaltassa:
a6b338db6481ff6d94ea62a11eecd90962d8903d was jus pushed with the aforementioned improvement.

PR #3276, opened 2026-05-17
Honor lower-triangle contract for H in mju_boxQP.

### LINKED ISSUE #3275: mju_boxQP still reads the upper triangle of H in two matvec paths

mju_boxQP documents H as lower-triangle authoritative: “Only the lower triangles of H and R are read from and written to.” However, mju_boxQPoption still calls full dense matvec in two places:

- gradient update: H * res
- clamped/search update: H * temp

Both currently use mju_mulMatVec(..., H, ..., n, n), so the upper triangle of H is still read even though the Cholesky path already follows the lower-triangle contract.

The fix is small: add/use a symmetric-lower matvec helper and replace those two full matvec calls. This makes the whole boxQP path honor the documented lower-H contract, allowing callers to leave/poison the upper triangle safely.

Local microbench of this exact helper shape:
- n=16: about 10% slower than full dense matvec
- n>=32: about 6-20% faster
- symmetric quadratic form helper was substantially faster than dense matvec + dot

So this is mainly a correctness/contract fix, with likely performance upside for typical non-tiny QPs.

### Alternatives

_No response_

### Additional context

_No response_

dparikh79:
Fixes #3275.

## Summary

`mju_boxQP` documents that only the lower triangle of the Hessian `H` is read, but the two `mju_mulMatVec` calls inside `mju_boxQPoption` (`src/engine/engine_util_solve.c:1528` and `:1571`) walk the full dense matrix, so the upper triangle leaks into the gradient and search-direction updates. This breaks the documented contract and prevents callers from leaving the upper triangle uninitialized. Issue #505 and commit `a6b338db` (2022) established this lower-triangle contract for the Cholesky path; this completes it for the two remaining matvec call sites inside `mju_boxQPoption`.

## Change

Added a file-local `mulMatVecSym` helper at `src/engine/engine_util_solve.c:1407` that computes `res = H*vec` while reading only the lower triangle, mirroring the convention of the existing `mulVecMatVecSym` quadratic-form helper a few lines above. Both call sites in `mju_boxQPoption` now use it. Fast path on `H*x` shape unchanged.

## Validation

Built locally and ran the box QP test target. `UpperTrianglePoisoned` (new, in `test/engine/engine_util_solve_test.cc`) fills the strict upper triangle of `H` with NaN and asserts the solver returns the same result as on the clean symmetric input. Reverse-validated: the new test fails on the prior `mju_mulMatVec` code as expected, and all 21 cases in the box QP suite pass with the fix applied.

Reported by @lshdlut.

### DIFF
--- a/src/engine/engine_util_solve.c
+++ b/src/engine/engine_util_solve.c
@@ -1407,6 +1407,21 @@ static mjtNum mulVecMatVecSym(const mjtNum* vec, const mjtNum* mat, int n) {
 }
 
 
+// multiply symmetric matrix with vector: res = mat*vec
+//   assumes symmetry of mat, ignores upper triangle
+//   res must not alias vec
+static void mulMatVecSym(mjtNum* res, const mjtNum* mat, const mjtNum* vec, int n) {
+  for (int i=0; i < n; i++) {
+    // diagonal + strict lower triangle: res[i] = sum_{j<=i} mat[i,j] * vec[j]
+    res[i] = mat[n*i+i] * vec[i] + mju_dot(mat+n*i, vec, i);
+    // strict upper mirror contribution: res[k] += mat[i,k] * vec[i] for k < i
+    for (int k=0; k < i; k++) {
+      res[k] += mat[n*i+k] * vec[i];
+    }
+  }
+}
+
+
 // minimize 0.5*x'*H*x + x'*g  s.t. lower <= x <=upper, explicit options
 //  additional arguments to mju_boxQP (see mju_boxQP documentation):
 //   maxiter     maximum number of iterations
@@ -1512,7 +1527,7 @@ int mju_boxQPoption(mjtNum* res, mjtNum* R, int* index,               // outputs
     oldvalue = value;
 
     // compute gradient
-    mju_mulMatVec(grad, H, res, n, n);
+    mulMatVecSym(grad, H, res, n);
     mju_addTo(grad, g, n);
 
     // find clamped dimensions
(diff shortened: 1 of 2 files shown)


src/engine/engine_util_solve.c
```diff
@@ -1407,6 +1407,21 @@ static mjtNum mulVecMatVecSym(const mjtNum* vec, const mjtNum* mat, int n) {
 }
 
 
+// multiply symmetric matrix with vector: res = mat*vec
+//   assumes symmetry of mat, ignores upper triangle
+//   res must not alias vec
+static void mulMatVecSym(mjtNum* res, const mjtNum* mat, const mjtNum* vec, int n) {
```
yuvaltassa, line 1413: 1. Let's say `restrict res`
2. Let's rename the function `mulSimVec`, note that `mulSymVecSparse` already exists so this naming would conform better.

yuvaltassa (approved):
I started commenting but you can ignore.

I'll approve and make the edits myself.

yuvaltassa:
Please note the changes I made to your original code, maybe take that into account in your next PR :slightly_smiling_face: 

Thanks for your contribution!

dparikh79:
Thanks! I went through the diff. Didn't think of tracing the AVX path in `engine_util_blas.c`, so I missed the `mju_addToScl` swap. Noting the naming convention as well.

Quick question for the future: Is there a canonical way to spot which mujoco utilities have SIMD specializations beyond grepping for `mjUSEAVX`?

PR #3278, opened 2026-05-20
Expose MJX light_active model field

tkelestemur:
This exposes `light_active` as a public `mjx.Model` field, matching `mujoco.MjModel.light_active` and the nearby MJX light fields like `light_pos` and `light_castshadow`.

The current WARP path already carries this value, but only under `mjx_model._impl.light_active`. That means code that wants to toggle or randomize lights has to reach into private implementation state, while similar light fields can be updated with normal public model operations. With this change, users can do `model.replace(light_active=...)` and include `light_active` directly in model pytree batching / `in_axes` setup.

The actual source-of-truth change is adding `light_active` to public `mjx.Model`. After regenerating the WARP files, `ModelWarp` no longer keeps a duplicate WARP-only copy of the field, and the render shim now reads `m.light_active` instead of `m._impl.light_active`.

I added focused tests to check that `put_model` copies `MjModel.light_active`, that `replace(light_active=...)` works, and that `light_active` participates in a batched model pytree.

Validated with:

`python -m pytest mjx/mujoco/mjx/_src/io_test.py -k 'light_active' -q`

`python -m pytest mjx/mujoco/mjx/_src/io_test.py::ModelIOTest -q`

`git diff --check`


mjx/mujoco/mjx/warp/render.py
```diff
@@ -14,17 +14,19 @@
 # ==============================================================================
 
 """DO NOT EDIT. This file is auto-generated."""
```
btaba, line 16: OOC were you able to run the [codegen](https://github.com/google-deepmind/mujoco/tree/main/mjx/mujoco/mjx/codegen) here to make this change?

mjx/mujoco/mjx/_src/io_test.py
```diff
@@ -203,6 +212,49 @@ def assert_not_weak_type(x):
     np.testing.assert_equal(mx.wrap_objid, m.wrap_objid)
     np.testing.assert_almost_equal(mx.wrap_prm, m.wrap_prm)
 
+  @parameterized.parameters('jax', 'warp', 'cpp')
```
btaba, line 215: I don't think these tests are really necessary. If you want to make doubly sure someone doesn't change the public API without consulting with the git blame on who added this, you can keep `test_put_model_light_active`

btaba (changes requested):
LGTM modulo one comment

mjx/mujoco/mjx/warp/render.py
tkelestemur: Yes, I reran the narrow WARP codegen for `warp/types.py` and `warp/render.py`. Adding `light_active` to public `mjx.Model` makes the generator drop it from `ModelWarp`, and the render shim is generated to read `m.light_active`.

mjx/mujoco/mjx/_src/io_test.py
tkelestemur: done in https://github.com/google-deepmind/mujoco/pull/3278/commits/cf1b28d19a8428bed44633fee1b75c090f4292a7

btaba (approved):

tkelestemur:
@btaba fixed conflict and should be ready to go.

btaba (approved):

PR #3283, opened 2026-05-22
Fix set_to_muscle() to accept tuple arguments for timeconst and range

### LINKED ISSUE #3282: `set_to_muscle()` Uses wrong arg types
### Intro

Hi!

I am using the python API for musculoskeletal modelling.

### My setup

Working on WSL and Windows, Mujoco 3.8.

`set_to_muscle()` of actuators does not take 2d `range` and `timeconst` arguements, and instead expects scalar floats. The function works if you only put in one value of course, but then you manually have to edit dynprm and gainprm to ave correctly set up muscles. I'd also expect the defaults to work/be applied if we don't explicitly specify the values. There are also no tests for `set_to_muscle()` yet.

### Steps for reproduction

Calling set_to_muscle() on any actuator fails.

``` python
import mujoco
spec = mujoco.MjSpec()
a = spec.add_actuator()
a.set_to_muscle(timeconst=(0.01, 0.04), tausmooth=0., range=(0.75, 1.05),
                        force=-1., scale=200.,  lmin=0.5, lmax=1.6, vmax=1.5, fpmax=1.3, fvmax=1.2)
```

This works, but sets up the component incorrectly:

``` python
import mujoco
spec = mujoco.MjSpec()
a = spec.add_actuator()
a.set_to_muscle(timeconst=0.01, tausmooth=0., range=0.75,
                        force=-1., scale=200.,  lmin=0.5, lmax=1.6, vmax=1.5, fpmax=1.3, fvmax=1.2)
```

### Confirmations

anishesg:
The `set_to_muscle()` method in the Python bindings incorrectly declared `timeconst` and `range` parameters as C-style array pointers (`double timeconst[2]`), which prevented Python callers from passing tuples or lists as intended for musculoskeletal modeling. The underlying C function expects 2-element arrays representing min/max pairs, but the pybind11 binding only accepted scalar values.

Changed the lambda signature in `python/mujoco/specs.cc` to use `std::array<double, 2>` for both `timeconst` and `range` parameters, matching the pattern used in `set_to_dcmotor()`. Updated the default argument for `timeconst` from a scalar `-1` to `std::array<double, 2>{-1, -1}` for consistency. The arrays are correctly unpacked using `.data()` when calling the underlying `mjs_setToMuscle()` C function.

Added a test in `specs_test.py` that verifies tuple arguments are properly mapped to the `dynprm` and `gainprm` arrays, and that muscle actuator types are set correctly.

Fixes #3282

### DIFF
--- a/python/mujoco/specs_test.py
+++ b/python/mujoco/specs_test.py
@@ -1636,6 +1636,17 @@ class SpecsTest(absltest.TestCase):
     self.assertEqual(actuator.gainprm[7], 1.3)
     self.assertEqual(actuator.gainprm[8], 1.2)
 
+    # Verify biasprm matches gainprm (copied by set_to_muscle)
+    self.assertEqual(actuator.biasprm[0], 0.75)
+    self.assertEqual(actuator.biasprm[1], 1.05)
+    self.assertEqual(actuator.biasprm[2], -1.)
+    self.assertEqual(actuator.biasprm[3], 200.)
+    self.assertEqual(actuator.biasprm[4], 0.5)
+    self.assertEqual(actuator.biasprm[5], 1.6)
+    self.assertEqual(actuator.biasprm[6], 1.5)
+    self.assertEqual(actuator.biasprm[7], 1.3)
+    self.assertEqual(actuator.biasprm[8], 1.2)
+
     # Verify actuator type settings
     self.assertEqual(actuator.dyntype, mujoco.mjtDyn.mjDYN_MUSCLE)
     self.assertEqual(actuator.gaintype, mujoco.mjtGain.mjGAIN_MUSCLE)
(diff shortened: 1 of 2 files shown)


python/mujoco/specs_test.py
```diff
@@ -1602,6 +1602,45 @@ def test_actuator_shortname(self):
...
+    # Verify gainprm contains range and muscle parameters
+    self.assertEqual(actuator.gainprm[0], 0.75)
+    self.assertEqual(actuator.gainprm[1], 1.05)
+    self.assertEqual(actuator.gainprm[2], -1.)
+    self.assertEqual(actuator.gainprm[3], 200.)
+    self.assertEqual(actuator.gainprm[4], 0.5)
+    self.assertEqual(actuator.gainprm[5], 1.6)
+    self.assertEqual(actuator.gainprm[6], 1.5)
+    self.assertEqual(actuator.gainprm[7], 1.3)
+    self.assertEqual(actuator.gainprm[8], 1.2)
```
Balint-H, line 1637: Should probably also test biasprm while we are at it

anishesg: good catch, added

anishesg:
going to close - saw the issue got handled in another PR.

Issue #3284, opened 2026-05-22
Mujoco sysid sensitivity to min bound on frictionloss, damping and armature


proyan:
### Intro

Hi!

I am an engineer at Agile Robots, and I've been testing the Mujoco sysid recently.
Context: The solver is very sensitive to bounds at 0 (comparison is made for damping/armature/frictionloss min_values of 0 and 1e-8)
TLDR: 
1) The sysid solver convergence worsens because of multithreaded openblas (with an interesting exception). 
2) In addition, there is some other issue with mujoco that I've not been able to investigate.

### My setup

- MuJoCo: `3.8.1`
- Base image: `ubuntu:24.04`
- API: `Python`

### What's happening? What did you expect?

mujoco_sysid optimizer convergence is sensitive to OpenBLAS thread count with near-zero dynamics bounds

## Summary
I have a self-contained Docker environment. `mujoco_sysid` optimizer convergence changes substantially depending on OpenBLAS thread count when frictionloss armature or damping parameter has both `nominal` and `min_value` set to a near-zero positive value (`1e-8`).

## Environment

- Repository target: `mujoco_sysid`
- MuJoCo: `3.8.1`
- Docker image: `ubuntu:24.04`, forced to `linux/amd64`
- Python package versions in the image: `mujoco==3.8.1`, `numpy==2.4.6`, `scipy==1.17.1`
- Default comparison run: `OPENBLAS_NUM_THREADS` unset in the Docker image and unset at `docker run` time
- Single-thread comparison run: `OPENBLAS_NUM_THREADS=1` set only at `docker run` time

## Test case

The reproducer optimizes inertial and joint dynamics parameters from 10 local trajectories. For each run, exactly one field (armature, damping or frictionloss) has `nominal=min_value=1e-8`; the other dynamics fields use `nominal=min_value=0.0`.

Each run uses `max_iters=50`. The logs below report the optimizer's final residual for different runs

## Observed behavior

Lower final residual is better.

| min_value changed | OpenBLAS setting | Final residual `y` |
| --- | --- | ---: |
| `armature=1e-8` | unset | 6.282 |
| `armature=1e-8` | `OPENBLAS_NUM_THREADS=1` | 2.307 |
| `damping=1e-8` | unset | 970.5 |
| `damping=1e-8` | `OPENBLAS_NUM_THREADS=1` | 5.652 |
| `frictionloss=1e-8` | unset | 5.019 |
[...]

PR #3287, opened 2026-05-26
Improve sysid robustness, prompted by #3284

kevinzakka:
It adds 3 robustness aids to mujoco.sysid to hopefully alleviate issues like #3284. Specifically, it adds:

- `ParameterDict.move_off_bounds()`: this function nudges any parameter starting at a box bound into the interior. Additionally, `optimize` will now warn you when it detects that a parameter is at a bound.
- `x_scale` is now accepted as a kwarg in the mujoco solver. It rescales parameters with widely different magnitudes, a common source of ill-conditioning in sysid. The scipy backend already supported this.
- `check_conditioning=True` is now an arg in `optimize` and it will warn you when the condition number of the Hessian exceeds 1e12, a heuristic we chose well below the float64 conditioning limit.

### DIFF
--- a/python/mujoco/sysid/_src/optimize.py
+++ b/python/mujoco/sysid/_src/optimize.py
@@ -179,6 +179,24 @@ def optimize(
         extras={},
     )
 
+  # Warn if any non-frozen parameter component starts at (or essentially at)
+  # a box bound. Optimization can stall in that corner on ill-conditioned or
+  # rank-deficient problems; both the mujoco and scipy backends are affected.
+  lo, hi = bounds
+  rng = hi - lo
+  safe_rng = np.where(rng > 0, rng, 1.0)
+  at_bound = ((x0 - lo) <= 1e-3 * safe_rng) | ((hi - x0) <= 1e-3 * safe_rng)
+  at_bound &= rng > 0
+  if at_bound.any():
+    logging.warning(
+        "%d of %d non-frozen parameter components start at (or essentially "
+        "at) a box bound. Optimization can stall in that corner on "
+        "ill-conditioned problems; consider calling "
+        "initial_params.move_off_bounds() before optimize().",
+        int(at_bound.sum()),
+        x0.size,
+    )
+
   def optimized_residual_fn(x):
     residuals, _, _ = residual_fn(x, opt_params)
     return np.concatenate(residuals)
--- a/python/mujoco/sysid/_src/optimize.py
+++ b/python/mujoco/sysid/_src/optimize.py
@@ -26,18 +26,17 @@ import scipy.optimize as scipy_optimize
 import scipy.special
 
 
-XScale = Literal["jac"] | np.ndarray | float
-
-
 def _warn_if_ill_conditioned(
     initial_params: parameter.ParameterDict,
     residual_fn: Callable[..., Any],
     threshold: float = 1e12,
+    eps: float | None = None,
 ) -> None:
-  """Warn if cond(JᵀJ) at the starting point exceeds ``threshold``.
+  """Warn if cond(J^T J) at the starting point exceeds ``threshold``.
 
   Costs one extra finite-difference Jacobian. ``threshold=1e12`` corresponds
-  to cond(J) ~ 1e6, well below the float64 limit (~1e16).
+  to cond(J) ~ 1e6, well below the float64 limit (~1e16). ``eps`` defaults to
+  the finite-difference step used by the backends.
   """
   x0 = initial_params.as_vector()
   bounds = initial_params.get_bounds()
@@ -46,7 +45,8 @@ def _warn_if_ill_conditioned(
     residuals, _, _ = residual_fn(x, initial_params)
     return np.concatenate(residuals)
 
-  eps = np.finfo(np.float64).eps ** 0.5
+  if eps is None:
+    eps = np.finfo(np.float64).eps ** 0.5
   r0 = f(x0).reshape(-1, 1)
   jac = np.asarray(
       mujoco_minimize.jacobian_fd(
@@ -59,13 +59,10 @@ def _warn_if_ill_conditioned(
       )[0],
       dtype=np.float64,
   )
-  # eigvalsh on the gram matrix so rank-deficient directions are visible
-  # when n_params > n_residual components (SVD would drop to min(m, n)).
-  ev = np.maximum(np.linalg.eigvalsh(jac.T @ jac), 0.0)
-  cond_jtj = float(ev[-1] / ev[0]) if ev[0] > 0 else float("inf")
+  cond_jtj = float(np.linalg.cond(jac)) ** 2
   if cond_jtj > threshold:
     logging.warning(
-        "cond(JᵀJ) ≈ %.1e at the starting point; the problem may be "
+        "cond(J^T J) ~ %.1e at the starting point; the problem may be "
         "ill-conditioned. Consider x_scale='jac' or regularizing.",
         cond_jtj,
     )
@@ -84,7 +81,6 @@ def _scipy_least_squares(
     verbose = 2
   else:
     verbose = 0
-  x_scale = kwargs.pop("x_scale", "jac")
   loss = kwargs.pop("loss", "linear")
 
   jac_arg: str | Callable[..., Any]
@@ -117,7 +113,6 @@ def _scipy_least_squares(
       bounds=bounds,
       max_nfev=max_nfev,
       verbose=verbose,
-      x_scale=x_scale,
       loss=loss,
       jac=jac_arg,  # pyright: ignore[reportArgumentType]
       **kwargs,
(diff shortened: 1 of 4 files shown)


yuvaltassa:
I agree with @aftersomemath , the residual scaling is definitely a `mujoco.minimize` feature, and a good one too.
Do you mind splitting this out into a pre-PR?

python/mujoco/sysid/_src/parameter.py
```diff
@@ -139,6 +139,19 @@ def sample(self, rng: np.random.Generator | None = None) -> np.ndarray:
...
     return rng.uniform(self.min_value.flatten(), self.max_value.flatten())
 
+  def move_off_bound(self, fraction: float = 0.05) -> None:
+    """Shift values within 0.1% of a bound to ``lo + fraction*(hi-lo)`` (or
+    symmetric for the upper bound). Interior components are unchanged."""
+    lo, hi = self.get_bounds()
+    rng = hi - lo
+    safe_rng = np.where(rng > 0, rng, 1.0)
+    v = self.as_vector().copy()
+    at_lo = (v - lo) <= 1e-3 * safe_rng
```
aftersomemath: `1e-3` should probably be a parameter as well.

python/mujoco/sysid/tests/test_integration.py
```diff
@@ -233,3 +235,41 @@ def test_arm_recover_armature():
     assert (result_dir / "results.pkl").exists()
     assert (result_dir / "confidence.pkl").exists()
     assert (result_dir / "arm.xml").exists()
+
+
+def _rank_1_residual_fn(x, p):
```
aftersomemath, line 240: If I understand correctly, this residual function has a rank 1 Jacobian, but only because its output is scalar. A better test would be a function with 2 outputs, that depends on only one variable. Then the jacobian will be 2x2 and rank 1.

python/mujoco/sysid/_src/optimize.py
```diff
@@ -83,6 +128,7 @@ def _mujoco_least_squares(
     x0: np.ndarray,
     residual_fn: Callable[..., Any],
     bounds: tuple[np.ndarray, np.ndarray],
+    x_scale: XScale = 1.0,
```
aftersomemath: Not needed if we are relying on `**kwargs` as the doc string suggests.

python/mujoco/sysid/_src/optimize.py
```diff
@@ -155,8 +203,21 @@ def optimize(
     optimizer: Backend — ``"mujoco"`` (default), ``"scipy"``, or
       ``"scipy_parallel_fd"`` (scipy with MuJoCo finite-difference Jacobian).
     verbose: If True, log parameter comparison table after optimization.
-    **optimizer_kwargs: Forwarded to the backend (e.g. ``max_iters``,
-      ``verbose``, ``loss``).
+    check_conditioning: If True, estimate ``cond(JᵀJ)`` at the starting
```
aftersomemath: Is it ok to use unicode like supscript T?

python/mujoco/sysid/_src/optimize.py
```diff
@@ -179,6 +240,27 @@ def optimize(
...
+  if check_conditioning:
+    _warn_if_ill_conditioned(initial_params, residual_fn)
+
+  # Warn if any non-frozen parameter component starts at (or essentially at)
+  # a box bound. Optimization can stall in that corner on ill-conditioned or
+  # rank-deficient problems; both the mujoco and scipy backends are affected.
+  lo, hi = bounds
+  rng = hi - lo
+  safe_rng = np.where(rng > 0, rng, 1.0)
+  at_bound = ((x0 - lo) <= 1e-3 * safe_rng) | ((hi - x0) <= 1e-3 * safe_rng)
```
aftersomemath: We should probably check that x0 is within the bounds before executing this check

python/mujoco/sysid/_src/optimize.py
```diff
@@ -155,8 +203,21 @@ def optimize(
...
+      Costs one extra finite-difference Jacobian.
+    **optimizer_kwargs: Forwarded to the backend. Common ones:
+
+      * ``max_iters``: maximum number of optimizer iterations.
+      * ``verbose``: per-backend verbosity flag (separate from this
+        function's ``verbose``).
+      * ``loss``: scipy loss function name (scipy backends only).
+      * ``x_scale``: per-parameter scaling. ``"jac"`` is adaptive
+        ``D_i = 1/||J(:,i)||`` per iteration; an explicit array or
+        positive scalar is used as ``D`` directly. Defaults: ``"jac"``
```
aftersomemath: Scipy does not simply default to "jac": https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.least_squares.html#id2

python/mujoco/sysid/tests/test_integration.py
```diff
@@ -233,3 +235,41 @@ def test_arm_recover_armature():
...
+  params.add(sysid.Parameter("a", 1.0, -10.0, 10.0))
+  params.add(sysid.Parameter("b", 1.0, -10.0, 10.0))
+
+  with caplog.at_level(logging.WARNING, logger="absl"):
+    sysid.optimize(
+        initial_params=params, residual_fn=residual_fn,
+        optimizer="scipy", verbose=False, check_conditioning=True,
+        max_iters=1,
+    )
+  fired = any("cond(JᵀJ)" in r.message for r in caplog.records)
```
aftersomemath: unicode?

python/mujoco/sysid/_src/optimize.py
```diff
@@ -26,6 +26,51 @@
...
 
+XScale = Literal["jac"] | np.ndarray | float
+
+
+def _warn_if_ill_conditioned(
+    initial_params: parameter.ParameterDict,
+    residual_fn: Callable[..., Any],
+    threshold: float = 1e12,
+) -> None:
+  """Warn if cond(JᵀJ) at the starting point exceeds ``threshold``.
```
aftersomemath: unicode?

python/mujoco/sysid/_src/optimize.py
```diff
@@ -26,6 +26,51 @@
...
+  to cond(J) ~ 1e6, well below the float64 limit (~1e16).
+  """
+  x0 = initial_params.as_vector()
+  bounds = initial_params.get_bounds()
+
+  def f(x):
+    residuals, _, _ = residual_fn(x, initial_params)
+    return np.concatenate(residuals)
+
+  eps = np.finfo(np.float64).eps ** 0.5
```
aftersomemath: eps should be the same as what will be passed to mujoco/scipy. This is their default, but the user could override.

python/mujoco/sysid/_src/optimize.py
```diff
@@ -26,6 +26,51 @@
...
+          r=r0,
+          eps=eps,
+          n_res=0,
+          bounds=[bounds[0].reshape(-1, 1), bounds[1].reshape(-1, 1)],
+      )[0],
+      dtype=np.float64,
+  )
+  # eigvalsh on the gram matrix so rank-deficient directions are visible
+  # when n_params > n_residual components (SVD would drop to min(m, n)).
+  ev = np.maximum(np.linalg.eigvalsh(jac.T @ jac), 0.0)
```
aftersomemath: Is this better than using Numpy's condition number function? https://numpy.org/doc/stable/reference/generated/numpy.linalg.cond.html

aftersomemath (commented):
Minor comments.

python/mujoco/sysid/_src/parameter.py
kevinzakka: Done.

python/mujoco/sysid/tests/test_integration.py
kevinzakka: Done.

python/mujoco/sysid/_src/optimize.py
kevinzakka: Done.

python/mujoco/sysid/_src/optimize.py
kevinzakka: Done.

python/mujoco/sysid/_src/optimize.py
kevinzakka: Done.

python/mujoco/sysid/_src/optimize.py
kevinzakka: Done.

python/mujoco/sysid/tests/test_integration.py
kevinzakka: Done.

python/mujoco/sysid/_src/optimize.py
kevinzakka: Done.

python/mujoco/sysid/_src/optimize.py
kevinzakka: Done.

python/mujoco/sysid/_src/optimize.py
kevinzakka: Done.

aftersomemath:
LGTM. Just need to answer the outstanding question about mu adaptation in #3294.

yuvaltassa (approved):

PR #3293, opened 2026-05-26
MuJoCo Live: pre-fetch model assets in parallel before compilation.

kevinzakka:
The compiler can't move on to the next mesh until the previous fetch resolves, so a Menagerie scene takes 60-100 serial round trips before the model finishes loading.

The first commit changes `mju_getXMLDependencies` to read its XML through `mju_openResource` instead of tinyxml2's `LoadFile`. The function now works against any registered resource backend (VFS, HTTP, `github:`, ...). That's the prerequisite for calling it from JavaScript on a URL-backed model.

The second commit adds a JS-side prefetcher in `live.js`. It asks `Module.getXMLDependencies` for the transitive list of asset URLs, fetches them with `Promise.all`, and primes each into the WASM-side `FetchCache` through a new `Module.primeFetchCache` binding. The compiler subsequently finds every resource it asks for already in memory.

`FetchCache` is also refactored to store `std::string` directly, dropping the `UniquePtrWasm<char[]>` aliases and the manual `malloc`/`free` bookkeeping.

## Results

Measured on `anybotics_anymal_c/scene.xml` (89 fetches, ~71 MB) via headless Chromium:

| scenario           | before  | after  | speedup |
|--------------------|--------:|-------:|--------:|
| cold, no throttle  |  9.19 s | 2.08 s |   4.4 x |
| cold, slow-4g      | 19.80 s | 2.89 s |   6.9 x |
| warm, no throttle  |  1.27 s | 1.24 s | (WASM compute floor) |
| warm, slow-4g      |  1.31 s | 1.27 s | (WASM compute floor) |

Warm wall is unchanged because we've hit the WASM compute floor: XML parse, mesh decode, GPU upload, first render.

Verified end-to-end on `anybotics_anymal_c`, `unitree_g1`, and `shadow_hand`. All compile cleanly with the expected page title.

### DIFF
--- a/src/experimental/studio/emscripten.cc
+++ b/src/experimental/studio/emscripten.cc
@@ -27,6 +27,7 @@
 #include <string_view>
 #include <unordered_map>
 #include <utility>
+#include <vector>
 
 #include <mujoco/mujoco.h>
 #include "experimental/platform/hal/graphics_mode.h"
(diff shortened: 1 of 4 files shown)


saran-t (approved):

saran-t:
Are you unconditionally doing the prefetch without checking first whether the asset is http or https? For the drag-and-drop use case does this cause an error?

PR #3294, opened 2026-05-27
Add x_scale to mujoco.minimize.least_squares

kevinzakka:
Adds an `x_scale` parameter to `mujoco.minimize.least_squares`, matching scipy's `least_squares(method='trf', x_scale=...)`:

- `1.0` (default): no scaling, identical to current behavior.
- `'jac'`: adaptive, `D_i = 1/||J(:,i)||` recomputed each iteration.
- array of shape `(n,)` or positive scalar: used as `D` directly.

### DIFF
--- a/python/mujoco/minimize.py
+++ b/python/mujoco/minimize.py
@@ -156,6 +156,7 @@ def least_squares(
     output: Optional[TextIO] = None,
     iter_callback: Optional[Callable[[List[IterLog]], None]] = None,
     check_derivatives: bool = False,
+    x_scale: Union[float, np.ndarray, str] = 1.0,
 ) -> Tuple[np.ndarray, List[IterLog]]:
   """Nonlinear Least Squares minimization with box bounds.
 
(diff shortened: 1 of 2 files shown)


python/mujoco/minimize.py
```diff
@@ -156,6 +156,7 @@ def least_squares(
     output: Optional[TextIO] = None,
     iter_callback: Optional[Callable[[List[IterLog]], None]] = None,
     check_derivatives: bool = False,
+    x_scale: Union[float, np.ndarray, str] = 1.0,
```
aftersomemath: Can the default be None to match scipy?
https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.least_squares.html

python/mujoco/minimize.py
```diff
@@ -178,6 +179,13 @@ def least_squares(
     output: Optional file or StringIO to which to print messages.
     iter_callback: Optional iteration callback, takes trace argument.
     check_derivatives: Compare user-defined Jacobian and norm against fin-diff.
+    x_scale: Per-parameter scaling. Setting ``x_scale=D`` solves the problem in
+      the change of variables ``z = x / D`` and un-scales the result. ``1.0``
+      (default) is no scaling. ``'jac'`` sets ``D_i = 1 / ||J(:, i)||`` at each
+      iteration (matches scipy's ``least_squares(method='trf', x_scale='jac')``).
```
aftersomemath: Scipy supports scaling with methods other than trf, and the method closest to ours is LM, so maybe the comment should simply be "(matches scipy's x_scale parameter)"

python/mujoco/minimize.py
```diff
@@ -203,6 +211,22 @@ def least_squares(
...
+  # Resolve x_scale -> D of shape (n, 1). For 'jac', D is refreshed each iter.
+  adaptive_scale = isinstance(x_scale, str)
+  if adaptive_scale and x_scale != 'jac':
+    raise ValueError(
+        f"x_scale must be 'jac', a positive scalar, or array, got {x_scale!r}."
+    )
+  if adaptive_scale:
+    D = np.ones((n, 1))
+  else:
+    D = np.asarray(x_scale, dtype=np.float64)
```
aftersomemath, line 224: If the default argument is set to None to match scipy, then this logic will need to change.

aftersomemath (commented):
Looks good

python/mujoco/minimize.py
kevinzakka: Done.

python/mujoco/minimize.py
kevinzakka: Done.

PR #3301, opened 2026-05-30
Use box midpoint to choose finite-difference direction in jacobian_fd

Nas01010101:
## Problem

In `minimize.py`, `jacobian_fd` chooses each coordinate's finite-difference direction so the perturbation steps *away* from the nearer bound. It compared `x` against

```python
mid = 0.5 * (bounds[1] - bounds[0])
eps_vec = np.where(x > mid, -eps, eps)
```

but `0.5 * (bounds[1] - bounds[0])` is half the box **width**, not the box midpoint. For bounds that are not centered on the origin this picks the wrong direction. Concretely, with `bounds = [10, 20]` the comparison value is `5.0` (true midpoint is `15.0`), so a point at the lower bound `x = 10` satisfies `x > 5` and steps **−eps**, i.e. to `9.9999998` — outside the box.

## Fix

Compare against the actual midpoint:

```python
mid = 0.5 * (bounds[0] + bounds[1])
```

At the lower bound the step is now `+eps` (inward); at the upper bound `−eps` (inward).

## Tests

Adds `test_jacobian_fd_respects_off_center_bounds`, which records every point `jacobian_fd` evaluates with `x` at the lower bound of the off-center box `[10, 20]` and asserts they all stay within `[lo, hi]`. It fails before this change (evaluates `9.9999998`) and passes after; the rest of `minimize_test.py` remains green.

### DIFF
--- a/python/mujoco/minimize.py
+++ b/python/mujoco/minimize.py
@@ -505,7 +505,7 @@ def jacobian_fd(
   if bounds is None:
     eps_vec = eps * np.ones((n, 1))
   else:
-    mid = 0.5 * (bounds[1] - bounds[0])
+    mid = 0.5 * (bounds[0] + bounds[1])
     eps_vec = np.where(x > mid, -eps, eps)
   eps_vec *= np.maximum(1.0, np.abs(x))
   eps_vec = (eps_vec + x) - x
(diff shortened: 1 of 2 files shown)


yuvaltassa (approved):
Thanks!!

yuvaltassa (approved):

Issue #3099, opened 2026-02-15
Can't open usd file


yangfengzzz:
### Intro

Hi!

I build OpenUSD and open compile flag about usd. But ths simulate still can't open usd file. Is there officlal usd file can be loaded? I try the usda in the project like USDGolden.usda, but it's not work.

### My setup

Mujoco 3.5, Ubuntu 24.04 OpenUSD 25.11

havess:
Hey @yangfengzzz ! I'm going to need a bit more info to repro this:

Can I get your exact build steps and an example USD file you're trying to load?

Issue #3004, opened 2026-01-01
Bugs found when building MuJuCo with USD


Pap-let:
### Intro

Hi!

I am a graduate student at ECNU, I use MuJoCo for my research on EmbodiedAI.

### My setup

MuJoCo with USD on arm64 MacOS 26.1

### What's happening? What did you expect?

I built the mujuco myself with the MUJOCO_WITH_USD=True and expecting to drag and load usd files in the simulate.cc. 
Here are some bugs I have met with.

1. The official documentation on [MuJuCo](https://mujoco.readthedocs.io/en/latest/OpenUSD/building.html) does not match the latest building pipeline, the document is still using **USD_DIR** to specify the self-built USD lib instead of the lastest **pxr_DIR** and does not mention that the user need to set MUJOCO_WITH_USD to True in order to build mujuco with usd.
2. There seem to be some issue on the **mjUSEUSD** variable since I didn't find where it is defined in the whole project but its defination is checked in many src files such as "simulate/main.cc". Even if I set MUJOCO_WITH_USD to True I still can't drag and load usd files in **simulate.cc**. The **simulate.cc** won't print OpenUSD version as it is assumed to. It will try to load a usd file as an xml and throw XML Error because **mjUSEUSD** is never defined.
3. According to the github update log there should be no need to manually set plugin info with `export PXR_PLUGINPATH_NAME=/usr/local/lib/mujocoUsd/resources/*/plugInfo.json`, the official documentation needs an update.

### Steps for reproduction

1. build the mujuco with usd using 
```
cmake .. -DCMAKE_BUILD_TYPE=Release -DMUJOCO_WITH_USD=True
cmake --build . -j 30
sudo cmake --install . 
```
2. run `simulate` and try drag and load a usd file, it will throw you some error with XML.

### Confirmations

- [x] I searched the [latest documentation](https://mujoco.readthedocs.io/en/latest/overview.html) thoroughly before posting.
- [x] I searched previous [Issues](https://github.com/google-deepmind/mujoco/issues) and [Discussions](https://github.com/google-deepmind/mujoco/discussions), I am certain this has not been raised before.

SagarJadhav30:
I’m looking into this and plan to submit a patch + doc update. Assigning to myself.

SagarJadhav30:
cmake .. -DCMAKE_BUILD_TYPE=Release -DMUJOCO_WITH_USD=ON
cmake --build . -j
sudo cmake --install .
[...]

PR #3303, opened 2026-05-31
fix: fix USD loading for simulate (#3004, #3099)

nithin434:
Route USD file extensions through the USD decoder and   content-type mapping so drag-and-drop loads .usd/.usda/.usdc/.usdz instead of falling back to XML. #3004, #3099

### DIFF
--- a/simulate/main.cc
+++ b/simulate/main.cc
@@ -233,6 +233,12 @@ mjModel* LoadModel(const char* file, mj::Simulate& sim) {
     extension = filename_str.substr(dot_pos);
   }
 
+  std::string content_type;
+  if (extension == ".usd" || extension == ".usda" ||
+      extension == ".usdc" || extension == ".usdz") {
+    content_type = "model/usd";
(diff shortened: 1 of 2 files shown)


nithin434:
@yuvaltassa @havess can you review it

simulate/main.cc
```diff
@@ -241,7 +247,9 @@ mjModel* LoadModel(const char* file, mj::Simulate& sim) {
   } else if (extension == ".xml") {
     mnew = mj_loadXML(filename, nullptr, loadError, kErrorLength);
   } else {
-    mjSpec* spec = mj_parse(filename, nullptr, nullptr, loadError, kErrorLength);
+    mjSpec* spec = mj_parse(
```
havess, line 250: mj_parse will grab the file extension when loaded with no content_type and the usd decoder is registered with all 4 (usd, usdc, usda, usdz).

So I don't think this change is doing anything?

PR #3304, opened 2026-06-01
Use ring buffer for the SimProfiler

ttktjmt:
`Update()` called `erase(begin())` + `push_back()` on 11 vectors every frame, _O(N)_ per buffer. Replace with a `head_` write cursor and direct indexed writes (`buf[head_] = val`), advancing the cursor modulo 200 at the end of each update. Pass `head_` as ImPlot's `offset` parameter so the circular buffer renders correctly without data movement.


okmatija:
Thanks for the PR! We have not been very good with writing tests for the studio components, perhaps you could kick this off by adding a simple one for the SimProfiler. You could make sure it passes with and without your change e.g., by implementing the test in a first commit and doing the refactor in the next one.

ttktjmt:
@okmatija Thank you for the review!
Sure, I can add a simple test for `SimProfiler` as a first step. The design of the test contracts depends on whether they need to cover behavior both before and after this change, or only after.

Contracts covering both before and after the change:

- Calling `Update()` N times does not crash
- If `N ≤ max_frames`, exactly N frames are recorded
- If `N > max_frames`, it does not crash
- The most recently recorded values are retrievable
- After `Clear()`, subsequent calls to `Update()` work correctly

Contracts focused only on the behavior after this change:

- A single call to `Update()` increments the recorded frame count by 1 (i.e., `head` advances from 0 to 1)
- Calling `Update()` max_frames + 1 times wraps head back via % max_frames (i.e., `head` goes from max_frames to 1)
- Calling `Clear()` resets `head` to 0
- The N-th call to `Update()` writes to slot N-1 (e.g., the 1st call writes to slot 0, the 2nd to slot 1, with `dim_body` values)

Which scope would you prefer?

okmatija:
We don't want to maintain detailed tests I think, just some basic sanity checking is enough IMO. Also, looking at the API again I think we don't really have a useful public function we can use to implement the tests (we don't really want to check private implmeentation details like where `head_` is), perhaps we could add a `Summary()` function which returns a `Summary` struct some useful info e.g., counts and averages? With that data we could implement one test that does something like:

* Record N < max_frames and check summary.
* Record one more frame and check summary.
* Record one more frame (so N > max_frames) and check summary.
* Clear() and check summary is all zero or whatever makes sense.

The summary data may be useful in the UI somewhere. WDYT @haroonq ?

ttktjmt:
Sounds good. I'll keep it to one basic sanity test `src/experimental/platform/sim/sim_profiler_test.cc` built on a new `Summary` accessor.

Concrete shape I'm planning:

```cpp
struct MetricStats { float average = 0, min = 0, max = 0; };
struct Summary {
  int num_frames = 0;  // <= capacity, recorded frames
  int capacity = 0;    // == kProfilerMaxFrames
  MetricStats cpu_total, cpu_collision, cpu_prepare, cpu_solve, cpu_other;
  MetricStats dim_dof, dim_body, dim_constraint, dim_sqrt_nnz, dim_contact, dim_iteration;
};
Summary GetSummary() const; // nested method with same name is not allowed
```

Note: the ring buffer pre-fills every slot with 0 and `head_` wraps, so right now there's no way to distinguish a partially-filled buffer from a full one, I think. I can add a small counter `num_frames_` (reset in `Clear()`, incremented in `Update()`, saturating at max frames) so the counts and averages are computed only over real frames.

The single test follows your four steps:

- record capacity - 1
- → +1 (== capacity)
- → +1 (wrap, num_frames saturates)
- → `Clear()` (back to zero)

feeding deterministic mjData timer values so the averages are exact rather than depending on wall-clock timing.

@haroonq Happy to adjust if you'd prefer something else.

haroonq:
This seems fine to me.  Thanks for doing it!

ttktjmt:
@okmatija @haroonq 
I added the test. Please let me know when you want to adjust something

src/experimental/platform/sim/sim_profiler.cc
```diff
@@ -95,27 +94,60 @@ void SimProfiler::Update(const mjModel* model, const mjData* data) {
...
+    }
+    metric.average = static_cast<float>(sum / n);
+    return metric;
+  };
+
+  summary.cpu_total = stats(cpu_total_);
+  summary.cpu_collision = stats(cpu_collision_);
+  summary.cpu_prepare = stats(cpu_prepare_);
+  summary.cpu_solve = stats(cpu_solve_);
+  summary.cpu_other = stats(cpu_other_);
```
okmatija, line 129: Lets add a comment above this group like `// CpuTimeGraph stats`, and have a `// DimensionsGraph stats` comment for the ones below

src/experimental/platform/sim/sim_profiler.cc
```diff
@@ -95,27 +94,60 @@ void SimProfiler::Update(const mjModel* model, const mjData* data) {
...
+  const int n = num_frames_;
+  auto stats = [n](const std::vector<float>& buffer) {
+    MetricStats metric;
+    if (n == 0) {
+      return metric;
+    }
+    metric.min = metric.max = buffer[0];
+    double sum = 0;
+    for (int i = 0; i < n; ++i) {
+      float value = buffer[i];
```
okmatija: nit: I would save a line and just inline the buffer[i] calls

src/experimental/platform/sim/sim_profiler.cc
okmatija, line 42: I think you can do this instead:

```
  void SimProfiler::Clear() {
    head_ = 0;
    num_frames_ = 0;
  
    for (std::vector<float>* v : {&cpu_total_, &cpu_collision_, &cpu_prepare_,
                                  &cpu_solve_, &cpu_other_, &dim_dof_,
                                  &dim_body_, &dim_constraint_, &dim_sqrt_nnz_,
                                  &dim_contact_, &dim_iteration_}) {
      v->assign(kProfilerMaxFrames, 0);
    }
  }
```

src/experimental/platform/sim/sim_profiler.h
```diff
@@ -27,17 +27,40 @@ class SimProfiler {
...
 
+  // Returns aggregate statistics over the currently recorded frames.
+  Summary GetSummary() const;
+
   // Displays the profiling data using ImPlot.
   void CpuTimeGraph(ImVec2 plot_size = ImVec2(-1, 0));
   void DimensionsGraph(ImVec2 plot_size = ImVec2(-1, 0));
 
  private:
+  static constexpr int kProfilerMaxFrames = 200;
```
okmatija: nit: I think this could be named `kMaxFrames`, this better matches `num_frames_`

src/experimental/platform/sim/sim_profiler.h
```diff
@@ -27,17 +27,40 @@ class SimProfiler {
...
+    float average = 0;
+    float min = 0;
+    float max = 0;
+  };
+
+  // Snapshot of aggregate profiling statistics. Useful for tests and for a
+  // textual readout in the UI.
+  struct Summary {
+    int num_frames = 0;  // Recorded frames, saturating at capacity.
+    int capacity = 0;    // Maximum frames retained (== kProfilerMaxFrames).
```
okmatija: nit: this could also be named `max_frames`, and that would look nicer

test/experimental/platform/sim/sim_profiler_test.cc
```diff
@@ -0,0 +1,113 @@
...
+  mjsGeom* geom = mjs_addGeom(body, nullptr);
+  geom->type = mjGEOM_SPHERE;
+  geom->size[0] = 0.1;
+  std::unique_ptr<ModelHolder> holder = ModelHolder::FromSpec(spec);
+  ASSERT_THAT(holder->model(), NotNull()) << holder->error();
+
+  // Records one frame with a known average step time. The component CPU timers
+  // are left at zero, so cpu_total (and hence cpu_other) is exactly `step_ms`.
+  SimProfiler profiler;
+  mjData* data = holder->data();
```
okmatija: I think its better to just inline `holder->data()` everywhere

test/experimental/platform/sim/sim_profiler_test.cc
```diff
@@ -0,0 +1,113 @@
...
+
+using SimProfilerTest = MujocoTest;
+
+// Checks the average, min and max of a single profiler metric.
+#define EXPECT_METRIC_EQ(metric, avg, lo, hi) \
+  {                                           \
+    EXPECT_FLOAT_EQ((metric).average, (avg)); \
+    EXPECT_FLOAT_EQ((metric).min, (lo));      \
+    EXPECT_FLOAT_EQ((metric).max, (hi));      \
+  }
```
okmatija, line 38: I think you can cast to float when you call EXPECT_FLOAT_EQ so you can pass in mjtNum directly

test/experimental/platform/sim/sim_profiler_test.cc
```diff
@@ -0,0 +1,113 @@
...
+  mjData* data = holder->data();
+  auto record_frame = [&](float step_ms) {
+    data->timer[mjTIMER_STEP].duration = step_ms;
+    data->timer[mjTIMER_STEP].number = 1;
+    profiler.Update(holder->model(), data);
+  };
+
+  const int capacity = SimProfiler().GetSummary().capacity;
+  ASSERT_GT(capacity, 1);
+  const float nv = static_cast<float>(holder->model()->nv);
```
okmatija: Move this line to just before you call EXPECT_METRIC_EQ. This is better becuase it also reduces the scope of nv variable, so the program has fewer things to think about

test/experimental/platform/sim/sim_profiler_test.cc
```diff
@@ -0,0 +1,113 @@
...
+
+  // Record capacity - 1 frames with step times 1, 2, ..., capacity - 1.
+  for (int i = 1; i <= capacity - 1; ++i) {
+    record_frame(static_cast<float>(i));
+  }
+  {
+    SimProfiler::Summary s = profiler.GetSummary();
+    EXPECT_EQ(s.num_frames, capacity - 1);
+    EXPECT_EQ(s.capacity, capacity);
+    // Mean of 1..capacity-1, and cpu_other absorbs the whole step time.
```
okmatija: I dont understand the "absorbs the whole step time" part, can you clarify, or just remove the comment

test/experimental/platform/sim/sim_profiler_test.cc
```diff
@@ -0,0 +1,113 @@
...
+    EXPECT_METRIC_EQ(s.cpu_total, capacity / 2.0f, 1.0f,
+                     static_cast<float>(capacity - 1));
+    EXPECT_FLOAT_EQ(s.cpu_other.average, s.cpu_total.average);
+    // Dimensions are constant across frames.
```
okmatija: You can remove this comment

test/experimental/platform/sim/sim_profiler_test.cc
```diff
@@ -0,0 +1,113 @@
...
+  // so the retained window is 2..capacity+1 and num_frames saturates.
+  record_frame(static_cast<float>(capacity + 1));
+  {
+    SimProfiler::Summary s = profiler.GetSummary();
+    EXPECT_EQ(s.num_frames, capacity);
+    EXPECT_METRIC_EQ(s.cpu_total, (capacity + 3) / 2.0f, 2.0f,
+                     static_cast<float>(capacity + 1));
+  }
+
+  // Clear resets the counts and statistics (capacity is intrinsic and stays).
```
okmatija: nit: `// Clear the profiler and check for empty summary`. I think you can also remove the two EXPECT_METRIC_EQ calls, the stats are undefined if num_frames is 0

ttktjmt: The stats should return `0` when `num_frames == 0` as implemented below.

```cpp
    if (n == 0) {
      return metric;
    }
```

Happy to drop them if you'd prefer, but just checking.

ttktjmt:
@okmatija Thank you for the review!
I've addressed all your comments except [this one](https://github.com/google-deepmind/mujoco/pull/3304#discussion_r3381053576), which is still under discussion.

test/experimental/platform/sim/sim_profiler_test.cc
```diff
@@ -0,0 +1,107 @@
...
+  geom->type = mjGEOM_SPHERE;
+  geom->size[0] = 0.1;
+  std::unique_ptr<ModelHolder> holder = ModelHolder::FromSpec(spec);
+  ASSERT_THAT(holder->model(), NotNull()) << holder->error();
+
+  // Records one frame with a known average step time. The component CPU timers
+  // are left at zero, so cpu_total (and hence cpu_other) is exactly `step_ms`.
+  SimProfiler profiler;
+  auto record_frame = [&](float step_ms) {
+    holder->data()->timer[mjTIMER_STEP].duration = step_ms;
```
okmatija: You could have a `int frame` variable that you increment when you call `record_frame()`  (which can have no arguments). Maybe something like this:

```
 int duration = 0;

 // Record one frame with unit step time 
 auto record_frame = [&]() {
    data->timer[mjTIMER_STEP].duration = static_cast<mjtNum>(++duration);
    data->timer[mjTIMER_STEP].number = duration;  // Q: Does this need to also increment?
    profiler.Update(holder->model(), data);
  };
```

test/experimental/platform/sim/sim_profiler_test.cc
okmatija: When I wrote 'stats are undefined if num_frames is 0' I meant that the average member is undefined because you're dividing by 0. When you `return metric;` in your snippet you are returning an zero'd struct but avg=min=max=0 is not undefined e.g., it is the correct summary if you have 5 values all of which are 0.

okmatija:
LGTM, just one more suggestion related to `record_frame()`. Also something to think about but not change... the summary function could be called every frame and would compute the metrics again each frame by iterating over all the data, which is fine for now, if it becomes a problem later we coudl change to an incremental implementation. Right now we're just calling it in a test so its fine (no need to comment either IMO).

test/experimental/platform/sim/sim_profiler_test.cc
ttktjmt: I also think this is a better approach. The `number` should stay as `1`, otherwise the `Update()` computes `avg_total = duration / number` which becomes `1` all the time.

okmatija:
Hi @ttktjmt, can you sync this to the latest main, this should fix the CI and then I can this over the line!

I would do this myself, but don't know how (maybe this is because you didn't enable maintainer edits when you opened the PR?). Thanks for the contribution!

ttktjmt:
Sure, I can merge the current remote main into my feature branch. I think you can also directly commit to my source branch because I enabled it.

okmatija (approved):

PR #3318, opened 2026-06-05
fix: correct set_to_muscle() arg types from C arrays to std::array

### LINKED ISSUE #3282: `set_to_muscle()` Uses wrong arg types

Abzaek:
Closes #3282.

The pybind11 binding for `MjsActuator.set_to_muscle` declared `timeconst` and `range` as bare C arrays (`double[2]`), which pybind11 cannot properly handle for tuple/list defaults. Changed to `std::array<double, 2>` consistent with how `set_to_dcmotor` and other array-valued setters handle these parameters. The default for `timeconst` is now `{-1, -1}` matching the existing pattern for `range`.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -39,6 +39,9 @@ General
 Bug fixes
 ^^^^^^^^^
 - Fixed a bug in the ``mjz`` :ref:`decoder <mjpDecoder>` where unnormalized paths would fail to be read.
+- Fixed Python ``MjsActuator.set_to_muscle`` ignoring the ``timeconst`` and ``range`` arguments. The pybind11 lambda
+  declared these as bare C arrays (``double[2]``) which decayed to pointers; they now use ``std::array<double, 2>``
+  consistent with the other array-valued setters (:issue:`3282`).
 
 Version 3.9.0 (May 27, 2026)
 ----------------------------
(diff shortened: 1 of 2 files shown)


thowell:
should be fixed in https://github.com/google-deepmind/mujoco/commit/4e10f20be662a25c946d8b4638026c3dbafb04b9

PR #3323, opened 2026-06-06
Fix island disabled warp field dimension mismatch

mhasek:
Post the new MJWarp update sync an issue has come up where the `ENABLE_ISLANDS` flag being set to false causes the `_get_data_into_warp` call to fail. In this case the island fields are not allocated and set to a size of 0 on the mjwarp side but they are sized to the size of nv/ntree on the host MjData side. 

This PR checks if the flag is set to false and if there are fields that have size 0 and skips those fields. This PR also makes additions to the `test_get_data_into_warp` to validate this.

### DIFF
--- a/mjx/mujoco/mjx/_src/io.py
+++ b/mjx/mujoco/mjx/_src/io.py
@@ -1307,6 +1307,10 @@ def _get_data_into_warp(
       if isinstance(value, np.ndarray) and value.shape:
         result_field = getattr(result_i, field.name)
         if result_field.shape != value.shape:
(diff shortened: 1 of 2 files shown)


btaba (approved):

tkelestemur:
@btaba Seems like this is merged by @copybara-github here https://github.com/google-deepmind/mujoco/commit/58a8c6ee562ad0752ceeb670488a0015e08319ac.

Should we close this PR?

btaba:
Yes thank you!

PR #3325, opened 2026-06-09
TEST: gcc-12 -Wrestrict source fix

kevinzakka:


kevinzakka:
Folded into #3315 (commit 'Build the gcc jobs with LTO off too'). gcc-12 confirmed clean with the two source fixes + LTO-off.

PR #3350, opened 2026-06-17
CI: drop system_headers from Python-binding ccache sloppiness

kevinzakka:
Drop `system_headers` from the Python-binding ccache sloppiness so header-only changes aren't masked into stale-object ABI failures (full mechanism in the commit message).


mmossg (approved):
Good catch, thanks!

PR #3352, opened 2026-06-20
Add test for ellipsoid shell inertia with a != b != c

Pradyuman-aviator:
## Summary

Resolves `TODO (@thowell )` in `user_objects_test.cc` by adding a shell inertia test for an **asymmetric** ellipsoid (a ≠ b ≠ c), extending the existing `ShellInertiaEllipsoid` test which only covers the sphere case (a = b = c).

## Motivation

The current test uses equal semi-axes, which means any bug in how the engine maps inertia per-axis (e.g. an Ix/Iy/Iz swap, or an error specific to one axis) would be invisible  symmetry hides it. Testing with genuinely different semi-axes (0.3, 0.5, 0.7) exercises the general case and catches per-axis errors the symmetric test can't.

## My Approach

- Picked fully asymmetric semi-axes: **a=0.3, b=0.5, c=0.7**.
- Computed the expected mass for Body 2 using MuJoCo's own surface-area method (Thomsen's approximation, p=1.6075) × default density (1000) → **3084.614015**.
- Re-implemented the same Thomsen approximation and the ε-expansion method (expand semi-axes by ε=1e-6, subtract solid inertia of the original from the expanded version) directly in the test, so the expected Ix/Iy/Iz values are derived the same way the engine computes them internally.
- Kept the existing numerical cross-check pattern (Body 3 vs. Body 4: solid minus a slightly larger solid ≈ thin shell) as an independent, formula-free ground truth.

## Changes

**XML**
- Added 4-body fixture with `size="0.3 0.5 0.7"`, following the existing pattern:
  - Body 1: default density
  - Body 2: explicit mass (computed value above)
  - Body 3 & 4: ε-expanded solids for numerical subtraction

**C++ (`ShellInertiaEllipsoidAsymmetric`)**
- Implements Thomsen's approximation + ε-expansion analytically to compute expected Ix, Iy, Iz.
- `EXPECT_NEAR` assertions against Body 1 and Body 2, using `10 * kInertiaTol` (looser than default, since Thomsen's formula is itself an approximation  a tighter tolerance would flag approximation noise, not real bugs).
- Retains the Body 3/Body 4 subtraction check as an independent numerical validation.
- Skipped on float32 builds, consistent with the other shell inertia tests (ε=1e-6 needs float64 precision).

### DIFF
--- a/test/user/user_objects_test.cc
+++ b/test/user/user_objects_test.cc
@@ -1038,6 +1038,93 @@ TEST_F(MjCGeomTest, ShellInertiaEllipsoid) {
   mj_deleteModel(m);
 }
 
+TEST_F(MjCGeomTest, ShellInertiaEllipsoidAsymmetric) {
+  if constexpr (sizeof(mjtNum) == sizeof(float)) {
+    GTEST_SKIP() << "ShellInertia tests use radii differences of ~1e-8, which vanish in float32 precision";
+  }
+  // test ellipsoid with all-different semi-axes: a != b != c
+  static constexpr char xml[] = R"(
+  <mujoco>
+    <worldbody>
+      <body>
+        <geom type="ellipsoid" size="0.3 0.5 0.7" shellinertia="true"/>
+      </body>
+      <body>
+        <!-- mass is difference of body 4 and 3 masses -->
+        <geom type="ellipsoid" size="0.3 0.5 0.7" mass="3084.614015" shellinertia="true"/>
+      </body>
+      <body>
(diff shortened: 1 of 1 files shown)


Pradyuman-aviator:
@thowell  If got time have a look Thankyou!

Pradyuman-aviator:
gentle bump @yuvaltassa can you please run workflows ?

Pradyuman-aviator:
just a gentle bump @yuvaltassa  can you run the workflows ?

kevinzakka:
Stop spamming this repo.

Issue #3343, opened 2026-06-14
Any use of KTX textures crash mujoco simulate, and the ogl renderer.


gedalia:
### Intro

Hi!

I am a software engineer at Boston Dynamics, I support MuJoCo for our robotics simulations.

### My setup

I've been experimenting with the filament renderer (we rolled our own backend so it's not exactly the same as the mujoco one)  Currently we are using mujoco 3.7 via C and python, on linux and windows. 

### What's happening? What did you expect?

Attempting to use KTX textures in MuJoCo causes an immediate crash in mjr_uploadTexture with the error:
 ERROR: Number of channels not supported: 1
Upon inspecting the source code, mjCTexture::LoadKTX does not actually parse the KTX container or decompress the texture. Instead, it treats the entire raw binary file payload as a 1D pixel array of 1 channel, setting the width (w) to the file size in bytes. This blows up downstream code, and is pretty confusing till you realize that the ktx header needs to be parsed out of the bytestream. 

I understand that you aren't looking to support advanced rendering in the opengl renderer, but this means that adding any pbr ktx textures to a mjcf file (even if you leave the base texture a png) crashes the ogl renderer and the pip installed python viewer. Which really throws a wrench in some of our dev workflows.
Adding libktx loading for opengl should be fairly straightforward..
https://github.khronos.org/KTX-Software/libktx/index.html

But right now this seems to be the worst of all possible setups forcing wholesale switching to a filament renderer and breaking the python viewer with any use of compressed textures, or patching mujoco. I don't think there are callbacks for overriding any of this functionality, which still wouldn't really help with the pypi wheel.

### Steps for reproduction

Use a ktx texture in a mjcf file.

### Confirmations

- [x] I searched the [latest documentation](https://mujoco.readthedocs.io/en/latest/overview.html) thoroughly before posting.
- [x] I searched previous [Issues](https://github.com/google-deepmind/mujoco/issues) and [Discussions](https://github.com/google-deepmind/mujoco/discussions), I am certain this has not been raised before.

Nas01010101:
Saw this is assigned, so ignore if you've got it covered. PR in case it's useful.

It just skips textures `mjr_uploadTexture` can't handle (not 3/4-channel) with a warning instead of aborting, so KTX stops crashing `simulate`. Only stops the crash though, ofc still won't render without libktx 😕

haroonq:
[...]

PR #3354, opened 2026-06-21
render: skip textures the classic renderer can't upload

Nas01010101:
## Summary
KTX textures crash the classic OpenGL renderer and the Python viewer with `Number of channels not supported: 1`: `mjCTexture::LoadKTX` stores the file as an opaque single-channel blob for the Filament renderer, and `mjr_uploadTexture` aborts via `mju_error` because the classic path only handles 3- or 4-channel textures.

## Fix
Skip textures the classic renderer can't upload (`nchannel` not 3 or 4) with a warning instead of aborting, so the model still loads and the other textures render. `LoadKTX` and the Filament path are unchanged. The guard also covers the cube/skybox path, which previously over-read a single-channel blob as RGB.

## Test
Adds a `render_test.py` regression test that builds a model with a KTX texture and constructs an `MjrContext`; it aborts before this change and passes after.

## Proof
Rendered with the fix applied — the scene renders instead of crashing:

IMG_8194

Addresses #3343.

### DIFF
--- a/python/mujoco/render_test.py
+++ b/python/mujoco/render_test.py
@@ -14,6 +14,8 @@
 # ==============================================================================
 """Tests for MuJoCo Python rendering."""
 
+import os
+
 from absl.testing import absltest
 import mujoco
 import numpy as np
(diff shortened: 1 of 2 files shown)


haroonq (approved):

haroonq:
Thanks!

haroonq (approved):

PR #3358, opened 2026-06-23
Fix x86-windows builds.

BillyONeal:
Resolves:  `engine_memory.c.obj : error LNK2019: unresolved external symbol  __InterlockedExchangeAdd64 referenced in function _stackalloc`

This change suggested by GPT 5.5.

This was detected by https://github.com/microsoft/vcpkg/pull/52555


BillyONeal:
@kevinzakka Is there something that can be done to make this acceptable to merge? I observe https://github.com/google-deepmind/mujoco/blob/cc7fb98cf45ed697589c06e7e1d12b338ca53e98/src/engine/engine_crossplatform.h#L85 doesn't currently have this applied.

PR #3361, opened 2026-06-24
Probabilistic state estimation

jordanrule:
### Motivation

`mjModel.sensor_noise` has been a documented, XML-writable field since the native noise injection was removed (see changelog). Its presence signals _intent_: the user wants to associate a noise scale with each sensor. Until now, nothing in the engine read it at runtime.

This PR completes the design by adding `mj_sensorLogLik`, which turns that stored metadata into a callable function — the observation log-likelihood needed for importance weighting in a particle filter or the $R$ matrix diagonal in a Kalman-family estimator.

### What changed

| File | Change |
|---|---|
| `src/engine/engine_sensor.c` | Implementation of `mj_sensorLogLik` (~35 lines) |
| `src/engine/engine_sensor.h` | Internal declaration + doc comment |
| `include/mujoco/mujoco.h` | Public API declaration |
| `test/engine/engine_sensor_test.cc` | 7 new test cases under `SensorLogLikTest` |

### API

```c
// Gaussian log-likelihood of obs given current sensordata and per-sensor noise model.
// obs has the same layout as d->sensordata (length nsensordata).
// Sensors with sensor_noise[i] <= 0 are excluded and do not contribute to the result.
// Returns sum_i sum_j log N(sensordata[adr+j]; obs[adr+j], sigma_i^2).
// Suitable as a particle filter importance weight: w = exp(mj_sensorLogLik(m, d, obs)).
MJAPI mjtNum mj_sensorLogLik(const mjModel* m, const mjData* d, const mjtNum* obs);
```

Sensor noise is set in MJCF exactly as before:

```xml
<sensor>
  <jointpos name="encoder" joint="hip" noise="0.02"/>
  <velocimeter name="imu_vel" site="imu" noise="0.005"/>
</sensor>
```

A minimal particle filter weight update then looks like:

```c
// for each particle i:
mj_step(m, particles[i]);                             // propagate
log_weights[i] += mj_sensorLogLik(m, particles[i], observation);  // weight
```

### Design decisions

**Skipping sensors with `noise <= 0`** mirrors the existing convention in `apply_cutoff` where `cutoff = 0` means "ignore". This makes the function a no-op for models that haven't set noise values, with zero behavioral change to existing simulation code.

**All channels of a sensor share one `sigma`** — `sensor_noise` is a single scalar per sensor (`nsensor × 1`), not per channel. This matches the array layout already in `mjModel` and is the right level of granularity for most sensor models (an IMU accelerometer has the same spec noise on all three axes).

**Normalization constant is included** — `−dim·(log σ + ½ log 2π)` is part of the returned value. This makes the result a proper log-probability (not just a log-kernel), which matters when comparing likelihoods across models with different sensor configurations or when computing marginal likelihoods.

**No PRNG, no engine-side randomness** — consistent with the original removal rationale. `mj_sensorLogLik` is a pure function of `(m, d, obs)`. The user manages their own RNG for process noise and for sampling from the likelihood.

**`static const mjtNum kHalfLog2Pi`** — precomputed at compile time to avoid a `log` call in the inner loop. The literal `0.9189385332046727417803796` is ½·ln(2π) to 25 significant figures, verified against `python -c "import math; print(0.5*math.log(2*math.pi), '%.25f' % (0.5*math.log(2*math.pi)))"`.

### Tests

| Test | What it checks |
|---|---|
| `AllNoisesZeroReturnsZero` | No modeled sensors → exact 0 |
| `ZeroNoiseSensorIsIgnored` | Perturbing a `noise=0` channel has no effect |
| `PerfectObservationMatchesNormalization` | Perfect obs gives `−(log σ + ½ log 2π)` |
| `OneSigmaResidualReducesByHalf` | Residual of 1σ reduces log-likelihood by exactly 0.5 |
| `AdditiveAcrossSensors` | Two identical sensors → twice one sensor's contribution |
| `MultiDimSensorAllChannelsUseSameSigma` | `velocimeter` (dim=3) applies sigma to all 3 channels |
| `SymmetricInResidual` | `ll(+r) == ll(−r)` (Gaussian is symmetric) |
| `NoSensorsReturnsZero` | Empty model with `obs=nullptr` returns 0 cleanly |

Tests use the existing `MujocoTest` fixture and `MjTol`/`EXPECT_MJTNUM_EQ` helpers so they are correct under both double- and single-precision builds.

### DIFF
--- a/include/mujoco/mujoco.h
+++ b/include/mujoco/mujoco.h
@@ -405,6 +405,13 @@ MJAPI void mj_sensorVel(const mjModel* m, mjData* d);
 // Evaluate acceleration and force-dependent sensors.
 MJAPI void mj_sensorAcc(const mjModel* m, mjData* d);
 
+// Gaussian log-likelihood of obs given current sensordata and per-sensor noise model.
+// obs has the same layout as d->sensordata (length nsensordata).
+// Sensors with sensor_noise[i] <= 0 are excluded; they do not contribute to the result.
+// Returns sum_i sum_j log N(sensordata[adr+j]; obs[adr+j], sigma_i^2).
+// Suitable as a particle filter importance weight: w = exp(mj_sensorLogLik(m, d, obs)).
+MJAPI mjtNum mj_sensorLogLik(const mjModel* m, const mjData* d, const mjtNum* obs);
+
 // Evaluate position-dependent energy (potential).
 MJAPI void mj_energyPos(const mjModel* m, mjData* d);
 
--- a/src/engine/engine_sensor.c
+++ b/src/engine/engine_sensor.c
@@ -1653,6 +1653,41 @@ void mj_sensorAcc(const mjModel* m, mjData* d) {
 }
 
 
+//-------------------------------- log-likelihood -------------------------------------------------
+
+// Gaussian log-likelihood of obs given current sensordata and per-sensor noise.
+// obs has the same layout as d->sensordata (length nsensordata).
+// Sensors with sensor_noise[i] <= 0 are skipped; they do not contribute.
+// Suitable as a particle filter importance weight: w = exp(mj_sensorLogLik(m, d, obs)).
+mjtNum mj_sensorLogLik(const mjModel* m, const mjData* d, const mjtNum* obs) {
+  // 0.5 * log(2*pi); precomputed normalization constant for a scalar Gaussian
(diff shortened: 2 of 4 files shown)


jordanrule:
hi @yuvaltassa @haroonq - thank you for maintaining this awesome project!  I have tried to ensure this PR meets all guidelines https://github.com/google-deepmind/mujoco/blob/main/CONTRIBUTING.md - but if I have missed anything, or there is any more process needed to get this through, please let me know.

Issue #3368, opened 2026-06-29
MJX collision checking produces overflow casting warning


amacati:
### Intro

Hi!

I am a graduate student at TU Munich, where I use MJX for RL training. 

### My setup

Latest MJX from the main branch, Ubuntu24 on x86

### What's happening? What did you expect?

The current implementation of MJX throws a warning for box-box and convex-convex collisions. The exact warning is `lax_numpy.py:2799: RuntimeWarning: overflow encountered in cast`.

The reason is that the collision checking functions for box-box and convex-convex collisions use a `jp.where` selection to set infinite values to the maximum possible float value instead. This is done via `jp.where(jp.isinf(dist), jp.finfo(float).max, dist)`. Since dist is float32, but `jp.finfo(float)` uses float64, this results in the overflow error. The fix is to replace the `jp.finfo` with `jp.finfo(dist.dtype)`. I have taken the liberty to open a PR that adds a minimal test and fixes the error in the two affected functions.

I know that `mjx` is getting slowly deprecated in favor of `warp`, but given the minuscule size of this fix, I hope you do consider taking it in.

This is likely related to discussion #2226 and fixes it.

### Steps for reproduction

1. Run the code below.
2. Code errors because of the warning.

### Code required for reproduction

```python
  import warnings
  import jax
  import mujoco
  from mujoco import mjx

  # Two penetrating boxes -> exercises the box_box collider.
  XML = """
  <mujoco>
    <worldbody>
      <geom type="box" size="0.025 0.025 0.025" pos="0 0 0.025"/>
      <body pos="0 0 0.04">
        <freejoint/>
        <geom size="0.048 0.01 0.01" type="box"/>
      </body>
    </worldbody>
  </mujoco>
  """

  m = mujoco.MjModel.from_xml_string(XML)
  mx = mjx.put_model(m)
  dx = mjx.put_data(m, mujoco.MjData(m))

  with warnings.catch_warnings():
      warnings.simplefilter("error", RuntimeWarning)  # turn the warning into a hard error
      dx = jax.jit(mjx.kinematics)(mx, dx)
      dx = jax.jit(mjx.collision)(mx, dx)  # box_box -> finfo(float64).max cast into f32
  print("done, no warning")  # Not reached, code errors out because of the warning
```

### Confirmations

- [x] I searched the [latest documentation](https://mujoco.readthedocs.io/en/latest/overview.html) thoroughly before posting.
[...]

PR #3369, opened 2026-06-29
Fix overflow cast in MJX collisions for box-box and convex-convex

amacati:
Fixes the overflow cast in MJX collisions encountered in #3368 and possibly #2226.

To repeat the central part of #3368: The current implementation of MJX throws a warning for box-box and convex-convex collisions. The exact warning is lax_numpy.py:2799: RuntimeWarning: overflow encountered in cast.

The reason is that the collision checking functions for box-box and convex-convex collisions use a `jp.where` selection to set infinite values to the maximum possible float value instead. This is done via `jp.where(jp.isinf(dist), jp.finfo(float).max, dist)`. jax uses float32 by default, and so `dist` is a float32 array.  `jp.finfo(float)` uses float64 however, because jax treats Python's float built-in as double precision. The dtype mismatch then results in the overflow error, because we are setting elements of a float32 array to float64 (or at least attempt it). The fix is to replace the `jp.finfo` with `jp.finfo(dist.dtype)`. This PR adds a minimal test and fixes the error in the two affected functions.

### DIFF
--- a/mjx/mujoco/mjx/_src/collision_convex.py
+++ b/mjx/mujoco/mjx/_src/collision_convex.py
@@ -714,7 +714,7 @@ def _box_box(b1: ConvexInfo, b2: ConvexInfo) -> Collision:
   # Go back to world frame.
   pos = b2.pos + pos @ b2.mat.T
   n = normal @ b2.mat.T
-  dist = jp.where(jp.isinf(dist), jp.finfo(float).max, dist)
+  dist = jp.where(jp.isinf(dist), jp.finfo(dist.dtype).max, dist)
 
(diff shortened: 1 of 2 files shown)


btaba (approved):

amacati:
@btaba I spotted a caveat in the tests: When running the full suite, jax caches the traced collision function and does not recompile. Later, we then skip the warning in the new tests even if the cast is still overflowing, because the code for the warning does not get executed. Clearing the cache and forcing a recompile seemed the best solution to me, so I added that in the latest push.

btaba (approved):

Issue #2155, opened 2024-10-18
MJX ray has no implementation for hfield geoms


nico-bohlinger:
### The feature, motivation and pitch

The `mjx.ray` function is great to get depth images in MJX and it is used in the recently added rangefinder, but the handling of `hfield` geoms is currently not supported. To add this, one would need to port the [mj_rayHfield](https://github.com/google-deepmind/mujoco/blob/dcf2816543dd6b1fda14faa2b6d242c3b9c60f7f/src/engine/engine_ray.c#L453) function to `mjx`.

This feature would be highly appreciated to enable depth rendering in hfield environments / terrains for learning robot locomotion.

### Alternatives

``

btaba:
Hi @nico-bohlinger thanks for the heads up on the missing feature. Would you like to implement it?

shreyans413:
Hi, can i work on this issue?

Shivam040:
Hi, can i work on this issue?

yuvaltassa:
We have it now in MuJoCo Warp, so probably not worth your time

thowell:
this feature is not planned for mjx. please see mujoco warp [ray](https://mujoco.readthedocs.io/en/latest/mjwarp/api.html#mujoco_warp.ray) for ray casting with height fields.

PR #3378, opened 2026-07-03
Add height field support to mjx.ray

elliot-at-liminalnook:
`mjx.ray` has no intersection function for `HFIELD` geoms, so the dispatch loop silently skips height fields: every ray that should hit terrain instead returns `(dist=-1, geomid=-1)` while the same ray hits with `mujoco.mj_ray`. Issue #2155 tracked this and was closed once the Warp backend gained hfield rays, but the JAX backend still lacks it.

This adds `_ray_hfield`, a port of `mj_rayHfield`: the base box, the two surface triangles per grid cell, and the four prism side faces, where a side hit only counts if the hit point lies below the terrain edge interpolated between the neighboring data points. It also makes `ray` warn when a candidate geom has a type with no registered ray function, instead of silently treating it as a miss, and updates the one-line MJX feature-support entry.

I want to be upfront that there is an open in-flight PR for the same feature, #2804, which I only found after writing this. I'm posting this as complementary rather than competing. The overlap is the core port; what this adds on top is the silent-miss warning for unsupported geom types, an exact-parity test battery (9 rays checked bitwise against `mj_ray`: straight-down and oblique surface hits, side-face hits below the terrain edge, the base box from below, occlusion by a sphere above the terrain, and misses), and the doc-table line. The implementations differ in shape: this one composes per-face candidate distances with a running `minimum` and a per-face validity mask, whereas #2804 uses a compiled-for-fixed-size early-out. I'm happy to converge on whichever implementation the maintainers prefer, drop mine in favor of #2804's, or fold these tests and the warning onto #2804; the tests are implementation-agnostic and should transfer either way.

Separately, while testing I hit a pre-existing crash unrelated to this change: `mjx.ray` raises when the model has no materials (`nmat == 0`), because the visibility filter indexes `mat_rgba[geom_matid]` with `geom_matid == -1` into an empty array. All the tests here put a material on every geom to avoid it. I'm happy to send a follow-up that guards that filter if useful.

I prepared this change with AI assistance (Claude); I have not added it as a commit co-author, since it cannot sign the CLA, and I've reviewed the change and take responsibility for it.

### DIFF
--- a/doc/mjx.rst
+++ b/doc/mjx.rst
@@ -510,7 +510,7 @@ The following table compares feature support between MJX-Warp and MJX-JAX compar
      - Positions and directions
    * - Ray
      - All, BVH for meshes, hfield, and flex
-     - Slow for meshes, hfield and flex unimplemented
+     - Slow for meshes and hfield, flex unimplemented
 
 
 .. [1] Differentiability is `mostly supported <https://github.com/google-deepmind/mujoco/issues/2259>`__ in MJX-JAX but is
--- a/mjx/mujoco/mjx/_src/ray.py
+++ b/mjx/mujoco/mjx/_src/ray.py
@@ -15,6 +15,7 @@
 """Functions for ray interesection testing."""
 
 from typing import Sequence, Tuple
+import warnings
 
 import jax
 from jax import numpy as jp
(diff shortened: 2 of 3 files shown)


btaba:
We encourage folks to use MuJoCo Warp for this feature https://github.com/google-deepmind/mujoco/issues/2155#issuecomment-4707886038

Is that sufficient for your use-case?

elliot-at-liminalnook:
That works!

PR #3384, opened 2026-07-08
Fix .mjz loading in simulate

omarrayyann:
Loading .mjz archives in simulate failed because the archive's VFS was destroyed between parse and compile; fixed by sharing one mjVFS across both (also now surfacing the real error message).


omarrayyann:
also confirmed that the issue doesn't exist in Studio

havess (approved):
nice! Thanks.

PR #3392, opened 2026-07-10
Preserve small MuJoCo semantics in USD decoder

### LINKED ISSUE #3391: USD decoder does not preserve sites, model names, unlimited joints, and disabled-collider metadata
### Intro

I am working on MJCF <-> USD roundtrip validation for MuJoCo models, including MuJoCo Menagerie assets. The goal is to make USD import preserve MuJoCo-authored USD semantics closely enough that compiled-model comparison and short deterministic rollouts are useful.

### My setup

- MuJoCo built from source from `google-deepmind/mujoco` `main`.
- Build option: `MUJOCO_WITH_USD=ON`.
- API path: USD decoder plugin through `mj_parse` / `mj_loadXML`-style import, then compiled-model inspection.
- Platform: Linux x86_64.
- Source USD files are MuJoCo-authored USD, generated from MJCF models and then imported back into MuJoCo.

Several small USD decoder behaviors make MuJoCo-authored USD imports noisier or less faithful than necessary:

1. `MjcSiteAPI` prims that are also `UsdGeomGprim`s can decode as both sites and disabled visual geoms.
2. The decoded `mjSpec` model name falls back to `"MuJoCo Model"` instead of using an available USD stage/default-prim name.
3. USD revolute/prismatic joints with un-authored limits can decode as MuJoCo limited joints with `range="-inf inf"`.
4. Disabled colliders routed through the visual-geom path can lose `MjcCollisionAPI` metadata such as `group`.

Expected behavior: these values should be preserved during USD import. They are either explicitly authored in USD or implied by the absence of authored data, so the decoded `mjSpec`/compiled model should not add extra geoms, lose a better model name, turn unbounded joints into limited joints, or drop MuJoCo collision metadata on disabled colliders.

### Steps for reproduction

Use the USD decoder plugin with MuJoCo-authored USD inputs. Each item below is independently reproducible:

1. Site double-decode:
   - Start from a model with a MuJoCo site.
   - Convert it to MuJoCo-authored USD, where the site prim has `MjcSiteAPI` and is also a `UsdGeomGprim`.
   - Import that USD back into MuJoCo.
   - Actual: the prim can appear both as a site and as an extra disabled visual geom.
   - Expected: it should decode as a site only.

2. Model name fallback:
   - Import a USD stage with a meaningful default prim or root layer display name.
   - Actual: decoded model name can be `"MuJoCo Model"`.
   - Expected: use the default prim display name, default prim name, or root layer display name when available.

3. Unlimited hinge/slide limits:
   - Convert and import `unlimited_joint_limits.xml`
   - Actual before the fix: unbounded hinge/slide joints can import as limited joints with `range="-inf inf"`.
   - Expected: unbounded revolute/prismatic USD joints should remain unlimited MuJoCo joints unless finite limits are authored.

4. Disabled collider metadata:
   - Start from a MuJoCo geom that is contact-disabled but still has MuJoCo metadata such as `group`.
   - Convert to USD; the converter authors it as a disabled collider with `physics:collisionEnabled=false` and `MjcCollisionAPI`.
   - Import that USD back into MuJoCo.
   - Actual before the fix: the disabled collider routes through the visual-geom path and can lose `mjc:group`.
   - Expected: it should remain contact-disabled but still preserve `MjcCollisionAPI` metadata.

Useful small source MJCFs for generating MuJoCo-authored USD repro files:

Unlimited joint limits:

```xml
<mujoco model="unlimited_joint_limits">
  <compiler angle="degree"/>
  <option timestep="0.01" gravity="0 0 0"/>
  <worldbody>
    <body name="limited_link" pos="0 0 0.1">
      <joint name="limited_hinge" type="hinge" axis="0 0 1" limited="true" range="-45 45"/>
      <geom name="limited_geom" type="capsule" fromto="0 0 0 0.2 0 0" size="0.02" mass="0.1"/>
    </body>
    <body name="unlimited_hinge_link" pos="0 0.3 0.1">
      <joint name="unlimited_hinge" type="hinge" axis="0 0 1"/>
      <geom name="unlimited_hinge_geom" type="capsule" fromto="0 0 0 0.2 0 0" size="0.02" mass="0.1"/>
    </body>
    <body name="unlimited_slide_link" pos="0 0.6 0.1">
      <joint name="unlimited_slide" type="slide" axis="1 0 0"/>
      <geom name="unlimited_slide_geom" type="box" size="0.05 0.03 0.03" mass="0.1"/>
    </body>
  </worldbody>
</mujoco>
```

Disabled collider metadata:

```xml
<mujoco model="disabled_collider_group">
  <compiler inertiagrouprange="0 2"/>
  <worldbody>
    <body name="body" pos="0 0 0.1">
      <geom name="disabled_group2"
            type="box"
            size="0.05 0.05 0.05"
            group="2"
            contype="0"
            conaffinity="0"
            mass="0.1"/>
    </body>
  </worldbody>
</mujoco>
```

Site double-decode:

```xml
<mujoco model="site_decode">
  <worldbody>
    <body name="body" pos="0 0 0.1">
      <geom name="body_geom" type="sphere" size="0.03" mass="0.1"/>
      <site name="marker" type="sphere" size="0.01" pos="0.1 0 0" group="3"/>
    </body>
  </worldbody>
</mujoco>
```

The repro path is:

```bash
# 1. Convert the minimal MJCF to USD using the MuJoCo USD converter.
# 2. Import that USD into MuJoCo with the USD decoder plugin.
# 3. Save or inspect the decoded compiled model.
```

### Confirmations

LouRohanNV:
## Summary

This PR combines four small USD decoder preservation fixes:

- Treat `MjcSiteAPI` prims as sites only, not both sites and visual geoms.
- Initialize `mjSpec::modelname` from the USD stage/default prim when possible.
- Only mark revolute/prismatic joints as MuJoCo-limited when finite lower/upper limits are authored.
- Parse `MjcCollisionAPI` metadata on disabled colliders that are imported through the visual-geom path.

Fixes #3391
## Why

These are narrow roundtrip-fidelity issues. They do not introduce new schema or change the broad USD import architecture, but they remove noisy and sometimes physics-relevant differences in compiled MuJoCo models.

### Sites

`MjcSiteAPI` prims can also be `UsdGeomGprim`s for guide/visual purposes. The decoder should route them to the site list and avoid also treating them as visual geoms.

### Model Name

The decoder now tries, in order:

1. default prim display name
2. default prim name
3. root layer display name without extension

If none are available, the existing MuJoCo default remains unchanged.

### Unlimited Joint Limits

The decoder now enables MuJoCo joint limits only when both lower and upper USD limit attributes are authored and not the unbounded fallback pair `[-inf, inf]`.

### Disabled Collider Metadata

If a disabled collider is routed through the visual path, it remains contact-disabled, but MuJoCo-specific metadata such as `group` is still decoded from `MjcCollisionAPI`.

## USD / non-USD impact

The changes are contained to `plugin/usd_decoder`, which is only built when `MUJOCO_WITH_USD=ON`.

## Validation

Each source branch built locally with:

```bash
cmake --build ./mujoco/build --target usd_decoder_plugin --parallel 8
```

Roundtrip fixture/model evidence from this repo:

- Site double-decode count warnings disappear for models that author guide/site prims.
- Root model names no longer become `"MuJoCo Model"` when a better USD name exists.
- `fixtures/unlimited_joint_limits.xml` stops reporting `limited: False != True`.
- Lite6 narrow's disabled `gripper_lite_body.group` metadata is preserved in the direct USD-imported MJB comparison.
````

### DIFF
--- a/plugin/usd_decoder/usd_decoder.cc
+++ b/plugin/usd_decoder/usd_decoder.cc
@@ -50,6 +50,7 @@
 #include <pxr/base/tf/staticTokens.h>
 #include <pxr/base/tf/token.h>
 #include <pxr/base/vt/types.h>
+#include <pxr/usd/sdf/layer.h>
 #include <pxr/usd/sdf/path.h>
 #include <pxr/usd/usd/common.h>
 #include <pxr/usd/usd/prim.h>
@@ -114,6 +115,41 @@ struct UsdCaches {
 };
 
 constexpr const char* kUsdPrimPathKey = "usd_primpath";
+
+std::string StripExtension(std::string filename) {
+  const std::size_t slash = filename.find_last_of("/\\");
+  if (slash != std::string::npos) {
+    filename = filename.substr(slash + 1);
+  }
+  const std::size_t dot = filename.find_last_of('.');
+  if (dot != std::string::npos) {
+    filename.erase(dot);
+  }
+  return filename;
+}
+
+std::string GetStageModelName(const pxr::UsdStageRefPtr& stage) {
+  pxr::UsdPrim default_prim = stage->GetDefaultPrim();
+  if (default_prim) {
+    std::string display_name = default_prim.GetDisplayName();
+    if (!display_name.empty()) {
+      return display_name;
+    }
+    std::string prim_name = default_prim.GetName().GetString();
+    if (!prim_name.empty()) {
+      return prim_name;
+    }
+  }
+
+  if (pxr::SdfLayerHandle root_layer = stage->GetRootLayer()) {
+    std::string layer_stem = StripExtension(root_layer->GetDisplayName());
+    if (!layer_stem.empty()) {
+      return layer_stem;
+    }
+  }
+
+  return "";
+}
 
 void SetUsdPrimPathUserValue(mjsElement* element,
                              const pxr::SdfPath& prim_path) {
@@ -2589,6 +2625,10 @@ void PopulateSpecFromTree(pxr::UsdStageRefPtr stage, mjSpec* spec,
 
 mjSpec* ParseStage(const pxr::UsdStageRefPtr stage) {
   mjSpec* spec = mj_makeSpec();
+  std::string model_name = GetStageModelName(stage);
+  if (!model_name.empty()) {
+    mjs_setString(spec->modelname, model_name.c_str());
+  }
 
   std::unique_ptr<Node> root = BuildKinematicTree(stage);
 
--- a/plugin/usd_decoder/usd_decoder.cc
+++ b/plugin/usd_decoder/usd_decoder.cc
@@ -1921,6 +1921,10 @@ void ParseUsdGeomGprim(mjSpec* spec, const pxr::UsdPrim& gprim,
     }
   }
 
+  if (gprim.HasAPI<pxr::MjcPhysicsCollisionAPI>()) {
+    ParseMjcPhysicsCollisionAPI(geom, pxr::MjcPhysicsCollisionAPI(gprim));
+  }
+
   if (gprim.HasAPI<pxr::MjcPhysicsImageableAPI>()) {
     auto imageable_api = pxr::MjcPhysicsImageableAPI(gprim);
     auto group_attr = imageable_api.GetGroupAttr();
--- a/plugin/usd_decoder/usd_decoder.cc
+++ b/plugin/usd_decoder/usd_decoder.cc
@@ -2348,9 +2348,12 @@ void ParseUsdPhysicsJoint(mjSpec* spec, const pxr::UsdPrim& prim, mjsBody* body,
       mj_joint->axis[2] = 1;
     }
 
+    pxr::UsdAttribute lower_attr = revolute.GetLowerLimitAttr();
+    pxr::UsdAttribute upper_attr = revolute.GetUpperLimitAttr();
     float lower, upper;
-    if (revolute.GetLowerLimitAttr().Get(&lower) &&
-        revolute.GetUpperLimitAttr().Get(&upper)) {
+    if (lower_attr.HasAuthoredValue() && upper_attr.HasAuthoredValue() &&
+        lower_attr.Get(&lower) && upper_attr.Get(&upper) &&
+        !(std::isinf(lower) && lower < 0 && std::isinf(upper) && upper > 0)) {
       mj_joint->limited = mjLIMITED_TRUE;
       if (spec->compiler.degree) {
         mj_joint->range[0] = lower;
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: plugin/usd_decoder/usd_decoder.cc [hunk 1]
```diff
@@ -98,6 +116,41 @@ struct UsdCaches {
 
 constexpr const char* kUsdPrimPathKey = "usd_primpath";
 
+std::string StripExtension(std::string filename) {
```
havess: nit: Can this use std::string_view?

plugin/usd_decoder/usd_decoder.cc
```diff
@@ -2430,9 +2352,12 @@ void ParseUsdPhysicsJoint(mjSpec* spec, const pxr::UsdPrim& prim, mjsBody* body,
       mj_joint->axis[2] = 1;
     }
 
+    pxr::UsdAttribute lower_attr = revolute.GetLowerLimitAttr();
+    pxr::UsdAttribute upper_attr = revolute.GetUpperLimitAttr();
     float lower, upper;
-    if (revolute.GetLowerLimitAttr().Get(&lower) &&
-        revolute.GetUpperLimitAttr().Get(&upper)) {
+    if (lower_attr.HasAuthoredValue() && upper_attr.HasAuthoredValue() &&
```
havess: Can we skip these here and just check the inf values?

plugin/usd_decoder/usd_decoder.cc
```diff
@@ -2459,9 +2384,12 @@ void ParseUsdPhysicsJoint(mjSpec* spec, const pxr::UsdPrim& prim, mjsBody* body,
       mj_joint->axis[1] = 0;
       mj_joint->axis[2] = 1;
     }
+    pxr::UsdAttribute lower_attr = prismatic.GetLowerLimitAttr();
+    pxr::UsdAttribute upper_attr = prismatic.GetUpperLimitAttr();
     float lower, upper;
-    if (prismatic.GetLowerLimitAttr().Get(&lower) &&
-        prismatic.GetUpperLimitAttr().Get(&upper)) {
+    if (lower_attr.HasAuthoredValue() && upper_attr.HasAuthoredValue() &&
```
havess: ditto

plugin/usd_decoder/usd_decoder.cc
LouRohanNV: That's more than a nit, I question my earlier decision to even make a helper here and found that we didn't need one.

plugin/usd_decoder/usd_decoder.cc
LouRohanNV: Yes, we can

plugin/usd_decoder/usd_decoder.cc
LouRohanNV: Yes we can

havess (approved):
Nice, thanks!

PR #3394, opened 2026-07-10
Check physics-purpose material bindings in USD decoder for contact properties

### LINKED ISSUE #3393: USD decoder misses physics-purpose material bindings on colliders
### Intro

I am working on MJCF <-> USD roundtrip validation for MuJoCo models. One validation method compares physics-relevant compiled model fields after converting MJCF to USD and importing the USD back into MuJoCo.

### My setup

- MuJoCo built from source from `google-deepmind/mujoco` `main`.
- Build option: `MUJOCO_WITH_USD=ON`.
- API path: USD decoder plugin import, then compiled-model inspection.
- Platform: Linux x86_64.
- Source USD files are MuJoCo-authored USD generated from MJCF with contact properties.

USD material bindings can use different purposes. A collider can bind one material for visual appearance and another material for physics.

Current behavior: the MuJoCo USD decoder computes the default/all-purpose material binding for colliders. That preserves the visual material assignment, but can miss a physics-purpose material carrying contact/friction properties.

Expected behavior: when a collider has a physics-purpose material binding, the decoder should read physics/contact properties from that material while still using the default binding for visual material assignment.

In roundtrip testing, missing the physics-purpose binding causes active collider friction and solver parameters to fall back to MuJoCo defaults even though the USD contains authored physics material data.

### Steps for reproduction

1. Convert the minimal MJCF below to MuJoCo-authored USD using the MuJoCo USD converter.
2. Import the USD into MuJoCo with the USD decoder plugin.
3. Inspect the decoded geom fields for `friction`, `solref`, `solimp`, `margin`, and `gap`.
4. Actual before the fix: contact properties can fall back to defaults because the physics-purpose material binding is not parsed.
5. Expected: the decoded geom preserves the authored contact/friction fields.

```xml
<mujoco model="active_contact_props">
  <option timestep="0.002"/>

  <asset>
    <material name="contact_mat" rgba="0.8 0.25 0.15 1"/>
  </asset>

  <worldbody>
    <geom name="contact_box"
          type="box"
          size="0.12 0.08 0.04"
          pos="0 0 0.04"
          friction="0.37 0.11 0.07"
          solref="0.013 1.7"
          solimp="0.82 0.93 0.006 0.47 1.8"
          margin="0.03"
          gap="0.006"
          material="contact_mat"/>
  </worldbody>
</mujoco>
```

The repro path is:

```bash
# 1. Convert fixtures/active_contact_props.xml to USD.
# 2. Import the USD with MuJoCo's USD decoder plugin.
# 3. Save or inspect the decoded model and compare the active geom contact fields.
```

### Confirmations

anishesg:
The USD decoder previously only checked default/all-purpose material bindings when importing colliders, missing physics-purpose material bindings that carry contact properties like friction, solref, solimp, margin, and gap. In USD, a collider can bind one material for visual appearance (default purpose) and a separate material for physics simulation (physics purpose).

The fix modifies `ParseUsdPhysicsCollider` in `plugin/usd_decoder/usd_decoder.cc` to query physics-purpose material bindings before falling back to default bindings. When a physics-purpose binding exists with physics material APIs, the decoder now correctly extracts contact properties from that material. The default binding continues to provide visual material assignment. This ensures MJCF→USD→MJCF roundtrips preserve authored contact properties.

Fixes #3393

### DIFF
--- a/plugin/usd_decoder/usd_decoder.cc
+++ b/plugin/usd_decoder/usd_decoder.cc
@@ -110,10 +110,14 @@ struct UsdCaches {
   pxr::UsdGeomXformCache xform_cache;
   pxr::UsdShadeMaterialBindingAPI::BindingsCache bindings_cache;
   pxr::UsdShadeMaterialBindingAPI::CollectionQueryCache collection_query_cache;
+  pxr::UsdShadeMaterialBindingAPI::BindingsCache physics_bindings_cache;
+  pxr::UsdShadeMaterialBindingAPI::CollectionQueryCache
+      physics_collection_query_cache;
   std::map<pxr::SdfPath, mjsMaterial*> parsed_materials;
 };
 
 constexpr const char* kUsdPrimPathKey = "usd_primpath";
+const pxr::TfToken kPhysicsMaterialPurpose("physics");
 
 void SetUsdPrimPathUserValue(mjsElement* element,
                              const pxr::SdfPath& prim_path) {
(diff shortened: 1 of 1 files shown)


LouRohanNV:
I cloned this PR locally and built the USD decoder target with USD enabled:

```bash
cmake --build /tmp/mujoco-pr3394-build --target usd_decoder_plugin --parallel 8
```

It currently does not compile against the OpenUSD headers used by MuJoCo. The failure is in `plugin/usd_decoder/usd_decoder.cc`, where the physics-purpose `ComputeBoundMaterial` call passes the material purpose token before the cache pointers. For the cached overload, the cache pointers come first and the material purpose token comes after them.

The relevant compiler error is:

```text
error: no matching function for call to
UsdShadeMaterialBindingAPI::ComputeBoundMaterial(
    TfToken&, BindingsCache*, CollectionQueryCache*)
```

I also think the behavior should be a little more careful than only checking whether a physics-purpose material binding exists. A physics-purpose material can exist without actually authoring the physics/contact APIs, and in that case the decoder should still be able to fall back to the default/all-purpose material for physics/contact fields.

I pushed a branch with the version I have been testing here:

https://github.com/LouRohanNV/mujoco/tree/roundtrip-pr/usd-decoder-physics-materials

The main differences in that branch are:

- use the cached `ComputeBoundMaterial` overload with the correct argument order;
- prefer physics-purpose material bindings for contact properties;
- track whether physics/contact APIs were actually parsed, rather than only whether a physics-purpose material was bound;
- fall back to the default/all-purpose material when the physics-purpose material does not provide physics/contact APIs;
- keep separate caches for default material lookup and physics-purpose material lookup.

Maybe this PR could be updated in that direction, or we could use that branch as the basis for the final fix.

anishesg:
ah good catch on the arg order, fixed. also switched to tracking whether physics APIs were actually parsed rather than just checking if the binding exists, and added separate caches like your branch. should compile now

havess:
Generally LGTM, @LouRohanNV does this work for your test assets?

LouRohanNV:
@havess - Yes, this PR ended up matching my fix branch line-per-line, so yes, it works for my test assets.

PR #3396, opened 2026-07-11
Remove quadratic scratch from native island discovery

### LINKED ISSUE #3388: Proposal: remove quadratic native island-discovery scratch for large tree counts
### The feature and motivation 

Native `mj_island` currently materializes an `ntree x ntree` byte adjacency matrix and an
uncompressed `ntree x ntree` integer column-index array before flood fill. The same construction was
still present on main at `dc7581a` (checked 2026-07-10):
[engine_island.c](https://github.com/google-deepmind/mujoco/blob/dc7581acfaac722b1251d3b2435ba45d47068385/src/engine/engine_island.c#L380-L415).

This makes temporary stack memory quadratic even when the constraint incidence is sparse. I would
like to ask whether maintainers would support a native-only PR that constructs connected components
directly from constraint/tree incidence, without materializing the tree adjacency matrix.

I tested a file-local union-by-canonical-minimum-root prototype. It preserves ascending,
deterministic island IDs and leaves the existing `mj_floodFill` API and tests in place. For a
generated singleton family (one free body and one body-to-world `connect` per tree, giving `3*ntree`
scalar rows), the measured peak `mj_island` stack follows the source allocation formulas exactly:

| `ntree` | current dense path | direct-incidence prototype | ratio |
|---:|---:|---:|---:|
| 64 | 22,816 B | 1,056 B | 21.6x |
| 256 | 336,928 B | 4,128 B | 81.6x |
| 1,024 | 5,279,776 B | 16,416 B | 321.6x |
| 4,096 | 84,033,568 B | 65,568 B | 1,281.6x |

For this family, the current peak is `5*ntree^2 + 36*ntree + 32` bytes and the prototype peak is
`16*ntree + 32` bytes. A focused 1,024-tree test also fails on the current path with a 2 MiB arena
and passes unchanged with the prototype.

The complete synthetic matrix supports investigating the direct stage, but not a general
end-to-end speed claim. These are paired baseline/candidate geometric speedups over four topology
families and four tree counts, with 99% bootstrap intervals; values above one favor the prototype:

| measured scope | speedup | 99% CI | pairs |
|---|---:|---:|---:|
| direct `mj_island` | 1.6847x | [1.5494, 1.8333] | 480 |
| `mj_fwdPosition` | 1.1782x | [1.1409, 1.2174] | 480 |
| full `mj_step` | 1.0355x | [1.0254, 1.0460] | 480 |

The prespecified full-step gate required a 99% lower bound of at least 1.05x; it fails, and even the
upper bound is below 1.05x. All four pooled full-step topology-family lower bounds exceed the 0.98
regression threshold, but 19 individual workload/mode cells have lower bounds below 0.98. This is
therefore stage-level evidence under a pooled-family policy, not a per-cell no-regression result.

A separate Callgrind run at 4,096 trees collected 58,071,749 instructions across three matched
`mj_island` invocations on the current path and 9,092,275 on the prototype (6.39x). The three dense
adjacency clears account for approximately 50.33 million instructions, or 86.7% of the current-path
total. This profile includes setup-time island calls as well as the explicit probe call; it is a
separate mechanistic check, not another timing sample. Both Callgrind processes emitted a nonfatal
`brk segment overflow` warning.

Correctness checks completed so far:

- exact agreement between dense reference, sparse-incidence BFS, and DSU on 10,000 deterministic
  generated hypergraphs;
- exact equality of all island topology/count/map arrays on 16 generated models plus shipped
  humanoid and flex fixtures;
- zero recorded relative error in `qpos`, `qvel`, and `qacc` for those tested horizons;
- production-path tests for static-first incidence, repeated scalar rows, and per-row flex equality
  rescanning, with mutation tests;
- 15/15 focused CTest cases and 300/300 shuffled repetitions;
- complete output checksums matched across repeated arena rewinds for all four generated topology
  families at 1,024 and 4,096 trees.
- a fresh final-commit rerun passed the 18-model runtime parity matrix with exact island arrays and
  zero recorded trajectory error.

This relates to the large-scene scaling direction in #1671, but the proposal is limited to native
island discovery.

Questions before preparing a PR:

1. Is a native-only change acceptable, or should this be coordinated with MuJoCo Warp first?
2. Is removing quadratic scratch worth the measured small-model stage regression, or would a
   dense/direct threshold be preferable?
3. Would maintainers prefer retaining the current `O(nefc)` first-tree scratch, instead of the
   prototype's second constraint-incidence scan?
4. Which real models and platforms should be required in the benchmark matrix for a PR?

### Alternatives

- Keep the dense path below a tree-count threshold and use direct incidence above it. This may
  preserve current small-model performance but adds a second production path.
- Build sparse CSR adjacency and keep flood fill. This is conventional, but requires edge counting,
  allocation, and duplicate handling before traversal.
- Use direct-incidence DSU while retaining `efc_tree` scratch. This avoids the second constraint scan
  and may reduce the 64-tree regression at the cost of `O(nefc)` temporary memory.
- Keep the existing implementation. It is competitive at small `ntree`, but retains quadratic
  scratch and scans.

### Additional context

- Tested MuJoCo source: `f04b0fcd3d5a8f1af798c1fb18f7abac8e89c270`; current main was checked
  separately and retains the dense construction.
- Environment: Docker on WSL2, Ubuntu 24.04, GCC 13.3, Release `-O3 -DNDEBUG`, LTO, AVX, one pinned
  CPU.
- Timing protocol: five warmup process pairs and 30 measured process pairs, randomized within each
  pair. Each process loaded identical MJCF and ran the same mode-specific preparation and timed
  region. Direct-stage runs restored the same post-constraint arena boundary before each
  `mj_island`; position and step runs timed their corresponding native entry points. Full output
  checksums matched within every reported pair.
- Full-step baseline-first and candidate-first estimates were 1.0363x and 1.0347x. An independent
  workload-adjusted order-effect audit found 1.0076x with 99% CI [0.9947, 1.0207], so no order bias
  was detected; this is not proof of absence.
- The timing matrix is synthetic. The measured full-step effect misses the prespecified claim gate.
- The full MuJoCo suite, sanitizers, non-Linux platforms, and real user workloads have not yet been
  validated.
- MuJoCo Warp is a separate implementation. It was not modified or benchmarked here, and this work
  is not a Warp port. Its current island code also uses dense tree adjacency, which is why I am
  asking whether scope should be coordinated rather than assuming native-only ownership.

teerthsharma:
## Summary

`mj_island` currently materializes an `ntree x ntree` byte adjacency matrix and an uncompressed `ntree x ntree` integer column array before flood fill. This change computes the same connected-component (H0) partition directly from constraint/tree incidence with a file-local disjoint-set forest.

The canonical representative is always the minimum tree index, so island IDs remain deterministic and ascending. Repeated scalar rows retain the existing reuse behavior, flex equality rows are rescanned because their incidence can differ, and `mj_floodFill` remains unchanged.

This removes the quadratic allocations and the `O(nefc)` first-tree scratch from the production path. There is no public API, documented-semantics, persistent-cache, or MuJoCo Warp change.

## Memory scaling

For the generated singleton family used in #3388, peak `mj_island` stack use changes from `5*ntree^2 + 36*ntree + 32` bytes to `16*ntree + 32` bytes:

| `ntree` | dense path | direct incidence | reduction |
|---:|---:|---:|---:|
| 64 | 22,816 B | 1,056 B | 21.6x |
| 256 | 336,928 B | 4,128 B | 81.6x |
| 1,024 | 5,279,776 B | 16,416 B | 321.6x |
| 4,096 | 84,033,568 B | 65,568 B | 1,281.6x |

The formulas reproduce every measured size exactly. The regression test reverse-validates the failure mode: a 1,024-tree model fails on the dense path in a 2 MiB arena when the 4,194,304-byte column array is requested, and passes unchanged with direct incidence.

## Validation

Fresh verification was run from PR head `218b07d7` on current `main` at `4a03a617` in pinned Ubuntu 24.04 containers:

- full GCC 13.3 Release build completed;
- 1,851/1,851 CTest entries passed serially (one pre-existing locale test skipped);
- all 15 focused island cases passed, including static-first incidence, repeated rows, flex rescanning, and the bounded arena;
- 300/300 shuffled island-test executions passed;
- Clang 18 AddressSanitizer passed all 15 island cases with leak detection enabled;
- no DSU helper is exported from `libmujoco.so`;
- the final tree is byte-identical to the implementation used for the reported direct-incidence evidence and contains no cache or public `mjData` changes.

The native parity harness passed 59/59 checks, including 10,000 deterministic generated hypergraphs. Exact island arrays and trajectories matched on 18/18 generated and shipped models; maximum recorded relative error in `qpos`, `qvel`, and `qacc` was 0.

## Native benchmark evidence

The complete paired matrix reported in #3388 used four generated topology families, four tree counts, three scopes, five warmup pairs, and 30 measured pairs per cell. Baseline and candidate benchmark sources were byte-identical, each measured pair had matching complete-output checksums, and intervals use 20,000 deterministic paired-bootstrap resamples at 99% confidence:

| scope | geometric speedup | 99% CI | pairs |
|---|---:|---:|---:|
| direct `mj_island` | 1.6847x | [1.5494, 1.8333] | 480 |
| `mj_fwdPosition` | 1.1782x | [1.1409, 1.2174] | 480 |
| full `mj_step` | 1.0355x | [1.0254, 1.0460] | 480 |

An independent Callgrind profile at 4,096 trees measured a 6.39x instruction-count ratio in favor of direct incidence. Dense adjacency clearing accounted for 86.7% of current-path instructions in that profile, providing a mechanistic check independent of wall-clock timing.

The timing matrix is synthetic. The direct-stage and scratch-scaling claims are supported; the full-step result is reported as measured context, not as a universal MuJoCo or Warp speedup claim.

Fixes #3388.


yuvaltassa:
I see what you're doing and I think I like it, but I want to do it a bit differently and also measure perf and various things that won't be fun over GH review. Is attribution important to you? If not and it's ok with you, I prefer to do it myself using this PR as a reference.

teerthsharma:
I see what you're doing and I think I like it, but I want to do it a bit differently and also measure perf and various things that won't be fun over GH review. 
 --I completely understand—managing complex performance profiling and refactoring over GitHub comments isn't efficient.



 Is attribution important to you? 
--To be transparent, I am an independent student researcher currently balancing my studies with significant student debt. Open source is the primary way I propagate my topological approaches to find genuine opportunities in Frontier Labs Because of that, attribution and building a track record are incredibly important to me; it is exactly what will help alleviate the daily struggle to make ends meet.



prefer to do it myself using this PR as a reference.
--Since handling the profiling and execution over a standard PR review is tough, I would love the chance to collaborate directly with you on an apprenticeship or research basis collaboration further from GitHub in any form paid or even unpaid to get this integrated.

However, if that kind of collaboration isn't possible right now, I completely understand. I would still love to see this feature integrated using this PR as a reference  , even without any credits other then reference I will understand  . Thank you for your time and the feedback Hope I can catch your eyes few more times hahah !! @yuvaltassa

yuvaltassa:
Ha!

Okay I buy that. Let's do a substantive review and I'll do the cosmetic nits myself later.

1. Look at `floodFill`'s test coverage, I want that levelof coverage for dsu. You'll need to add MJAPI to expose these from the binary (but not mujoco.h)
2. Memory is the reason to do this, sure, but those numers are arithmetic not "benchmark evidence". But we do want a benchmark, like the ones in `test/benchmark/`. Show me that your union-find is not slower than floodFill, or at least not much slower. The question here is how does one invent a sensible corpus of graphs? I leave it you 🙂

yuvaltassa:
I should also add that I expect both implementations to be extremely fast in any reasonable scenario. This is for completeness.

But unit test coverage is definitely necessary.

teerthsharma:
> 2. Memory is the reason to do this, sure, but those numers are arithmetic not "benchmark evidence". But we do want a benchmark, like the ones in `test/benchmark/`. Show me that your union-find is not slower than floodFill, or at least not much slower. The question here is how does one invent a sensible corpus of graphs? I leave it you 🙂

I’ve settled on a deterministic S²–Vietoris–Rips corpus for the benchmark.
The idea is to place bodies on a sphere and connect nearby bodies using geodesic distance. The Rips radius controls expected graph degree, giving reproducible sparse, critical, and more connected island configurations. I then lower those graphs into MuJoCo-style tree incidences, including contacts, grounded bodies, articulated bridges, repeated constraints, inactive trees, and temporally stable/birth/death cases. The corpus covers sizes from 64 to 4096 nodes, multiple edge orders, many-small-island and giant-island cases, redundant cycles, and bridge changes. Every case has an independently verified connected-component partition, checksum, and nidof , unto testing right now will update you as soon as possible

teerthsharma:
I tried to address both points carefully:

1. The DSU helpers are exposed from the binary through the private MJAPI seam in `engine_island.h`, without adding anything to public `mujoco.h`.
2. I added direct DSU unit coverage and a benchmark under `test/benchmark/` comparing the actual exported DSU helpers with `mj_floodFill`.

The full island test suite passes **30/30 tests** in a fresh WSL2 Release build. This includes empty inputs, activation, path compression, redundant and reversed edges, inactive trees, adversarial chains, deterministic assignment, a 2,000-seed differential test, and the new flex-stiffness island test.

For the graph corpus, I used six deterministic S²–Vietoris–Rips families. They represent sparse stable islands, critical bridges, dense components, grounded/static incidences, repeated constraints, and large near-critical graphs. Corpus generation and the independent connected-component oracle are outside the timed region.

| Corpus | Nodes | Incidences | DSU vs `mj_floodFill` |
|---|---:|---:|---:|
| Stable sparse | 64 | 68 | **8.46× faster** |
| Critical bridge | 256 | 504 | **13.97× faster** |
| Supercritical dense | 256 | 1,521 | **11.45× faster** |
| Grounded/static repeated | 256 | 1,886 | **10.79× faster** |
| Stable repeated | 1,024 | 1,205 | **55.43× faster** |
| Critical large | 1,024 | 3,626 | **23.80× faster** |

Both implementations produced the same independently verified component checksum for every corpus.

I also rebased onto the latest branch state and adapted the new flex-stiffness coupling to the DSU union pass. I kept the topology-generation machinery out of the production binary; only the small reproducible corpus generator is part of the benchmark target.

I hope this is closer to the level of coverage and evidence you had in mind. I’d appreciate any further corrections or suggestions.

yuvaltassa:
Nice! The benchmark did its job, DSU is not slower. Now that it has convinced us, let's delete it, we don't want to maintain a benchmark comparing production code to an algorithm it no longer uses. Leave mj_floodFill and its existing tests in place.

Remaining nits, then I think we're done:

1. Naming: drop the `_mjPRIVATE_` wrappers. Make the static functions non-static and export them directly with MJAPI from engine_island.h, exactly like mj_floodFill: `mj_dsuInit`, `mj_dsuUnion`, `mj_dsuFind`, `mj_dsuAssign` (maybe Union -> Merge and Find -> Root? I think that's a little better, but leave it up to you). Add a one-line comment stating the contract (canonical root is the minimum tree index), and note that mj_dsuFind requires an activated tree (parent[tree] >= 0).
2. Comments: restore the two blocks you condensed in the flex-stiffness loop: the mj_flexCG / sleep-invariant rationale, and especially the cross-reference to engine_derivative's flexStiff_active / flexInterp_processed, which keeps two code sites in sync.
3. Error message: restore the explicit mjERROR("constraint %d is between two static bodies"). Routing that failure through dsuUnion's "self-incidence" message is a worse diagnostic.
4. Style: this file uses lowercase comments ("union always links...", "flex stiffness couples...").
5. Add a one-line changelog entry to doc/changelog.rst, you can credit yourself 🙂 

Optionally, the PR description could be much shorter. The memory-scaling table and the bounded-arena test are the whole argument. But I don't care much about that.

teerthsharma:
I have tried to follow the instructions perfectly hoping we can proceed to merge

yuvaltassa (approved):
will pull this in, perhaps with some minot edits shortly. thanks!

teerthsharma:
> will pull this in, perhaps with some minot edits shortly. thanks!

should I now work on the warp implementation of same?

yuvaltassa:
mmm copybara is not happy, I think you need to rebase to head?

teerthsharma:
> mmm copybara is not happy, I think you need to rebase to head?

hope this time the Copybara doesnt get angry

yuvaltassa (approved):

teerthsharma:
should I make a fresh PR with shorter commit using commit290513 with shorter PR description as well cause Copybara import errors are not visible to me

yuvaltassa:
sorry lots of change today, can you try again?

yuvaltassa:
I can edit the commit description, don't worry about it, we just need to make copybara succeed once

teerthsharma:
> I can edit the commit description, don't worry about it, we just need to make capybara succeed once

I have no idea how the capybara works :/ what should I do now?

yuvaltassa:
complicated and not interesting, don't worry about it, some sort of code import bot, for our purposes

yuvaltassa:
just rebase on head

teerthsharma:
> just rebase on head

On it boss 🦭

yuvaltassa (approved):

teerthsharma:
it went through !! anything left which I can help to make sure this merges?

yuvaltassa:
All good. Note I needed to make some minor fixes. You would have caught this if you ran the tests under ASAN.

Do you understand what this fixes?

teerthsharma:
> All good. Note I needed to make some minor fixes. You would have caught this if you ran the tests under ASAN.
> 
> Do you understand what this fixes?

Yes I do and I won't repeat the mistake again a little rusty with C/cpp

teerthsharma:
@yuvaltassa thanks for the feedback time and effort it was fun collaborating 🦭

PR #3403, opened 2026-07-14
docs: add full keyboard and mouse shortcut reference for simulate

### LINKED ISSUE #3306: Docs: Adding full keyboard shortcut list for Simulate

Hello there. AFAIK, Mujoco Simulate doesn't yet have a full shortcut list anywhere.

### Alternatives

I would like to propose that we add the full shortcut list in Mujoco Docs while keeping what was developed in https://github.com/google-deepmind/mujoco/issues/735.

### Additional context

I have already collected (a somewhat) initial list of shortcuts and its description.

[simulate_shortcuts.md](https://github.com/user-attachments/files/28501142/simulate_shortcuts.md)

rkothari3:
Fixes #3306. Thanks to @scotgopal for the initial list in the issue.

## Summary
- Add a complete shortcut reference to the `simulate` section of the docs: simulation/camera, panels/files, visibility groups, visualization flags, rendering flags, and mouse. The `F1` overlay only covers the common commands.
- Tables are derived from the shortcut declarations in `simulate.cc`, `simulate.h`, `mjVISSTRING` and `mjRNDSTRING` rather than from observed behavior, so the less discoverable ones are included (`F1`-`F5`, `Backspace`, `Ctrl+L`, `Ctrl+A`, `Ctrl+C` are declared in `simulate.h`; `F6`/`F7` are absent from the overlay).
- Document speed-up as `=`, not `+`. The handler skips the key when Shift is held (`!state->shift`), so `+` (Shift+`=` on a US layout) does nothing, despite what the overlay shows.

## Validation
- `sphinx-build` of `doc/`: builds clean, no new warnings vs `main` (the two `autodoc` warnings for `mujoco_warp`/`mjx` are pre-existing).
- pre-commit passes.

## Notes
- `Segment` (`mjRNDSTRING`) declares `,`, but so does `Activation` (`mjVISSTRING`). Both are added to `ui0` with the vis flags first, so `matchshortcut` finds `Activation` and `Segment`'s shortcut never fires. The docs describe current behavior and omit `Segment`. Happy to file a separate issue or send a fix if you want one.

### DIFF
--- a/doc/programming/samples.rst
+++ b/doc/programming/samples.rst
@@ -122,7 +122,253 @@ Interaction is done with the mouse; built-in help with a summary of available co
 ``F1`` key. Briefly, an object is selected by left-double-click. The user can then apply forces and torques on the
 selected object by holding Ctrl and dragging the mouse. Dragging the mouse alone (without Ctrl) moves the camera. There
 are keyboard shortcuts for pausing the simulation, resetting, and re-loading the model file. The latter functionality is
-very useful while editing the model in an XML editor.
+very useful while editing the model in an XML editor. The complete set of shortcuts is given in
+:ref:`saSimulateShortcuts` below.
+
+.. _saSimulateShortcuts:
+
+simulate shortcuts
+^^^^^^^^^^^^^^^^^^
+
+The ``F1`` help overlay lists the most common commands. The tables below are the complete reference, and also apply to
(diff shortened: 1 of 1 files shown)


yuvaltassa:
I like this with the exception of `=`. Has anyone actually ever been confused by this?

yuvaltassa (approved):
Okay I'll pull this in and possibly make minor changes while I'm at it

PR #3407, opened 2026-07-16
Resolve OpenGL backend lazily

devshahofficial:
Draft follow-up for #3366, based on the original lazy-backend import fix from @kevinzakka.

This keeps `MUJOCO_GL` validation at import time but defers importing the selected backend until a `GLContext` is constructed. That lets `import mujoco` work on systems that never render and do not have OpenGL libraries available.

Local validation:
- `python3 -m py_compile python/mujoco/rendering/classic/gl_context.py python/mujoco/gl_context_test.py`
- direct importlib/mocked validation for backend delegation, native `AttributeError` conversion, and backend `ImportError` conversion
- `git diff --check`

Could not run `python/mujoco/gl_context_test.py` directly in this local Python because `absl-py` is not installed in the base environment.

Opening as draft to avoid duplicating the existing approved PR unless maintainers want this rebased version.

### DIFF
--- a/python/mujoco/gl_context_test.py
+++ b/python/mujoco/gl_context_test.py
@@ -1,0 +1,66 @@
+# Copyright 2025 DeepMind Technologies Limited
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


devshahofficial:
Quick CI read: the lone macOS ARM failure is in CMake/GoogleTest discovery for the unrelated `engine_inverse_test` (empty JSON while linking); the OpenGL diff is not on that path, and the rest of the matrix is green. I tried rerunning the failed job, but fork authors do not have permission on the upstream workflow. Could a maintainer kick that job once? If it reproduces, I’ll dig into it.

PR #3415, opened 2026-07-20
Fix duplicated word in simulation.rst

Atishyy27:
Fixes a duplicated word in the documentation.

In `doc/programming/simulation.rst`, in the `fwdinv` / inverse-dynamics discussion:

> Recall that that the inverse dynamics use analytical formulas and are always exact...

"that that" -> "that". Documentation-only, no code changes.


kevinzakka:
Easier to fix internally, thanks for raising though.

PR #3418, opened 2026-07-22
Fix least_squares result when xtol is reached

winklemad:
## Summary

- accept the Armijo-approved candidate before terminating on `xtol`
- add a regression test ensuring `least_squares` returns that accepted candidate

## Problem

When an accepted step satisfies the relative step-size tolerance, `least_squares` sets `DX_TOL` and exits before assigning `xnew` and `rnew`. It therefore returns the previous, higher-objective iterate and records that stale iterate in the final trace.

## Testing

- focused `test_xtol_returns_accepted_candidate`: passed
- complete `python/mujoco/minimize_test.py` module: 16 tests passed
- `trailing-whitespace`: passed
- `end-of-file-fixer`: passed
- `git diff --check`: passed

### DIFF
--- a/python/mujoco/minimize.py
+++ b/python/mujoco/minimize.py
@@ -416,6 +416,8 @@ def least_squares(
 
     # Check termination condition on step norm.
     if dx_norm < xtol * (xtol + np.linalg.norm(x)):
+      x = xnew
+      r = rnew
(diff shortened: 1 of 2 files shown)


kevinzakka (approved):
Thanks for catching this!

PR #3420, opened 2026-07-22
Add an IPC contact mode to the discrete integrator

smallquail:
## What

A new opt-in flag, `<option integrator="discrete"><flag ipc="enable"/></option>`, selecting a contact mode of the `discrete` integrator that provides **penetration-free contact for deformable flexes**: flex–flex contact including **self-collision** (vertex–triangle and edge–edge) and flex-vs-static-geom. All other constraints — rigid contacts, limits, friction, equality — remain on the native pipeline, solved simultaneously. Models without the flag are unaffected; the flag is an error with any other integrator.

## How it works

IPC methods prevent penetration by combining an implicit position-level step with continuous collision detection (CCD). This implementation makes that work *inside* the discrete integrator rather than beside it:

1. **Per-step implicit minimization.** Each step minimizes an incremental potential over the flex vertex positions (and coupled joint dofs). Flex elasticity is carried by the effective metric (elastic2d) or by the native edge-equality rows.

2. **Augmented-Lagrangian contact (why not just a barrier).** Classic IPC enforces non-penetration with a stiff log-barrier, which brings severe ill-conditioning and barrier-parameter tuning. Here each contact pair instead carries a persistent Lagrange multiplier — a running contact-force estimate — and a few outer iterations per step alternate between solving the contact-augmented problem and updating the multipliers. Forces converge to the right values at moderate, fixed stiffness: no barrier, no conditioning cliff.

3. **The discrete solve as the inner solver.** Each outer iteration's subproblem is one discrete constraint solve (the CG solver on the standard primal objective). The contact pairs are published as the metric's contact class, one packed row per pair, so the operator, the solver's allocation pass and the preconditioner see them exactly as they see passive flex contact. There is no separate optimizer. Passive flex contact is the same law at zero multiplier, so the two share one implementation and one row layout rather than defining contact twice.

4. **CCD-bounded commits.** Candidate pairs are gathered with a conservative motion margin; every accepted position update advances only along a verified intersection-free path, and a step commits only once its full motion has been absorbed — so the committed trajectory never tunnels, and a step that cannot complete warns rather than silently losing motion.

## Scope and limitations

- Dim-2 flexes, with either edge-equality or elastic2d elasticity.
- IPC ownership covers flex–flex (incl. self-collision) and flex-vs-static-geom; flex-vs-moving-rigid and rigid–rigid contacts remain native.
- Rigid bodies are carried through the same position-level step. Without flex contact the step reproduces the discrete integrator to round-off (`IpcTest.RigidNoContactMatchesDiscrete`); a model with no dim-2 flex is stepped by the mode as well.
- Also fixed here: passive flex contact under `jacobian="dense"` crashed in the velocity shift, because the Jacobian chain is only filled for sparse models; dense rows are now compacted on their nonzeros.
- For flex models the constraint stage of `mj_forward` is skipped: `mjData.qacc` holds the free-flight acceleration and acceleration-stage sensors do not include contact until the step completes. Inverse dynamics is not supported under the flag.

## Testing

`engine_ipc_test` covers the distance kernels (point–triangle, segment–segment, vertex–geom), pinned-vertex handling, and contact scenarios; exercised on cloth/bag models with heavy self-contact.

`engine_forward_test` adds two passive-flex-contact regressions: a sheet pinned to a hinged body, which the metric-row Jacobian chain keeps stable, and a dense-Jacobian model, which used to crash.


yuvaltassa (commented):
Reviewed locally at 9489d8f0a: full test suite green, ASAN clean on the IPC and passive-contact tests, and a two-mjData isolation + reset-reproducibility probe comes back exactly zero — the mjData migration of the AL state holds up. The revision reads very well: the metric contact class as the channel, `mj_fwdConstraintCG` instead of mutating `m->opt`, `mj_advance` as the shared tail, the step-map `qacc`, and the passive-force unification onto the published rows (all five walkers of the extended row format check out). Comments inline: a few remaining gaps (silent partial commit on the outer-cap exit, fwdinv load gate, stack budget at bag-scene candmax), some scale-dependence and portability points, and doc/changelog wording.

Three notes on the already-landed collision module (b924ac66b), out of this diff's range but noticed while reviewing against it: (1) the three broad-phase buffers in mjc_candidates (`stk`/`outel`/`stampG`) are malloc/freed on every call, once per outer iteration — the mjData stack fits the per-call lifetime; (2) in addCand, missing blank line before the "KEEP penetrating" and "closing-bound prune" comments, and the orphaned "unopposed;" continuation reads as a wrapping accident; (3) bvhBox takes mjData but uses only `d->bvh_aabb_dyn` — passing that pointer makes the dependence explicit.

src/engine/engine_ipc.c
```diff
@@ -0,0 +1,1032 @@
+// Copyright 2021 DeepMind Technologies Limited
```
yuvaltassa: Copyright 2021 (the test file and the landed collision module both say 2026). The comment-hygiene pass landed nicely — one straggler: `IPC_ASET_TOI` (L48) is still referenced by nothing but its own comment.

src/engine/engine_ipc.c
```diff
@@ -0,0 +1,1032 @@
...
+#define IPC_NPT 8                    // max vertices in one pair: point-triangle 1+3, edge-edge 2+2
+#define IPC_CDAMP_FLEX 0.1           // normal dashpot coefficient (cde = IPC_CDAMP_FLEX*m/h^2)
+#define IPC_ASET_AGE 25              // drop a never-engaged pair after this many quiet updates
+#define IPC_DECAY 0.3                // stiffness decay per quiet update: scale = IPC_DECAY^age * mu
+#define IPC_VEL_TOL 0.05             // step convergence (m/s): |dx|_inf <= IPC_VEL_TOL*h
+#define IPC_ALPHA_LB 1e-6            // advance xfree only when the CCD fraction exceeds this
+#define IPC_STALL_MAX 64             // outer iterations without a feasible advance: the step froze
+#define IPC_ASET_TOI 1e-6            // admit a candidate iff its CCD time of impact < 1 - this
+#define IPC_DELTACAP 0.001           // standoff cap (m)
+#define IPC_GHAT (3 * IPC_DELTACAP)  // detection band (m)
```
yuvaltassa: Replacing `IPC_MU_SCALE` with `mjFLEXCONTACT_OMEGA2` is the right move — one constant shared with the passive-contact law. The remaining three are still dimensional absolutes: `IPC_VEL_TOL` 0.05 m/s, `IPC_DELTACAP` 1 mm, `IPC_GHAT` 3 mm. A mm-scale or slow-motion model gets a fixed 3 mm detection band and a 50 mm/s convergence tolerance. Derive from model scale (`stat.meansize`, `opt.tolerance`), expose as options, or at minimum document the scale assumptions.

src/engine/engine_ipc.c
```diff
@@ -0,0 +1,1032 @@
...
+    if (isartictree[t]) {
+      aoff[t] = N + N_artic;
+      atid[na_artic++] = t;
+      N_artic += m->tree_dofnum[t];
+    } else
+      aoff[t] = -1;
+  }
+  int N_total = N + N_artic, Na = (N_total > 0 ? N_total : 1);  // full solver vector length
+  int nstate = nfv;
+  int candmax = npt * 192 + 8192;  // candidate capacity (feature-heavy bins reach ~160k at npt ~1100)
```
yuvaltassa: Stack budget: `cand` + `aset` + `amerge` are each candmax entries; at the bag scene's npt~1100 that is ~220k entries x ~140 B x 3 arrays ~ 90 MB of mjData stack, far beyond a default arena — the model needs an explicit large `memory` attribute and the failure mode is an mj_stackAlloc error mid-step. Worth documenting the requirement (the flag's doc section would fit), or deriving candmax from the model rather than a flat per-vertex multiplier.

src/engine/engine_ipc.c
```diff
@@ -0,0 +1,1032 @@
...
+    // converged. A partial advance commits only part of the motion while time still advances by h,
+    // which reads as slow motion; partial advances are fine within the loop, and a frozen CCD
+    // leaves through the stall path below
+    if (beta >= 1.0 - 1e-6 && (last_ls_alpha >= 1.0 - 1e-9 || newton_converged_out))
+      break;
+    if (ac > IPC_ALPHA_LB)
+      stall = 0;
+    else if (++stall >= IPC_STALL_MAX) {
+      stalled = 1;
+      break;
```
yuvaltassa, line 964: The stall warning covers the frozen-CCD exit, but a slow grind — `ac` above the lower bound every iteration yet beta still short of 1 when outer_cap runs out — exits silently and commits the partial blend. Since the commit-completeness invariant is beta ~ 1, suggest warning on `beta < 1-1e-6` at exit regardless of path; the stall message's "advanced %.3f of the motion" phrasing already fits.

src/engine/engine_forward.c
```diff
@@ -1671,6 +1694,12 @@ void mj_RungeKutta(const mjModel* m, mjData* d, int N) {
 
 // runtime option validation for the discrete integrator
 void mj_checkDiscrete(const mjModel* m) {
+  // the IPC contact mode is a mode of the discrete integrator: its inner solves are the
+  // discrete solve, so no other integrator can host it
+  if (mjENABLED(mjENBL_IPC) && m->opt.integrator != mjINT_DISCRETE) {
+    mjERROR("flag ipc requires integrator='discrete'");
+  }
```
yuvaltassa, line 1702: One more gate worth adding here: `mjENBL_FWDINV` with flag ipc survives load and then hits the mjERROR in mj_discreteAcc mid-step, through mj_compareFwdInv in mj_step. Rejecting the combination at validation keeps the failure at load time like the others in this function.

src/engine/engine_ipc.c
```diff
@@ -0,0 +1,1032 @@
...
+  }
+  if (mmin >= 1e29) mmin = 1e-9;
+  return mjFLEXCONTACT_OMEGA2 * mmin;
+}
+
+// cnt -> decay exponent c (AL normal-contact aging): c = cnt>=0 ? cnt : max(-cnt-6, 0).
+static inline int ipc_cntExp(int cnt) { return cnt >= 0 ? cnt : (-cnt - 6 > 0 ? -cnt - 6 : 0); }
+
+// stable per-pair hash for the persistent cnt store: contact type + sorted vertex/feature indices.
+static unsigned long ipc_pairHash(const mjcFlexPair* con) {
```
yuvaltassa: `unsigned long` is 32-bit on Windows (LLP64), halving the pair-hash width there — more collisions, and a collision silently shares warm-start state between two pairs. `uint64_t`. Related convention nit: raw `sqrt()` where the engine uses `mju_sqrt` (matters under mjUSESINGLE).

src/engine/engine_ipc.c
```diff
@@ -0,0 +1,1032 @@
...
+  int idx[4];                  // == mjcFlexPair.idx
+  int gi;                      // == mjcFlexPair.g
+  mjtNum lam;                  // AL multiplier
+  mjtNum ld0, ln[3], lcw[4];   // per-outer linearization at xfree
+  int liv[4], lniv;
+  int cnt;                     // active-set state machine
+  mjtNum s;                    // materialized AL slack
+} ipcCon;
+#define IPC_PAIR(con) ((const mjcFlexPair*)(con))
+
```
yuvaltassa, line 78: The by-cast layout pun on mjcFlexPair is fine, but nothing enforces the ==-comments' contract — a reorder of either struct compiles clean and corrupts silently. A few `_Static_assert`s (`offsetof(ipcCon, type/idx/gi) == offsetof(mjcFlexPair, type/idx/g)`) make it self-checking.

src/engine/engine_ipc.c
```diff
@@ -0,0 +1,1032 @@
...
+  int nfree_flex = nfree;  // flex free-DOF count
+  // articulated trees: every tree with dofs that is not a flex vertex, appended to the solver
+  // vector after the flex block at offset aoff[tree]
+  int ntree = m->ntree, na_artic = 0;
+  char* isartictree = (char*)mj_stackAllocByte(d, (ntree > 0 ? ntree : 1), 1);
+  for (int t=0; t < ntree; t++)
+    isartictree[t] = 0;
+  for (int b=1; b < m->nbody; b++) {
+    if (m->body_dofnum[b] == 0 || isflexvert[b]) continue;
+    isartictree[m->dof_treeid[m->body_dofadr[b]]] = 1;
```
yuvaltassa, line 524: Question: the articulated complement includes SLEEPING trees — they get appended into the solve, and the commit writes their qvel/qpos (values unchanged when nothing pushes them, but written on frozen dofs). Under sleep+ipc on a rigid-only model this is reachable (mj_checkDiscrete only fences sleep+flex). Exclude sleeping trees from the append (matching the constraint filter), or fence sleep+ipc for now?

doc/XMLreference.rst
```diff
@@ -757,6 +757,25 @@ from its default.
...
+   with any other integrator. It applies model-wide: every flex the mode supports has its contact solved this way.
+   Contact is passive under this flag whatever :ref:`passive<flex-contact-passive>` says, since the flag replaces
+   the penalty form of passive contact — the same contact law with the multiplier held at zero — with the
+   augmented-Lagrangian solve, rather than returning any flex to the constraint solver. Flex contact is solved by a
+   barrier-free augmented-Lagrangian outer loop around the
+   discrete solve: each step minimizes an incremental potential subject to linearized contact constraints,
+   re-linearizing contact at trial positions, and every committed position update is verified intersection-free
+   by continuous collision detection, so flex contact cannot tunnel. Rigid bodies are carried through the same
+   position-level step with their contacts kept in the constraint solver, and a model without 2D flexes takes that
+   step as well. Supported for dim-2 flexes; edge-equality elasticity rides the native constraint rows and
```
yuvaltassa: This sentence is unclear to me: "edge-equality elasticity rides the native constraint rows and elastic2d elasticity the effective metric." (Also appears verbatim in the changelog entry.) Possible rewording: "a flex with edge equality constraints keeps its elasticity in the constraint solver, while elastic2d elasticity is integrated implicitly through the effective metric."

doc/changelog.rst
```diff
@@ -33,6 +33,18 @@ Engine
      elasticity or passive flex contact under ``implicit``/``implicitfast`` now raise a runtime error carrying this
      migration note.
 
+- Added the ``ipc`` :ref:`flag<option-flag-ipc>`, a prototype contact mode of the ``discrete`` integrator for
+  penetration-free flex contact. Each step minimizes an incremental potential subject to linearized contact
+  constraints, using a barrier-free augmented Lagrangian whose subproblems are the discrete solve. Every committed position update is
```
yuvaltassa: Bad line wrapping to 120 (this line is 133 chars).

doc/changelog.rst
```diff
@@ -33,6 +33,18 @@ Engine
...
+  verified intersection-free by continuous collision detection, so flex contact cannot tunnel. Supported for dim-2
+  flexes; edge-equality elasticity rides the native constraint rows and elastic2d elasticity the effective metric.
+
+  .. admonition:: Breaking API changes
+     :class: attention
+
+     :ref:`mjData` gained the ``flexvert_lambda`` and ``flexvert_conage`` fields (per-flex-vertex contact multipliers
+     and contact age, used by the ``ipc`` integrator), changing its size and layout.
```
yuvaltassa: ABI not API — and honestly I don't think we need this admonition at all: mjData field additions are routine and go unflagged (the discrete-integrator entry above added ~30 `efm_*` fields with none). Also "used by the ipc integrator" — it's a flag. Suggest deleting the admonition.

include/mujoco/mjdata.h
```diff
@@ -21,9 +21,6 @@
 #include <mujoco/mjtype.h>
 #include <mujoco/mjmodel.h>
 
-
-
-
 //------------------------------------- Contact ----------------------------------------------------
```
yuvaltassa, line 27: Lots of unnecessary blank-line deletions in this file — formatting churn in regions the PR doesn't otherwise touch; please restore.

include/mujoco/mjdata.h
```diff
@@ -243,6 +235,11 @@ typedef struct mjData_ {
   mjtNum* flexvert_length;   // flex vertex lengths                              (nflexvert x 2)
   mjtNum* bvh_aabb_dyn;      // global bounding box (center, size)               (nbvhdynamic x 6)
 
+  // computed by mj_IPC (flag ipc): recoverable from qpos/qvel, but carried across steps so
+  // the augmented-Lagrangian contact solve resumes from the previous multipliers
```
yuvaltassa: Comment is wrong and too long — "recoverable from qpos/qvel" is exactly what the multipliers are not: they are dual state, which is why they must be carried across steps. Something like "AL contact state carried across steps (flag ipc)" on the section line.

include/mujoco/mjdata.h
```diff
@@ -323,7 +320,8 @@ typedef struct mjData_ {
   mjtNum* qacc_smooth;       // unconstrained acceleration                       (nv x 1)
 
   // computed by mj_fwdConstraint/mj_inverse
-  mjtNum* qfrc_constraint;   // constraint force                                 (nv x 1)
+  mjtNum* qfrc_constraint;   // constraint force; under the IPC contact mode
+                             // this includes flex contact with no efc row       (nv x 1)
```
yuvaltassa: Fit in one line.

src/engine/engine_ipc.c
smallquail: done

src/engine/engine_ipc.c
smallquail: done

src/engine/engine_forward.c
smallquail: done

src/engine/engine_ipc.c
smallquail: done

src/engine/engine_ipc.c
smallquail: done

doc/XMLreference.rst
smallquail: done

doc/changelog.rst
smallquail: done

doc/changelog.rst
smallquail: done

include/mujoco/mjdata.h
smallquail: done

include/mujoco/mjdata.h
smallquail: done

include/mujoco/mjdata.h
smallquail: done

src/engine/engine_ipc.c
smallquail: Removed the fixed capacity. The broad phase now pushes each candidate onto the mjData arena one at a time, the way mj_collision pushes its pairs, so every query yields an exactly sized contiguous array and nothing is reserved up front. The active set is arena-allocated the same way: seeded at the candidate count, then each merge writes a fresh block bounded by the old set plus the admitted candidates. No per-vertex constant remains. If the arena fills, the broad phase warns and stops adding, and the active set raises the standard out-of-memory error, so the remedy is the model's `memory` attribute, as for contacts.

On the humanoid bag (1195 flex vertices) peak stack use per step drops from 69 MB to 2.8 MB, or 70 MB to 3.6 MB including the arena. Trajectories are bit-identical over 1000 steps on both bag scenes, and step time is unchanged (13.93 vs 13.90 ms and 27.92 vs 27.90 ms, min of six interleaved runs). Across a collection of 106 IPC scenes the candidate count peaks at 2.3 per flex vertex, so the old 192 was roughly 80x the need.

Two related edits in the same pass: the broad phase's three malloc'd buffers now come from the mjData stack, per your note on the collision module, and the first-query gap pass, whose values were overwritten before being read, is removed.

src/engine/engine_ipc.c
smallquail: Fenced for now: the sleep flag is rejected together with the ipc flag, next to the existing sleep fences, and the reference lists the combination. Excluding sleeping trees from the append would be only part of the job. Sleeping bodies are also invisible to the flex broad phase, which handles static geometry only, so cloth would pass through a sleeping box, and waking on flex contact depends on the island bookkeeping that the whole-problem inner solve bypasses. Those belong with the sleep-with-flex work in the discrete integrator, done together with a test. A test now covers all four of the flag's option fences.

src/engine/engine_ipc.c
smallquail: Documented for now: the three are absolute values tuned for metre-scale models with millimetre-thick flexes, stated at the defines and in the flag's reference entry. A TODO records the intended fix for the rest gap: compute it per flex pair from the radii, as the engine's flex contact does, instead of the hard-coded cap, with the detection band following from it.

src/engine/engine_ipc.c
```diff
@@ -0,0 +1,1057 @@
...
+    // advance xfree by the CCD fraction
+    if (ac > IPC_ALPHA_LB)
+      for (int i=0; i < 3 * npt; i++)
+        xfree[i] = (1.0 - ac) * xfree[i] + ac * x[i];
+    beta = beta + (1.0 - beta) * ac;
+    // terminate once the advance is complete and the full step was accepted or both blocks
+    // converged. A partial advance commits only part of the motion while time still advances by h,
+    // which reads as slow motion; partial advances are fine within the loop, and a frozen CCD
+    // leaves through the stall path below
+    if (beta >= 1.0 - 1e-6 && (last_ls_alpha >= 1.0 - 1e-9 || newton_converged_out))
```
yuvaltassa, line 1205: A collision-free update doesn’t establish AL convergence. A sheet approaching a plane at 50 m/s loses 99.5% of its speed while still 49.5 mm above it: the initial frozen slack resists approach, and this exit accepts the result. We need a convergence check independent of CCD acceptance.

src/engine/engine_ipc.c
```diff
@@ -0,0 +1,1057 @@
...
+    ipc_flexLamUpdate(x, xfree, rad, ghat, mass, aset, naset, d->flexvert_lambda, pt2vg, npt);
+    ipc_ageStep(d->flexvert_conage, aset, naset, pt2vg, npt, ageload);
+    // displacement x - xfree in the free-dof layout
+    for (int v=0; v < npt; v++)
+      if (fidx[v] >= 0)
+        for (int c=0; c < 3; c++)
+          dx[3 * fidx[v] + c] = x[3 * v + c] - xfree[3 * v + c];
+    // re-query over the swept segment with a fixed collar (3*ghat, no displacement scaling), which
+    // keeps the count bounded however far x flung
+    ncand = mjc_candidates(m, d, xfree, gv, ge, ngv, nge, rad, 3 * ghat, 3 * ghat, 0.0, xfree, x,
```
yuvaltassa: This query can discard crossing pairs before checking their swept motion. Passing a sheet from 5 cm above a plane to 5 cm below it returns zero candidates and `alpha=1`; expanding the bounds by its displacement detects the collision. Candidate generation needs conservative swept bounds.

src/engine/engine_collision_continuous.c
```diff
@@ -210,6 +211,12 @@ static void closestOnPoly(const mjtNum* p, const float* vbase, const int* pv, in
   }
 }
 
+// geom types mjc_GeomDist supports; a flex vertex has no continuous contact with the others
+int mjc_GeomSupported(int type) {
+  return type == mjGEOM_PLANE || type == mjGEOM_SPHERE || type == mjGEOM_CAPSULE ||
```
yuvaltassa, line 217: Vertex-to-geom tests miss a sphere crossing a triangle’s interior. A local `mj_step` reproduction lets a constrained cloth sheet pass completely through a sphere. We need face/edge coverage before treating these geoms as supported.

src/engine/engine_ipc.c
```diff
@@ -0,0 +1,1057 @@
...
+  }
+  int nfree_flex = nfree;  // flex free-DOF count
+  // articulated trees: every tree with dofs that is not a flex vertex, appended to the solver
+  // vector after the flex block at offset aoff[tree]
+  int ntree = m->ntree, na_artic = 0;
+  char* isartictree = (char*)mj_stackAllocByte(d, (ntree > 0 ? ntree : 1), 1);
+  for (int t=0; t < ntree; t++)
+    isartictree[t] = 0;
+  for (int b=1; b < m->nbody; b++) {
+    if (m->body_dofnum[b] == 0 || isflexvert[b]) continue;
```
yuvaltassa, line 696: A body carrying pinned flex vertices is still an articulated body, but this classification excludes it from integration. A hinged body with all cloth vertices pinned to it retains nonzero velocity while its position stays unchanged. Pinned points need to follow the parent’s kinematics.

src/engine/engine_ipc.c
```diff
@@ -0,0 +1,1057 @@
...
+      if (L > maxdisp) maxdisp = L;
+    }
+  mjtNum thresh = 3 * ghat + 4 * maxdisp;
+  // static features move on the flex side only, so their margin is half the two-sided one; the
+  // full margin on a many-piece convex bin produces hundreds of thousands of candidates
+  mjtNum threshGeom = ghat + 2.0 * maxdisp;
+  // the candidates live on the arena, pushed one at a time by the broad phase (mjc_candidates),
+  // so every query yields an exactly sized array with nothing reserved up front
+  mjcFlexPair* cand;
+  int ncand = mjc_candidates(m, d, x, gv, ge, ngv, nge, rad, thresh, threshGeom, maxdisp, xold,
```
yuvaltassa: IPC still produces contact forces with `contact="disable"` or with the flex’s collision masks set to zero. Both force assembly and CCD should respect the usual collision filtering.

src/engine/engine_core_constraint.c
```diff
@@ -2671,6 +2672,25 @@ static int mj_nc(const mjModel* m, mjData* d, int* nnz) {
...
+    // flex, and against the other geom types stay here
+    if (mjENABLED(mjENBL_IPC) && !con->exclude) {
+      int f0 = con->flex[0], f1 = con->flex[1];
+      int ipc = (f0 > -1 && m->flex_dim[f0] == 2) && (f1 > -1 && m->flex_dim[f1] == 2);
+      for (int s = 0; s < 2 && !ipc; s++) {
+        int fo = con->flex[1-s], g = con->geom[s];
+        if (con->flex[s] < 0 && g > -1 && fo > -1 && m->flex_dim[fo] == 2) {
+          ipc = (m->body_weldid[m->geom_bodyid[g]] == 0) && mjc_GeomSupported(m->geom_type[g]);
+        }
+      }
```
yuvaltassa: Removing the native contact also removes its friction, while IPC currently adds only normal forces. Is frictionless contact the intended scope? A sheet sliding at 1 m/s retains that speed with friction set to 1; supporting manipulation will require tangential forces coupled to the IPC normal load.

include/mujoco/mjdata.h
```diff
@@ -243,6 +243,10 @@ typedef struct mjData_ {
   mjtNum* flexvert_length;   // flex vertex lengths                              (nflexvert x 2)
   mjtNum* bvh_aabb_dyn;      // global bounding box (center, size)               (nbvhdynamic x 6)
 
+  // AL contact state carried across steps (flag ipc)
+  mjtNum* flexvert_lambda;   // flex contact multiplier                          (nflexvert x 1)
```
yuvaltassa, line 248: These persistent arrays are missing from `mjSTATE_INTEGRATION`. Restoring a loaded-contact state into fresh `mjData` changes the next trajectory, whereas `mj_copyData` reproduces it exactly. Please include both arrays in state save/restore.

src/engine/engine_ipc.c
```diff
@@ -0,0 +1,1057 @@
...
+                                                                            : 1);
+          for (int k=0; k < nqj; k++)
+            d->qpos[qa + k] = qcur_a[qa + k];
+        }
+      for (int i=0; i < nd; i++)
+        d->qvel[da + i] = qdelta[o + i - N] / h;
+    }
+  }
+  // publish the step's effective acceleration, as an integrator hands it to mj_advance
+  mju_copy(d->qacc, qacc_out, m->nv);
```
yuvaltassa: The committed acceleration includes IPC forces, but `qfrc_constraint` omits them. In a loaded-contact test, `M*qacc - qfrc_smooth - qfrc_constraint` equals the missing normal load. The reported forces should account for the accepted IPC response.

src/engine/engine_ipc.c
```diff
@@ -0,0 +1,1057 @@
...
+      for (int k=0; k < 3; k++)
+        dn += acon[c].lcw[p] * acon[c].ln[k] * (x[3 * v + k] - xold[3 * v + k]);
+    }
+    if (dd <= 0 && dn < 0) {
+      mjtNum mmin = 1e30;
+      for (int p=0; p < acon[c].lniv; p++) {
+        mjtNum mv = mass[acon[c].liv[p]];
+        if (mv < mmin) mmin = mv;
+      }
+      E += 0.5 * (IPC_CDAMP_FLEX * mmin / (h * h)) * dn * dn;
```
yuvaltassa: The merit includes a normal dashpot, but the inner solve appears to include neither its force nor its curvature. That makes the solve inconsistent with the objective used to accept it. Please add the matching terms or remove the dashpot from the merit.

test/engine/engine_ipc_test.cc
```diff
@@ -0,0 +1,844 @@
...
+  EXPECT_NEAR(PtTri(p_on, a, b, c), 0.0, 1e-12);
+}
+
+// segment-segment distance: perpendicular crossing, collinear gap, parallel
+// offset
+TEST_F(IpcTest, SegmentSegmentDistance) {
+  mjtNum p1[3] = {-1, 0, 0}, p2[3] = {1, 0, 0};
+
+  mjtNum q1[3] = {0, -1, 0.3}, q2[3] = {0, 1, 0.3};  // perpendicular, 0.3 above
+  EXPECT_NEAR(SegSeg(p1, p2, q1, q2), 0.3, 1e-12);
```
yuvaltassa: Seven of the 17 IPC tests fail with `mjUSESINGLE`. Some assertions require double-precision accuracy from float results, including this `1e-12` tolerance. Please use precision-appropriate tolerances and check the dynamics failures separately.

src/engine/engine_ipc.c
smallquail: done

src/engine/engine_ipc.c
smallquail: done

test/engine/engine_ipc_test.cc
smallquail: done

src/engine/engine_ipc.c
smallquail: done

src/engine/engine_ipc.c
smallquail: done

src/engine/engine_collision_continuous.c
smallquail: done

src/engine/engine_ipc.c
smallquail: done

src/engine/engine_core_constraint.c
smallquail: Frictionless is the intended scope of this PR: the mode carries normal forces only, and rigid contacts keep their native rows and friction. Tangential forces coupled to the IPC normal load are planned as a follow-up.

include/mujoco/mjdata.h
smallquail: Agreed, both arrays belong in the integration state. Two ways to carry them: extend `mjSTATE_WARMSTART` to include them (they are the contact solver's warm start; the component's size becomes nv + 2*nflexvert, unchanged for models without flex vertices, and no new enum value), or add a dedicated `mjtState` bit and include it in `mjSTATE_INTEGRATION`. Do you have a preference?

src/engine/engine_ipc.c
smallquail: The frozen slack that produced this was removed in 6dd69617a: the rows' wall now sits at λ/μ, so an unloaded pair exerts no force until the standoff. In this example the sheet now crosses the gap in one step and rests at the 1 mm standoff (49.5 mm and 99.5% of the speed lost before). What the exit accepts is a full step of the subproblem: solved to tolerance, a full line-search step on the merit, and the whole motion accepted by the CCD; the multipliers are updated once per step and converge across steps through the warm start.

include/mujoco/mjdata.h
yuvaltassa: Let’s leave state serialization out of scope for now and document IPC as experimental. Please explicitly note that `mj_getState`/`mj_setState` don’t capture its full state, so exact replay isn’t supported. I’d prefer to settle the design before extending the public state API.

src/engine/engine_collision_continuous.c
```diff
@@ -826,15 +933,20 @@ int mjc_candidates(const mjModel* m, const mjData* d, const mjtNum* x, const mjt
     }
     for (int v=0; v < nfv; v++) {  // flex verts only (rigid bodies carry no hard barrier)
       if (fidx[v] < 0) continue;
```
yuvaltassa: This still skips pinned vertices whose parent moves. A fully pinned paddle on a hinge passes through a plane because these candidates are missing and the native contacts are excluded. Moving pinned vertices need to participate here.

src/engine/engine_ipc.c
```diff
@@ -0,0 +1,1385 @@
...
+        a_gdq[i] = 0;
+      for (int a=0; a < na_artic; a++) {
+        int t = atid[a], o = aoff[t], da = m->tree_dofadr[t], nd = m->tree_dofnum[t];
+        for (int i=0; i < nd; i++)
+          a_gdq[da + i] = qfree[o + i - N];
+      }
+      mju_copy(d->qpos, qn_a, m->nq);
+      mj_integratePos(m, d->qpos, a_gdq, 1);
+      mj_kinematics(m, d);
+      ipc_pinnedPos(m, d, npt, pjnv, pt2vg, xfree);
```
yuvaltassa: CCD checks straight vertex paths, but the FK update moves pinned vertices along rotation arcs. A paddle can swing through a sphere with both endpoints clear. We need to check the actual rotational path, or explicitly restrict moving attachments for now.

src/engine/engine_ipc.c
```diff
@@ -0,0 +1,1385 @@
...
+    int va_k = m->flex_vertadr[fi], nv_k = m->flex_vertnum[fi];
+    for (int lv=0; lv < nv_k; lv++) {
+      int v = fxadr[k] + lv, bid = m->flex_vertbodyid[va_k + lv];
+      dofadr[v] = -1;
+      qpadr[v] = -1;
+      fidx[v] = -1;
+      mass[v] = 0;
+      rad[v] = rk;
+      pbody[v] = bid;
+      if (m->body_dofnum[bid] == 3) {
```
yuvaltassa: Three DOFs don’t necessarily mean three slide joints. A ball-jointed parent carrying pinned cloth takes this path and fails to rotate despite nonzero angular velocity. Please check the joint types and route these parents through articulated integration.

src/engine/engine_collision_continuous.c
smallquail: Done. Points that ride a moving body now take part in the geom candidates like free points, with their travel added to the reach; only points pinned to static bodies stay out. Their pairs also needed a stiffness scale: a pinned vertex carries no mass, so with the old zero the rows could not push and the CCD stalled at the plane. The scale is now the carrier's subtree mass shared among the points it carries. New test PinnedPaddleStopsOnPlane: a fully pinned paddle on a free vertical slide falls 20 cm onto the floor and rests on it, never below the plane. (A hinged carrier is now refused, see the next thread.)

src/engine/engine_ipc.c
smallquail: Restricted, as you suggested. A pinned vertex may ride a static body or a body whose chain has slide joints only, so every pinned point moves on the straight segment the CCD sweeps and its point Jacobian is constant over the step. A hinge, ball or free joint on the chain is refused when the model is stepped, with a message naming the flex, the body and the joint; since the compiler steps once, this surfaces at load. Documented in the flag's paragraph. The paddle tests now use a sprung vertical slide and still show the free cloth's weight reaching the carrier.

src/engine/engine_ipc.c
smallquail: Done. A free vertex is one whose body has exactly three slide joints; any other body holding a vertex is a pinned attachment and takes the articulated path, where the rule above applies. Test PinnedCarrierMustTranslate covers a hinge and a ball carrier; the ball case is the one that was taken for a vertex body.

include/mujoco/mjdata.h
smallquail: Done. The flag's documentation now states that the mode is experimental and that mj_getState/mj_setState do not capture the contact multipliers it keeps across steps, so exact replay from a saved state is not supported; the same note sits on the mjData fields and in the changelog.

src/engine/engine_core_constraint.c
smallquail: Friction is now in (ed16c175f). Each pair the mode resolves is published as the four facets of the pyramidal cone, `J_n ± μ J_t` over the pair's columns, with the normal row's reference plus or minus μ times a tangential one that asks for no tangential slip over the step. μ is the larger of the two sides' sliding friction, as for geom pairs; torsional and rolling friction are not applied, and zero gives the frictionless row. The facets are unilateral rows like the normal one, so the solver handles them as any pyramidal contact and no mjContact stands behind them.

The example above now holds: a cloth on a box tilted by 30° sticks with friction 1 and slides at g(sin a − μ cos a) with 0.3 and 0, within 15% (IpcTest.ClothOnInclineObeysCoulomb). The cost is four rows per pair: +37–39% on the knot and the two humanoid-bag scenes.

Two notes. The self-contact momentum test's CG cap is raised to 200: the impact step needs about 110 iterations with the facets under mjUSESINGLE, and a solve truncated at the cap commits a direction whose momentum error the test measures. And the line-search merit scores the normal walls at live gaps but not the facets, so it does not see a direction's tangential cost; a shared evaluator of the inner objective is the planned follow-up.

smallquail: Follow-up in 88ae22c1c: the line-search merit is now the solve's own objective at the trial state, evaluated once through the solver's machinery: the Gauss term over every dof in the effective metric with qacc_smooth as its centre, the native rows' cost from mj_constraintUpdate, and the contact cost of the full active set, the one-sided AL quadratic of each pair over its linearized gap at the committed position (the paper's form, not a live gap as I wrote above) and, for a pair with friction, its four facets. The facets carry a quarter of the pair's stiffness each, so a sticking pair costs what its normal row would and the multiplier update keeps its force law.

That replaced the flex inertia, articulated inertia, damping and metric-stiffness reconstructions the merit was assembled from. A probe against the old merit showed agreement to round-off on pure flex and on articulated joints with stiffness, and a 27% difference on a damped slide, where the old articulated residual's implicit-damping form carried the wrong linear term; the objective the CG minimizes is the one kept. Net effect on the knot pull: mean CG iterations 293 to 174, 17% less wall time; the two bag scenes within ±16% on diverged trajectories. The Windows failure (M_SQRT1_2 in the continuous-collision test) is fixed in the same commit.

src/engine/engine_ipc.c
```diff
@@ -0,0 +1,1508 @@
...
+      dofadr[v] = -1;
+      qpadr[v] = -1;
+      fidx[v] = -1;
+      mass[v] = 0;
+      rad[v] = rk;
+      pbody[v] = bid;
+      // a free vertex is one whose body is the three slide joints a flex vertex body carries; any
+      // other body holding a vertex is a pinned attachment (a ball joint has three dofs too)
+      int ja = m->body_jntadr[bid];
+      int slides = m->body_jntnum[bid] == 3 && m->jnt_type[ja] == mjJNT_SLIDE &&
```
yuvaltassa, line 677: A three-slide parent is still not necessarily an independent flex vertex. A 3x3 cloth fully pinned to one body with XYZ slide joints passes this check nine times, so the readback adds the same displacement to its qpos once per vertex. With qvel_x=1, h=0.002 and gravity=0, native discrete advances 0.002 m while IPC advances 0.018 m; IPC still reports qvel_x=1. Both precisions reproduce it on 88ae22c1c, and this code is unchanged here. Shared carrier DOFs need to go through the articulated path once.

smallquail: Done: a body is a free vertex only when it is the vertex, three slide joints, exactly one flex vertex on it, sitting at its origin. Any other body carrying vertices is a pinned attachment and takes the articulated path once for all its points, so a cloth pinned to a body on slides shares that body's dofs and its displacement is applied once. Your scene is now IpcTest.SlideCarrierMovesOnce: with qvel_x = 1, h = 0.002 and no gravity the carrier advances 0.020 m in ten steps and the cloth with it, matching the discrete integrator to round-off in both precisions.

src/engine/engine_ipc.c
```diff
@@ -0,0 +1,1418 @@
...
+              refc -= con->lcw[p] * con->ln[k] * (xold[3 * v + k] + h * Rqv[k] - xfree[3 * v + k]);
+            }
+            np++;
+          }
+          // D = mu/h^2: the inertia is in acceleration units, 1/h^2 heavier than in position form
+          esnpt[npair] = np;
+          esD[npair] = D * ih2;
+          esref[npair] = refc;   // the wall: J*qacc = ref puts the gap at lam/mu
+          esxnum[npair] = nx - esxadr[npair];
+          // friction: the row becomes the four facets J_n +- muf*J_t of MuJoCo's pyramidal cone
```
yuvaltassa: [P1] The four-facet objective generates friction and lift on separated cloth. On 470a68b, a sheet 5 mm above a plane with gravity=0, vx=5 m/s, h=2 ms and friction=1 ends one step at vx=3.51485 m/s, vz=1.48515 m/s and z=7.9703 mm; native discrete and friction=0 preserve free flight. Both precisions reproduce it. A positive normal gap still pays when mu*|tangential displacement| exceeds the gap, and the solver can reduce that cost by lifting the cloth. The working-set fix made the objectives consistent, but contact activation and the friction law now need to be defined together. Please cover unloaded separated motion as well as loaded sliding; belonging to a friction cone alone does not establish the intended contact response.

src/engine/engine_forward.c
```diff
@@ -1857,6 +1904,29 @@ void mj_implicit(const mjModel* m, mjData* d) {
...
+  return 0;
+}
+
+
+// the constraint stage, unless the IPC mode owns it (ipcSkipConstraint): then qacc stays at the
+// free-flight acceleration and the integrator computes the step
+static void fwdConstraintStage(const mjModel* m, mjData* d) {
+  if (ipcSkipConstraint(m)) {
+    mju_copy(d->qacc, d->qacc_smooth, m->nv);
```
yuvaltassa, line 1956: [P2] This changes acceleration sensors across the whole model, even for unrelated native contacts, and mj_IPC never refreshes them after its solve. A resting rigid sphere with an accelerometer reads ~0 instead of 9.81 m/s^2 when a static, non-colliding dim-2 cloth is present; the native result and rigid-body motion are otherwise correct. Reproduced after 200 steps in both precisions. Can we preserve supported sensor outputs or explicitly reject unsupported acceleration sensors for this experimental mode? The documentation's 'until the step completes' currently implies recovery that does not happen.

smallquail: Done: with a 2D flex in the model the step recomputes the acceleration-stage sensors from its own acceleration and constraint force before it commits, at the pre-step state, so after mj_step they read as under the discrete integrator; a user or plugin sensor at that stage is evaluated twice per step, and the documentation now says so. Your scene is IpcTest.AccelerationSensorsSeeContact: a sphere resting on the floor beside a pinned, non-colliding cloth reads 9.81 on its accelerometer after 200 steps in both precisions, matching the native run to 1.3e-5 m/s².

src/engine/engine_ipc.c
smallquail: Friction is removed from this PR: the pairs the mode resolves are frictionless and carry normal forces only, and the documentation says so. Your example preserves free flight, since a separated pair has no row that a tangential displacement can activate. Friction returns in a follow-up where contact activation and the friction law are defined together; the two Coulomb tests go with it.

src/engine/engine_core_constraint.c
smallquail: Update: friction is withdrawn from this PR and moves to a follow-up. The mode is frictionless again, as the first reply in this thread stated, with the rest of the changes since (the shared objective, the working set) kept.

PR #3436, opened 2026-07-27
Fix rendering race condition

### LINKED ISSUE #3435: MJX Warp: race condition between `mjx.render` and `mjx.refit_bvh`
### Intro

Hi!

I am a MuJoCo user working on manipulation. I ran into what looks like a scheduling issue between `mjx.refit_bvh` and `mjx.render`.

(The investigation below was carried out with help from Claude Opus.)

### My setup

- `mujoco` and `mujoco-mjx` at `d6650f84e58bdeb2de5a6949959bf64e96aa38d5` (I also checked `main`, which is unchanged in the relevant code).
- `warp-lang==1.14.0`, `jax==0.10.2`, `mujoco==3.10.1`. - Reproduced on both `macos-26` (`arm64`) and `ubuntu-24.04` (`x86_64`).

`mjx.refit_bvh(m, d, rc)` mutates the BVH owned by the render context, and `mjx.render(m, d, rc)` reads that BVH back when it casts rays. I expected the ray cast to observe the bounding volumes that were just fit for the same frame, but what I see is that it may instead observe the ones belonging to a different (old) frame, in which case geometry falls outside the traversed volumes and the ray reports whatever is behind it.

I think the root cause is that the two calls do not exchange anything through the JAX graph, so nothing constrains their relative order:

- In [`mjx/mujoco/mjx/warp/bvh.py`](https://github.com/google-deepmind/mujoco/blob/e23b5042013a1f7219d2950bf913ac981627bc61/mjx/mujoco/mjx/warp/bvh.py), `_refit_bvh_jax_impl` declares `has_side_effect=True` and allocates a `dummy` token output, but then discards it (`out = jf(...)` is never used, and the function returns `d = d.tree_replace({})`, i.e. the inputs unchanged). So the returned `mjx.Data` carries no dependency on the refit having run.
- In [`mjx/mujoco/mjx/warp/render.py`](https://github.com/google-deepmind/mujoco/blob/e23b5042013a1f7219d2950bf913ac981627bc61/mjx/mujoco/mjx/warp/render.py), `_render_jax_impl` declares `has_side_effect=False`, so the ray cast is a pure custom call that XLA can schedule freely.

Inspecting the optimized HLO for the five back-to-back renders (see the reproduction below), no ray cast has any edge back to a refit, and the schedule places all five refits ahead of all five ray casts (the refits are the `s32[1]` token outputs, the ray casts the tuple ones):

```console
ffi_call.25    refit   ffi-operands=NONE  control-preds=NONE
ffi_call.30    refit   ffi-operands=NONE  control-preds=NONE
ffi_call.35    refit   ffi-operands=NONE  control-preds=NONE
ffi_call.40    refit   ffi-operands=NONE  control-preds=NONE
ffi_call.45    refit   ffi-operands=NONE  control-preds=NONE
ffi_call.26    raycast ffi-operands=NONE  control-preds=NONE
ffi_call.31    raycast ffi-operands=NONE  control-preds=NONE
ffi_call.36    raycast ffi-operands=NONE  control-preds=NONE
ffi_call.41    raycast ffi-operands=NONE  control-preds=NONE
ffi_call.46    raycast ffi-operands=NONE  control-preds=NONE
```

The whole module contains zero occurrences of `control-predecessors` and zero `after-all` tokens, so I believe the ordering above is whatever the scheduler happened to pick rather than anything that is being enforced. Four of the five frames then come back with the box missing entirely: they are bit-identical to a render of the same scene with the box moved out of the camera frustum, and the one frame that survives is the last one, whose pose is the one the final refit left in the BVH.

I also want to note that the recipe in [MJX Documentation](https://mujoco.readthedocs.io/en/stable/mjx.html#batch-rendering) is exactly the shape that has no ordering guarantee, since it puts both calls inside one `jax.jit`:

```python
@jax.jit
def render_fn(mx, d, rc_pytree):
    # 1. Update the BVH for the current scene state
    d = mjx.refit_bvh(mx, d, rc_pytree)

    # 2. Render all configured cameras
    pixels, _ = mjx.render(mx, d, rc_pytree)
    ...
```

This bug occurs when several `refit`/`render` pairs sit side-by-side at the top-level of a signgel `jax.jit` call.

I believe the existing tests do not catch the failing shape because every one of them dispatches the `refit` as its own executable, e.g. `dx_batch = jax.jit(mjx.refit_bvh)(mx, dx_batch, rc.pytree())` followed by `jax.jit(mjx.render)(...)` in [`mjx/mujoco/mjx/warp/render_test.py`](https://github.com/google-deepmind/mujoco/blob/e23b5042013a1f7219d2950bf913ac981627bc61/mjx/mujoco/mjx/warp/render_test.py). JAX dispatches executables in order, so that pattern is safe.

One thing worth noting: the obvious/naive fix of threading the discarded `dummy` token into the returned `Data` is not sufficient. I tried it with a real, unfoldable dependency and confirmed (with help from Claude) in the optimized HLO that each ray cast then does depend transitively on its own refit, but the repro is completely unchanged at four bad frames out of five. It supplies the read-after-write edge (a ray cast can no longer precede its own refit) but not the write-after-read edge (a later refit can still clobber the BVH before an earlier ray cast has read it), which merely legalises the "batch all the refits first" schedule that XLA was already picking. `jax.lax.optimization_barrier` is no help either, since XLA expands it away before scheduling and the artificial edge disappears with it.

### Steps for reproduction

1. Build the model below and a render context with `nworld=1`.
2. Render five distinct mocap poses, each in its own `jax.jit` call. These are correct, since JAX dispatches them in order.
3. Render the same five poses from a single `jax.jit` call.
4. Compare. I'd expect the two to agree exactly, but four of the five frames differ, each by `0.2` in normalized depth over exactly the pixels where the box should be.

```xml
<mujoco>
  <option gravity="0 0 0"/>
  <asset>
    <material name="mat" rgba="1 1 1 1"/>
  </asset>
  <worldbody>
    <camera name="cam" pos="0 0 1" euler="0 0 0" resolution="64 64"
      sensorsize="0.036 0.036" focal="0.012 0.012"/>
    <geom name="plane" type="plane" size="2 2 0.1" material="mat"/>
    <body name="box_body" pos="0 0 0.1" mocap="true">
      <geom name="box" type="box" size="0.1 0.1 0.1" material="mat"/>
    </body>
    <!-- MJX needs at least one degree of freedom. This body is weightless,
         collision free, and in a geom group the render context does not
         enable, so it cannot perturb the rendered image. -->
    <body name="dof_body" pos="0 0 5">
      <joint type="free"/>
      <geom name="dof_geom" type="sphere" size="0.01" group="3"
        contype="0" conaffinity="0"/>
    </body>
  </worldbody>
</mujoco>
```

The box slides far enough between poses that a bounding volume fit to one pose does not overlap the box in any other pose, which is what turns the reordering into a visible, deterministic difference rather than a subtle one.

<details>
<summary>Repro script</summary>

```python
import jax
import mujoco
from mujoco import mjx
from mujoco.mjx._src import forward
import numpy as np

m = mujoco.MjModel.from_xml_path("sliding_box.xml")
mx = mjx.put_model(m, impl="warp")
dx = mjx.put_data(m, mujoco.MjData(m), impl="warp")
rc = mjx.create_render_context(
    mjm=m, nworld=1, cam_res=(64, 64), render_rgb=True, render_depth=True
)


def at(x_pos):
    mocap_pos = dx.mocap_pos.at[0, 0].set(x_pos)
    return forward.forward(mx, dx.replace(mocap_pos=mocap_pos))


poses = [at(x) for x in (-0.8, -0.4, 0.0, 0.4, 0.8)]


def render_one(d):
    d = mjx.refit_bvh(mx, d, rc.pytree())
    return mjx.render(mx, d, rc.pytree())[1]


# One refit/render pair per executable, which JAX dispatches in order.
separately = [np.asarray(jax.jit(render_one)(d)) for d in poses]

# Every pair in a single executable, where XLA chooses the order.
together = jax.jit(lambda *ds: tuple(render_one(d) for d in ds))(*poses)

for i, (want, got) in enumerate(zip(separately, together)):
    got = np.asarray(got)
    if not np.array_equal(got, want):
        print(f"frame {i}: max_diff={np.abs(got - want).max():.4g} "
              f"npix={int((got != want).sum())}")
```

</details>

Observed:

```console
$ python repro.py
frame 0: max_diff=0.2 npix=50
frame 1: max_diff=0.2 npix=40
frame 2: max_diff=0.2 npix=36
frame 3: max_diff=0.2 npix=40
```

Expected: no output.

### Confirmations

hartikainen:
Fixes #3435.

### DIFF
--- a/doc/mjx.rst
+++ b/doc/mjx.rst
@@ -254,8 +254,9 @@ Hold a reference to ``rc`` for the lifetime of your program and pass ``rc.pytree
 downstream JAX functions.  The pytree is a lightweight handle that refers back to the
 context via an internal registry.
 
-Once the context is created, you can render images within a compiled JAX function. This involves updating the bounding
-volume hierarchy (BVH) and executing the raycaster:
+Once the context is created, you can render images within a compiled JAX function.
+:func:`~mujoco.mjx.render` refits the bounding volume hierarchy (BVH) to the current scene state before
+executing the raycaster, so a single call is all that is needed:
 
 .. code-block:: python
 
@@ -263,18 +264,22 @@ volume hierarchy (BVH) and executing the raycaster:
 
     @jax.jit
     def render_fn(mx, d, rc_pytree):
-        # 1. Update the BVH for the current scene state
-        d = mjx.refit_bvh(mx, d, rc_pytree)
-
-        # 2. Render all configured cameras
+        # 1. Render all configured cameras
         pixels, _ = mjx.render(mx, d, rc_pytree)
 
-        # 3. Extract the RGB tensor for the first camera (index 0)
+        # 2. Extract the RGB tensor for the first camera (index 0)
         rgb = get_rgb(rc_pytree, 0, pixels)
 
-        return rgb, d
+        return rgb
 
-    rgb, d = render_fn(mx, d, rc.pytree())
+    rgb = render_fn(mx, d, rc.pytree())
+
+.. NOTE::
+   :func:`~mujoco.mjx.refit_bvh` remains available for pipelines that refit without rendering, but
+   calling it before :func:`~mujoco.mjx.render` is redundant. The two are deliberately fused into one
+   FFI call: they communicate through Warp-owned memory rather than through the JAX graph, so as
+   separate calls XLA would be free to run every refit ahead of every ray cast and pair a frame with
+   another frame's bounding volumes.
 
 .. WARNING::
    The batch dimension ``nworld`` is fixed when the render context is created via
--- a/mjx/mujoco/mjx/_src/bvh.py
+++ b/mjx/mujoco/mjx/_src/bvh.py
@@ -24,7 +24,11 @@ import mujoco.mjx.warp as mjxw
 
 
 def refit_bvh(m: Model, d: Data, ctx: Any):
-  """Refit the scene BVH for the current pose."""
+  """Refit the scene BVH for the current pose.
+
+  :func:`~mujoco.mjx.render` refits for itself, so this is only needed by
+  pipelines that use the BVH without rendering.
+  """
   if m.impl == Impl.WARP and d.impl == Impl.WARP and mjxw.WARP_INSTALLED:
     import mujoco.mjx.warp.render_context as mjxw_rc  # pylint: disable=g-import-not-at-top  # pytype: disable=import-error
     from mujoco.mjx.warp import bvh as mjxw_bvh  # pylint: disable=g-import-not-at-top  # pytype: disable=import-error
--- a/mjx/mujoco/mjx/_src/render.py
+++ b/mjx/mujoco/mjx/_src/render.py
@@ -37,7 +37,11 @@ def _require_segmentation_enabled(warp_rc) -> None:
 
 
 def render(m: Model, d: Data, ctx: Any) -> tuple[jax.Array, jax.Array]:
-  """Render packed RGB and depth buffers."""
+  """Render packed RGB and depth buffers.
+
+  Refits the scene BVH to the pose in ``d`` as part of the same FFI call, so
+  calling :func:`~mujoco.mjx.refit_bvh` beforehand is not needed.
+  """
   if m.impl == Impl.WARP and d.impl == Impl.WARP and mjxw.WARP_INSTALLED:
     from mujoco.mjx.warp import render as mjxw_render  # pytype: disable=import-error
     from mujoco.mjx.warp import render_context  # pytype: disable=import-error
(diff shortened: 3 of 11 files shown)


thowell:
@hartikainen thanks for reporting this issue and submitting a pr to fix it!

it would be great if we could resolve this without needing to change mujoco warp since my understanding is that the reported issue is related to the way jax optimizes jitted code and not the source code itself.

hartikainen:
> issue is related to the way jax optimizes jitted code and not the source code itself

Correct. I think we need to somehow tie `refit` and `render` together so that there's data flowing between them. I tried doing so with Claude but couldn't figure out a good way. I can try looking again later.

CC @tkelestemur @mhasek.

hartikainen (commented):
@thowell I drafted a fix for this with Claude. It required some weird prelude stuff for the warp shim. I can clean those up and make sure they're sound if you think the overall approach of fusing the `refit_bvh` and `render` together is reasonable.

btaba:
Will be fixed in #3435

PR #3437, opened 2026-07-27
Fix stretch stiffness basis for flexes in rotated parent bodies

smallquail:
The implicit effective metric assembles the stretch stiffness from world-space edge vectors, but a flex vertex body's slide dofs are expressed in its parent body's frame. When that frame is rotated the assembled operator is therefore not the Jacobian of the passive stretch force, which mj_flexPassiveStretch already maps into the dof frame with xmat^T. The metric is then inconsistent with the force it linearizes: implicit integration loses its stability guarantee, and models that the same flex handles comfortably in an unrotated frame diverge.

Apply the matching change of basis in both places that build or apply the stretch stiffness: mjd_flexStretch_mul rotates the input dof vector into world and the scattered result back, and mjd_flexStiff_assemble sandwiches each 3x3 block as R_bi^T * blk * R_bj. Both are no-ops when the parent is unrotated. Bending needs no change: its blocks are isotropic, and R^T (q I) R = q I.

This completes the fix in fe9dc584, which covered the passive force paths and the interp (trilinear) derivative but not the standard stretch one.

### DIFF
--- a/src/engine/engine_derivative.c
+++ b/src/engine/engine_derivative.c
@@ -245,7 +245,7 @@ void mjd_subQuat(const mjtNum qa[4], const mjtNum qb[4], mjtNum Da[9], mjtNum Db
 
   // add term linear in K * K
   mjtNum KK[9];
-  mju_mulMatMat3(KK, K, K);
+  mji_mulMatMat3(KK, K, K);
   mjtNum coef = 1.0 - (half_angle < 6e-8 ? 1.0 : half_angle / mju_tan(half_angle));
   mju_addToScl(Da_tmp, KK, coef, 9);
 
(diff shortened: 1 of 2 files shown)


src/engine/engine_derivative.c
```diff
@@ -1455,12 +1455,20 @@ void mjd_flexStretch_mul(const mjModel* m, mjData* d, mjtNum* res, const mjtNum*
         int v0 = vert[edge[e][0]], v1 = vert[edge[e][1]];
         int b0 = bodyid[v0],       b1 = bodyid[v1];
         g[e] = 0;
+        // the vertex bodies' slide dofs are expressed in their own (possibly rotated) frame while
```
yuvaltassa, line 1459: blank line above

src/engine/engine_derivative.c
```diff
@@ -1455,12 +1455,20 @@ void mjd_flexStretch_mul(const mjModel* m, mjData* d, mjtNum* res, const mjtNum*
...
+        // the stiffness is built from world-space edge vectors, so the operator must be sandwiched
+        // with R (dof -> world) and R^T (world -> dof); mj_flexPassiveStretch applies the same R^T
+        // to its world-space force. R = I for the common case of an unrotated parent body.
+        mjtNum w0[3] = {0}, w1[3] = {0};
+        if (m->body_dofnum[b0]) {
+          mju_mulMatVec3(w0, d->xmat + 9*b0, vec + m->body_dofadr[b0]);
```
yuvaltassa: use mji_ ?

(and elsewhere)

yuvaltassa (approved):
Nice catch!

quagla (approved):

PR #3439, opened 2026-07-28
Replace the flex metric factorization with a block preconditioner

smallquail:
Every step, the flex block of the effective metric M + K was factorized by sparse Cholesky, because K depends on the configuration. This replaces that factorization with the metric's per-vertex 3x3 diagonal blocks, prefactored.

### Where the time goes

`model/flex/bag.xml` (added here, 2868 dof), on upstream:

| | ms/step | share |
|---|---|---|
| build + factorize the metric | 9.64 | 50% |
| constraint solve, 39 CG iterations | 9.16 | 48% |
| everything else | 0.43 | 2% |
| **total** | **19.23** | |

The factorization goes away entirely. The blocks are O(n) to build and to apply, but weaker: CG runs about twice the iterations, and `qacc_smooth` becomes an iteration rather than a direct solve. Net **19.23 -> 8.29 ms/step**.

### Preconditioner, by case

The metric has two consumers, the CG constraint solver (`Mgrad = Mtilde \ grad`) and the `qacc_smooth` solve, and both apply the same operator.

| metric state | upstream | this PR |
|---|---|---|
| inactive: no flex elasticity, or an explicit integrator | `M^-1` | unchanged |
| bending only, `nefmK == 0` | `M^-1`, plus the exact constant bending factor from `mj_setConst` on the dofs it covers | unchanged |
| per-step stiffness, exactly factorizable | `M^-1`, plus a per-step sparse Cholesky of the flex block | `M^-1`, plus the per-vertex 3x3 blocks |
| per-step stiffness, M couples across the flex block | an inner PCG of up to 50 iterations over the row above, run once per outer CG iteration | `M^-1`, plus the per-vertex 3x3 blocks |

The constant bending factor is factored at model compile time and costs nothing per step, so it stays.

### Models affected

| model | upstream | this PR |
|---|---|---|
| `bag.xml`, per-step stretch | 19.23 ms/step, 39 CG | **8.29 ms/step**, 86 CG |
| `poncho.xml`, bending | 2.96 ms/step, 16.8 CG | 2.96 ms/step, 14.6 CG |
| `poncho_edgeequality.xml`, bending | 2.11 ms/step, 9.8 CG | 2.11 ms/step, 9.1 CG |
| `flag.xml`, `trampoline.xml`, metric inactive | unchanged path | unchanged path |

Only models carrying per-step stretch stiffness change in wall-clock. Bending-only models hold their timing and take slightly fewer CG iterations, because the preconditioner is now symmetric: it applies `M^-1` and the covered blocks to disjoint sets of dofs, where before the two overlapped and the operator was not symmetric, which PCG requires.

### Also in this change

- `mjd_effSolve` is renamed `mjd_effPrec`, since it is a preconditioner rather than a solve, and the name `mjd_effSolve` now belongs to the function that solves (M + K)x = b. The `qacc_smooth` call site in `mj_fwdAcceleration` is textually unchanged but now reaches the iterative solve.
- That solve converges on `opt.tolerance` rather than a hardcoded 1e-10, and raises `mjWARN_INERTIA` on reaching its iteration cap rather than returning an under-converged `qacc_smooth`.
- Covered dofs are located by walking the covered rows of the stiffness matrix. They are not necessarily 3-aligned from dof 0: any joint declared before a flexcomp shifts them.
- `model/flex/bag.xml` is added because no shipped model carried per-step stretch stiffness. The ponchos are bending-only, and `trampoline.xml` uses an explicit integrator, so the metric never activates there.

In `mjUSESINGLE` builds the default `opt.tolerance` of 1e-8 is below float epsilon, so the solve cannot reach it and would warn on every flex model. A floor on the tolerance would address that; it is not included here.


model/flex/bag.xml
```diff
@@ -0,0 +1,74 @@
+<!-- Copyright 2021 DeepMind Technologies Limited
```
yuvaltassa: New file — 2026.

```suggestion
<!-- Copyright 2026 DeepMind Technologies Limited
```

src/engine/engine_derivative.c
```diff
@@ -2901,128 +2901,193 @@ void mjd_effMulAdd(const mjModel* m, mjData* d, mjtNum* res, const mjtNum* vec)
...
+// Neither needs (M + K)^-1 exactly. A factorization costs O(n^1.5) or worse to build every step
+// and dominated the step; the blocks are O(n) to build and to apply, and the extra CG iterations
+// their weakness costs are cheaper than the factorization they replace.
+//
+// What this gives up: the metric is no longer inverted exactly anywhere, so qacc_smooth comes
+// from an iteration with a tolerance rather than a direct solve, and a flex whose (M + K) is too
+// ill-conditioned for block-diagonal preconditioning will converge slowly or hit the cap (which
+// mjd_effSolve reports). d->efm_active still records whether the metric would have been exactly
+// factorizable (see mjd_effBuild), but no longer selects a solve path: every covered model now
+// takes the same block-preconditioned route.
```
yuvaltassa: Reads well, but most of this is rationale, which lives in the commit message. Proposed tightening (17 → 5 lines):

```suggestion
// Build and factor the per-vertex 3x3 diagonal blocks of the flex part of (M + K), stored in
// d->efm_L, 9 numbers per covered vertex: O(n) to build and apply, approximate where the sparse
// factorization it replaces was exact. Both consumers use the blocks as a preconditioner: the CG
// constraint solver (Mgrad = Mtilde \ grad) and the qacc_smooth PCG in mjd_effSolve, which
// supplies the accuracy. efm_active still records exactness but no longer selects a solve path.
```

src/engine/engine_derivative.c
```diff
@@ -2901,128 +2901,193 @@ void mjd_effMulAdd(const mjModel* m, mjData* d, mjtNum* res, const mjtNum* vec)
...
+      }
     }
-  }
+    mju_cholFactor(Bk, 3, mjMINVAL);
+    adr[k++] = i;
+    i += 3;
+  }
+  d->efm_L = B;
+  d->efm_dofid = adr;
+  d->efm_L_rownnz = NULL;   // no sparse factor is built any more
```
yuvaltassa: These three arrays are now permanently NULL while their declared extents stay nonzero: `MJDATA_ARENA_POINTERS_EFM` in mjxmacro.h still lists `efm_L_rownnz`/`efm_L_rowadr` with extent `nefmdof` and `efm_L_colind` with extent `nefmL`. The Python bindings build these views through `InitPyArray` (structs.h), which passes the null pointer with a nonzero size to `pybind11::array_t` — pybind11 then allocates a fresh *uninitialized* array, so `data.efm_L_rownnz` on any stretch-flex model silently returns garbage. (`mj_copyData` is safe — it null-checks.)

Since the sparse factor is gone for good, remove the three fields from `mjData`, `mjxmacro.h` and `references.h`, and regenerate introspect (its docs still say "Cholesky factor of diag(M)+K"). They shipped in 3.11.0, so the removal wants a changelog line. If they must stay, the `(nefmdof x 1)`/`(nefmL x 1)` comments in mjdata.h and the xmacro extents need fixing instead.

src/engine/engine_derivative.c
```diff
@@ -2901,128 +2901,193 @@ void mjd_effMulAdd(const mjModel* m, mjData* d, mjtNum* res, const mjtNum* vec)
...
+// Apply the metric preconditioner: M^-1 on the dofs no factor covers, and on the covered ones
+// either the per-step 3x3 blocks or, when there is no per-step stiffness to block (bending-only
+// flexes), the exact constant bending factor from mj_setConst. The blocks subsume those rows when
+// they exist, which is why the two are alternatives rather than composed.
+//
+// PCG requires a symmetric preconditioner, so the covered and uncovered parts must not see each
+// other: the covered rows come only from their own factor (zero on uncovered columns), hence the
+// uncovered rows must likewise not read the covered columns. Zeroing the covered entries of the
+// right-hand side before the qLD sweep enforces that; a full sweep would leave the operator
+// block-triangular and PCG would stall on any flex sharing a kinematic tree with other dofs.
```
yuvaltassa: Tighter, and drops the "subsume" claim, which isn't quite true in mixed models (a stretch flex plus a separate bending-only flex: the bending flex's dofs aren't covered by any block, and the constant factor is skipped anyway, so they get plain M^-1 — same as upstream):

```suggestion
// Apply the metric preconditioner: the per-step 3x3 blocks when they exist, else the constant
// bending factor from mj_setConst, on the dofs they cover; M^-1 on all other dofs. PCG requires
// symmetry, so covered and uncovered dofs must not see each other: zeroing the covered entries
// of the right-hand side before the qLD sweep keeps the uncovered rows from reading them.
```

src/engine/engine_derivative.c
```diff
@@ -2901,128 +2901,193 @@ void mjd_effMulAdd(const mjModel* m, mjData* d, mjtNum* res, const mjtNum* vec)
...
+
+// iteration cap for the metric solve; the block preconditioner is weak enough that a stiff flex
+// can need most of these, so the cap is generous and exceeding it is reported, not silent
+#define mjEFF_MAXITER 100
 
+// accurate solve of (M + K) x = b by PCG preconditioned with the 3x3 blocks; used for qacc_smooth
+// where mjd_effPrec is only a preconditioner-grade approximation. Converges on the relative
+// residual to opt.tolerance; reaching mjEFF_MAXITER means x is under-converged, which is a
+// modeling problem (an effective metric too ill-conditioned for the block preconditioner), so
+// report it as mjWARN_INERTIA against the worst-residual dof rather than failing silently.
```
yuvaltassa: Same idea (9 → 6 lines including the define):

```suggestion
// iteration cap for the metric solve
#define mjEFF_MAXITER 100

// accurate solve of (M + K) x = b by PCG with the 3x3 block preconditioner, converging the
// relative residual to opt.tolerance; used for qacc_smooth. Reaching mjEFF_MAXITER means the
// metric is too ill-conditioned for the blocks: warn (mjWARN_INERTIA, worst-residual dof) and
// return x under-converged.
```

src/engine/engine_derivative.c
```diff
@@ -2901,128 +2901,193 @@ void mjd_effMulAdd(const mjModel* m, mjData* d, mjtNum* res, const mjtNum* vec)
...
+  mjtNum* z = mjSTACKALLOC(d, nv, mjtNum);
+  mjtNum* p = mjSTACKALLOC(d, nv, mjtNum);
   mjtNum* Ap = mjSTACKALLOC(d, nv, mjtNum);
-  mjtNum* psr = mjSTACKALLOC(d, d->nefmdof > 0 ? d->nefmdof : 1, mjtNum);
-  mjtNum* psz = mjSTACKALLOC(d, d->nefmdof > 0 ? d->nefmdof : 1, mjtNum);
-  int nbd = m->nefm0dof;
-  mjtNum* bfr = mjSTACKALLOC(d, nbd > 0 ? nbd : 1, mjtNum);
-  mjtNum* bfz = mjSTACKALLOC(d, nbd > 0 ? nbd : 1, mjtNum);
+  mju_zero(x, nv);
+  mju_copy(r, b, nv);
```
yuvaltassa: If a caller passes `x == b`, the zero destroys `b` before it is read and the solve returns 0. The old code supported aliasing (`if (x != b)`), as do `mjd_effPrec` and `effBlockApply`; both current callers pass distinct buffers, so this is latent — but it's MJAPI. Swapping the two lines fixes it (or document the restriction in the header):

```suggestion
  mju_copy(r, b, nv);
  mju_zero(x, nv);
```

src/engine/engine_derivative.c
```diff
@@ -2901,128 +2901,193 @@ void mjd_effMulAdd(const mjModel* m, mjData* d, mjtNum* res, const mjtNum* vec)
...
+  mjtNum bn = mju_dot(r, r, nv);
+  if (bn > mjMINVAL) {
+    // converge the relative residual to opt.tolerance; both sides are squared norms
+    mjtNum tol = m->opt.tolerance*m->opt.tolerance*bn;
```
yuvaltassa: Re the mjUSESINGLE note in the description: the consequence is worse than warning spam — the squared target of 1e-16 is unreachable in float, so every step of every stretch-flex model runs the full 100-iteration cap *and then* warns. I'd fold the floor into this PR, e.g.

```c
#ifdef mjUSESINGLE
  mjtNum tolerance = mju_max(m->opt.tolerance, 1e-6);
#else
  mjtNum tolerance = m->opt.tolerance;
#endif
```

src/engine/engine_derivative.c
```diff
@@ -2901,128 +2901,193 @@ void mjd_effMulAdd(const mjModel* m, mjData* d, mjtNum* res, const mjtNum* vec)
...
+    int converged = 0;
+    effBlockApply(m, d, z, r);
+    mju_copy(p, z, nv);
+    mjtNum rz = mju_dot(r, z, nv);
+    for (int it = 0; it < mjEFF_MAXITER; it++) {
+      mju_mulSymVecSparse(Ap, d->M, p, nv, m->M_rownnz, m->M_rowadr, m->M_colind);
+      mjd_effMulAdd(m, d, Ap, p);
+      mjtNum pAp = mju_dot(p, Ap, nv);
+      if (pAp < mjMINVAL) { converged = 1; break; }
```
yuvaltassa: Breakdown (tiny curvature with the residual not yet below tol) is reported as convergence and skips the warning. Since `bn > mjMINVAL` already guards the trivial-b case, leaving `converged` unset here would let the warning fire on a genuinely singular metric.

src/engine/engine_derivative.c
```diff
@@ -2901,128 +2901,193 @@ void mjd_effMulAdd(const mjModel* m, mjData* d, mjtNum* res, const mjtNum* vec)
...
+      mjtNum alpha = rz/pAp;
+      mju_addToScl(x, p, alpha, nv);
+      mju_addToScl(r, Ap, -alpha, nv);
+      if (mju_dot(r, r, nv) < tol) { converged = 1; break; }
+      effBlockApply(m, d, z, r);
+      mjtNum rznew = mju_dot(r, z, nv);
+      mju_scl(p, p, rznew/rz, nv);
+      mju_addTo(p, z, nv);
```
yuvaltassa: One pass instead of two, as the deleted code had it:

```suggestion
      mju_addScl(p, z, p, rznew/rz, nv);
```

src/engine/engine_derivative.c
```diff
@@ -2901,128 +2901,193 @@ void mjd_effMulAdd(const mjModel* m, mjData* d, mjtNum* res, const mjtNum* vec)
...
+      rz = rznew;
+    }
+
+    // out of iterations: x is under-converged, blame the dof carrying the largest residual
+    if (!converged) {
+      int worst = 0;
+      for (int i = 1; i < nv; i++) {
+        if (mju_abs(r[i]) > mju_abs(r[worst])) worst = i;
+      }
+      mj_warning(d, mjWARN_INERTIA, worst);
```
yuvaltassa, line 3065: This prints "Inertia matrix is too close to singular at DOF %zu. Check model." — a user whose real problem is metric conditioning vs. tolerance gets pointed at their inertia. Reusing the enum avoids an ABI addition, which seems right; flagging in case you'd rather adjust the warning text.

src/engine/engine_derivative.c
```diff
@@ -3044,263 +3109,6 @@ typedef struct {
   int nperm;             // emitted count
 } mjEffND;
```
yuvaltassa: `effNDOrder` and `effFactor` were its only consumers — this typedef (lines 3096–3110) is now dead and can go too.

src/engine/engine_derivative.c
```diff
@@ -3369,7 +3177,7 @@ void mjd_effBuild(const mjModel* m, mjData* d, int active, int flg_factor) {
 
   // preconditioner exactness (efm_active == 2): when every dof the stiffness touches sits on
   // a simple slider body (diagonal M row, no kinematic children), M has no coupling across
-  // the covered block, so blockdiag(qLD, flex factor) is exactly (M+K)^-1 and mjd_effSolve
+  // the covered block, so blockdiag(qLD, flex factor) is exactly (M+K)^-1 and mjd_effPrec
   // skips the refinement. Interp flexes outside the assembled CSR act only through the
   // matrix-free operator (no factor rows), which breaks exactness.
```
yuvaltassa: Stale on two counts: there is no refinement to skip (`mjd_effPrec` applies the blocks unconditionally), and blockdiag(qLD, blocks) is no longer exactly (M+K)^-1 even in this case, since the blocks drop the vertex–vertex K coupling. Same for the public doc in mjdata.h ("2 active + preconditioner exact"). Since nothing selects a path on 2 vs 1 anymore, consider collapsing `efm_active` to 0/1 — or keep 2 but re-document it as a record of exact factorizability.

test/xml/xml_write_read_test.cc
```diff
@@ -65,6 +65,7 @@ std::vector<std::string> GetWriteReadTestModels() {
             absl::StrContains(xml, "fromto_convex") ||
             absl::StrContains(xml, "cube_skin") ||
             absl::StrContains(xml, "cube_3x3x3") ||
+            absl::StrContains(xml, "bag") ||
             // exclude files that fail since we do not save pinned flex nodes
```
yuvaltassa: bag.xml is excluded *because* of its pins, so it belongs under this comment; and bare "bag" matches any future path containing the substring — "flex/bag" is safer:

```suggestion
            // exclude files that fail since we do not save pinned flex nodes
            absl::StrContains(xml, "flex/bag") ||
```

test/engine/engine_derivative_test.cc
```diff
@@ -2104,9 +2104,11 @@ TEST_F(DerivativeTest, FlexStiffAssembleInterp) {
...
 }
 
-// mjd_effSolve: exact-preconditioner fast path solves (M+K)x = b directly; the general
-// refinement path stays within its tolerance when exactness does not hold
-TEST_F(DerivativeTest, EffSolveExact) {
+// mjd_effSolve drives (M+K)x = b to opt.tolerance on the relative residual,
+// for every metric coverage case. It is a PCG whose preconditioner
+// (mjd_effPrec) is only approximate, so the accuracy comes from the iteration
+// and not from the preconditioner being exact.
+TEST_F(DerivativeTest, EffSolve) {
```
yuvaltassa, line 2111: Two cheap additions worth considering: (1) the under-convergence path has no test — `opt.tolerance = 0` on the stretch model should deterministically hit the cap and raise `mjWARN_INERTIA`; (2) since preconditioner symmetry is a headline fix of this PR, a `u·P(v) == v·P(u)` check on random vectors in the jointed-parent model would pin it down.

quagla (approved):

PR #3445, opened 2026-07-30
Fix LuGre bristle velocity integration

giusenso:
This PR solves the issue described in https://github.com/google-deepmind/mujoco/issues/3446.

---

The test introduced in the commit f95d50c12f0f55efa484fb901347157615855dae proves the efficacy of the fix. Prior to the change, placing a multi-output (`SO3`) actuator before a DC motor caused `actuator_id` and `actuator_outadr` to diverge, resulting in the LuGre bristle state integrating against an incorrect velocity vector.

the commit 4c2653273c0159ff6ee4b1c0c27206623e1d8d0a fixes the bug by indexing `d->actuator_velocity` using `m->actuator_outadr[actuator_id]`.


yuvaltassa (approved):
Very nice, thank you

PR #3463, opened 2026-08-05
Add continuous-collision infrastructure for deformables

smallquail:
The standard discrete collision pipeline generates contact points at a configuration; this new module prices gaps along **trajectories**:

- differentiable vertex-triangle / edge-edge / vertex-geom distance kernels with closest-point barycentrics (`mjc_PtTri`, `mjc_SegSeg`, `mjc_GeomDist`, plus sharp-feature extraction `mjc_GeomVerts`/`mjc_GeomEdges`);
- swept-volume contact-candidate generation over the flex BVH (`mjc_candidates`);
- per-pair gap evaluation with the gradient's direction and vertex weights (`mjc_pairGap`);
- conservative advancement bounding each pair's time of impact (`mjc_advance`).

### Design notes

- **`mjcPair` is the geometric identity of a candidate pair only.** Solver state (multipliers, ages) and cached linearizations belong to the consumer, not to this struct — the module neither reads nor writes them.
- **No hard-coded constants.** The two lengths the module needs are caller-supplied parameters:
  - the **standoff cap** bounds the *rest gap*: an active pair settles at a small positive distance δ = min(pair band, cap) rather than at zero. Strictly positive so the conservative advancement always sees a nonzero time of impact (a pair resting at exactly zero would lock the advance); the per-pair min keeps thin participants' skins proportionally thin, and the cap stops thick ones carrying a fat layer.
  - the **detection band** is the *tracking reach*, and is not a physical length: how early the broad phase starts holding a pair and how long the gap evaluation keeps it held. Wider costs candidates; narrower hands pairs to the consumer later.
- **Flex-flex gaps are measured at the midsurface**, not skin-to-skin: where the mesh geometry is tighter than the combined radii (a string threaded through a hem), a skin gap is permanently negative and the pair would be discarded as invalid — losing CCD coverage exactly where tunneling is likeliest. The broad phase adds the radii back into its reach, so detection range is unchanged.
- Engine-internal (`MJAPI` on entry points only for tests, marked not a supported API). No user-visible behavior change; changelog notes the new infrastructure.

### Testing

`engine_collision_continuous_test`: distance kernels across their regions (interior/edge/vertex, crossing/collinear/parallel), geom signed distance and sharp features, the pair gap with its gradient verified by central differences at every involved vertex, the conservative advancement (analytic cap on a crossing sweep; the conservative closing-rate bound; small-motion early-out), and candidate generation on stacked cloths (pairs found at 2 mm, none at 0.5 m). Full suite: 1349/1349.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -34,6 +34,13 @@ Actuation
 Engine
 ^^^^^^
 
+- Added continuous-collision infrastructure for deformables (``engine_collision_continuous``): differentiable
+  vertex-triangle, edge-edge and vertex-geom distance kernels with closest-point barycentrics, swept-volume contact
+  candidate generation over the flex bounding-volume hierarchy, per-pair gap evaluation with the gradient's vertex
+  weights, and a conservative advancement that bounds each pair's time of impact. Engine-internal with no consumer in
+  this change: the discrete pipeline generates contact points at a configuration, this module prices gaps along
+  trajectories, and it is the groundwork for continuous-contact (IPC-style) solvers for flex.
+
 - Replaced the per-step sparse Cholesky factorization of the flex block of the implicit effective metric M + K with
   its prefactored per-vertex 3x3 diagonal blocks. The blocks precondition the CG constraint solver and drive an
   iterative solve for ``qacc_smooth``, which now converges on :ref:`tolerance<option-tolerance>` rather than a fixed
--- a/src/engine/CMakeLists.txt
+++ b/src/engine/CMakeLists.txt
@@ -17,6 +17,8 @@ set(MUJOCO_ENGINE_SRCS
     engine_callback.c
     engine_callback.h
     engine_collision_box.c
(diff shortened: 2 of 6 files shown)


doc/changelog.rst
```diff
@@ -34,6 +34,13 @@ Actuation
 Engine
 ^^^^^^
 
+- Added continuous-collision infrastructure for deformables (``engine_collision_continuous``): differentiable
```
yuvaltassa, line 37: meh, none of this is in the public API, right? I'd just delete this and make it a line in the actual IPC commit when that lands.

yuvaltassa (commented):
Can you ask @kbayes to review? I feel like the unit tests could be improved and also would his eyes on this re consistency with the rest of the collision code.

quagla:
Superseded by b924ac66bdf3251e2192c153de2280f6f8d3ecfc

PR #3465, opened 2026-08-05
Cache MJX-Warp FFI callables across retraces

### LINKED ISSUE #3464: MJX-Warp creates a new FFI callable on every equivalent JAX retrace
## Intro

The discussion in [google-deepmind/mujoco_warp#1385](https://github.com/google-deepmind/mujoco_warp/pull/1385) exposed a separate issue: MJX creates a new Warp `FfiCallable` when JAX retraces the same work. The reporter's [minimal repro](https://github.com/gholmes829/mjx-warp-callable-leak) isolates the registry growth.

## My setup

- MuJoCo `main` at [`596b6f43`](https://github.com/google-deepmind/mujoco/commit/596b6f433d78005322d4abb8f153a110206365cf)
- MuJoCo / MJX 3.11.1
- Warp 1.15.0
- JAX 0.8.3
- Python 3.14.6
- Linux x86-64

The test below mocks FFI registration, so it does not require a GPU.

## What's happening? What did you expect?

`jax_callable_variadic_tuple()` defines an inner `func_wrapper`. Every retrace creates a new function object and passes it to `warp_ffi.jax_callable()`.

Warp uses that function object as part of its registry key. The new identity misses Warp's cache even though the original shim, flattened signature, PyTree, graph mode, output dimensions, and argument options are unchanged.

The code below models two retraces of the same shim. It calls `warp_ffi.jax_callable()` twice. It should call it once and reuse the resulting callable.

Each `FfiCallable` has its own graph cache. Creating another callable also creates another place for captured graphs and related resources to accumulate.

The long-running OOM described in `mujoco_warp#1385` has not been independently reproduced. The duplicate registration itself is deterministic and does not require a GPU.

MJX should cache the constructed Warp callable by the original shim and its call structure. Arrays and JAX tracers are inputs to the callable, not part of its identity.

The reporter found that reusing one JIT or switching to `WARP_STAGED` avoids the failure in their workload. Those are useful controls, but they do not fix the duplicate registration shown here for `GraphMode.WARP`.

## Steps for reproduction

1. Check out MuJoCo `main` at `596b6f43`.
2. Install the MJX-Warp dependencies.
3. Save the code below as `repro.py`.
4. Run `python repro.py`.
5. Observe `jax_callable constructions: 2`. The expected count is 1.

## Minimal model for reproduction

No model is required.

## Code required for reproduction

```python
from unittest import mock

from mujoco.mjx.warp import ffi


def shim(values: tuple[int, ...], output: int):
  del values, output


def create_callable(*args, **kwargs):
  del args, kwargs
  return mock.Mock()


with mock.patch.object(
    ffi.warp_ffi,
    'jax_callable',
    side_effect=create_callable,
) as jax_callable:
  # Model two retraces constructing the same MJX wrapper.
  wrapper = ffi.jax_callable_variadic_tuple(shim)
  wrapper_again = ffi.jax_callable_variadic_tuple(shim)

  wrapper((1, 2))
  wrapper_again((3, 4))

  print('jax_callable constructions:', jax_callable.call_count)
  assert jax_callable.call_count == 1
```

On `main`, the script prints:

```text
jax_callable constructions: 2
```

and fails the assertion.

## Confirmations

- [x] Searched the current documentation for MJX-Warp FFI and retracing guidance.
- [x] Searched existing MuJoCo issues and discussions; no duplicate was found.

shi-eric:
## Summary

`jax_callable_variadic_tuple()` built a new inner wrapper on every JAX retrace. Warp keys its FFI callable registry by wrapper identity, so structurally equivalent retraces registered separate `FfiCallable` objects and maintained separate graph caches.

A process-wide MJX cache now reuses callables when the original shim, flattened signature, PyTree, graph mode, and callable options match. Runtime arrays and JAX tracers remain per-call inputs and are not included in the cache key.

Lookup and construction are protected by a lock to prevent concurrent traces from registering the same callable twice. Invocation remains outside the lock.

The regression test checks that:

- equivalent traces of the same shim reuse one callable;
- distinct shims with the same signature remain separate;
- a different tuple structure creates a separate callable.

## Validation

- `ffi_test.py` and `types_test.py`: 3 passed
- CUDA `StepTest::test_step2` (humanoid, batch size 7, `GraphMode.WARP`): passed
- CUDA `StepTest::test_step3` (humanoid, batch size 7, `GraphMode.WARP_STAGED`): passed
- the short retrace repro kept the Warp registry at 4 callables across 36 equivalent retraces
- the step callable reached Warp's 32-entry graph-cache limit and remained there (33 captures total including setup)
- disabling the MJX cache in the repro restored the previous callable growth: 4, 6, 8, 10, 12, 14
- pre-commit hooks passed
- `git diff --check` passed

CUDA validation used JAX 0.8.3, Warp 1.15.0, and an NVIDIA RTX PRO 6000 Blackwell GPU. The reporter's long-running OOM workload was not exercised. CUDA validation independently reproduced the callable-growth mechanism and confirmed that callable registration and graph-cache growth remain bounded with this change.

Fixes #3464.

Context: [google-deepmind/mujoco_warp#1385](https://github.com/google-deepmind/mujoco_warp/pull/1385)

### DIFF
--- a/mjx/mujoco/mjx/warp/ffi.py
+++ b/mjx/mujoco/mjx/warp/ffi.py
@@ -17,6 +17,7 @@
 import dataclasses
 import functools
 import inspect
+import threading
 import typing
 from typing import Any, Callable, Optional, Sequence, Tuple, Union
 
@@ -28,6 +29,15 @@ from mujoco.mjx.third_party.warp._src.jax import ffi as warp_ffi
 
 from mujoco.mjx._src.types import tree_path_to_attr_str
 from mujoco.mjx.warp import types as mjx_warp_types
+
+
+# ``warp_ffi.jax_callable`` keys its registry by wrapper function and
+# configuration. Cache generated wrappers here by the original shim and MJX's
+# flattened call structure so equivalent retraces reuse the same Warp entry.
+_JAX_CALLABLE_VARIADIC_TUPLE_REGISTRY: dict[
+    tuple[Any, ...], Callable[..., Any]
+] = {}
+_JAX_CALLABLE_VARIADIC_TUPLE_REGISTRY_LOCK = threading.Lock()
 
 
 def flatten_signature(signature: inspect.Signature, args: Tuple[Any, ...]):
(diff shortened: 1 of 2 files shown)


mjx/mujoco/mjx/warp/ffi.py
```diff
@@ -30,6 +31,15 @@
 from mujoco.mjx.warp import types as mjx_warp_types
 
 
+# ``warp_ffi.jax_callable`` keys its registry by wrapper function and
+# configuration. Cache generated wrappers here by the original shim and MJX's
+# flattened call structure so equivalent retraces reuse the same Warp entry.
+_JAX_CALLABLE_VARIADIC_TUPLE_REGISTRY: dict[
```
btaba, line 37: do you recommend we use a fixed registry size here or will the size of this grow unbounded?

btaba (commented):
Thanks @shi-eric for the amazing contribution and bugfix! Just one quick question

mjx/mujoco/mjx/warp/ffi.py
shi-eric: Hey @btaba, I think it's better to keep this unbounded for consistency with Warp's FFI callable registry: https://github.com/NVIDIA/warp/blob/60064d1833f618e4baeae8f71050d0257c564bbf/warp/_src/jax/ffi.py#L39. The only cap in Warp is on each callable's graph cache. Adding a cap here would not remove the corresponding Warp registration when an entry is evicted. If the same structure came up again, we would just create another registration.

For now, this change should go a long way toward addressing the existing problem. I recommend only revisiting the uncapped behavior if we find a workload that actually needs a cap.

btaba: Thanks for the explanation @shi-eric LGTM!

btaba (approved):

PR #3466, opened 2026-08-06
Integrate passive flex contact implicitly

smallquail:
Contact of a flex with [`passive`](https://mujoco.readthedocs.io/en/latest/XMLreference.html#flex-contact-passive) collisions is applied as a penalty force rather than handed to the constraint solver. That force was an explicit spring of fixed stiffness `1e4`, and the timestep bounds what an explicit spring can be: any stiffness worth having oscillates faster than the step resolves. The result was a force too soft to keep sheets apart, so interpenetration was routine.

This PR carries the contact's curvature in the effective metric `M + K` instead, alongside the flex's own stretch and bending stiffness. At a 2 ms timestep it holds roughly 50x the stiffness the explicit force could.

drape

`model/flex/drape.xml`, added here, at t = 0.29 s: three cloths dropped on a sphere, each resting on the one below it and on the sphere without interpenetrating.

### What changes

- The contact block `k·JᵀJ` is appended to the per-vertex candidate list that `mjd_flexStiff_assemble` already builds for the flex stencils, so it adds entries to an existing matrix rather than introducing a new one. The accompanying shift `−h·K·v` is what damps the stiff modes; without it the scheme is implicit in name only.
- Stiffness is now a natural frequency scaled by the participating vertex mass rather than a fixed `1e4`, so a single value suits models of any scale. It remains not user-specified.
- The feature requires an integrator whose constraint solve runs in that metric (`implicit` or `implicitfast` with the CG solver, pyramidal cones and sleep disabled), and is rejected with an error otherwise.
- `model/flex/sphere_passive.xml` is replaced by `drape.xml`.

### Scope

Passive handling covers contacts whose every dof is a flex vertex carried by the metric: flex against flex, flex against itself, and flex against **static** geometry, which contributes no dofs of its own. For those the Hessian is assembled in full. Contact with a body that can move would have that body's dofs dropped from the Hessian, and stays on the constraint solver.

### Limitation

Friction is not modelled on this path — the force is purely normal, and all passive contacts are frictionless regardless of the specified `condim`. A flex therefore slides freely over static geometry: in the model above the sheets drape, then slide off the sphere. This is documented in XMLreference. Being a penalty force, it also does not guarantee non-penetration.

### Tests

`ImplicitIntegratorTest.PassiveFlexContactIsImplicit` drops a sheet on a sheet and asserts the upper one stays above the lower one's lowest point, and that peak speed stays bounded. On `main` the same model reaches 143 m/s with self-collision enabled; here it stays at 2.2.


yuvaltassa (approved):
Much of the ancillary text here (changelog item, comments) can be compacted, otherwise LGTM

PR #3469, opened 2026-08-10
Publish passive flex contact as a metric rank-1 class

smallquail:
Passive flex contact (`flex_passive`) is integrated implicitly: its stiffness `k·JᵀJ` has to be part of the effective metric `M + h²K` that the solvers work with, or the step is not stable. Until now that stiffness was written into the flex stiffness CSR, the sparse matrix that holds the elastic stiffness. That needed contact-specific code in three places, which all had to agree on what a contact is: the stiffness assembler `mjd_flexStiff_assemble` (a `flg_contact` parameter and contact branches in its candidate counting, candidate fill and value fill), a helper mapping contacts onto vertex slots (`contactFlexSlots`), and the velocity shift `−h·K·v` in `mjd_effShift`, which recomputed every contact Jacobian a second time through `mjd_flexContact_mul`. It also made the CSR's sparsity depend on the contact set, and expanded each contact, a rank-1 term, into `nnz²` matrix entries.

### The change

The metric already has a way to carry rank-1 terms without assembling them: tendon and actuator rows are stored as scaled sparse rows and applied through `mjd_effRank1Next`, whose contract is that a new kind of term becomes a new case there, invisible to every consumer. This PR makes passive contact such a case.

`effContactBuild` linearises each passive contact once per step into a row `scale·k·J_n`, stored in two packed arrays on `mjData`. Everything that needs contact reads those rows: the metric matvec and its island variant, the solver's allocation pass and Newton's Hessian assembly through the existing iterator, and the shift directly. Contact stays in Newton's Hessian while leaving the CSR. The contact branches in the assembler, its `flg_contact` parameter, `contactFlexSlots` and `mjd_flexContact_mul` are deleted, so there is one definition of contact instead of three, and the CSR again depends only on the mesh (on `model/flex/drape.xml` it drops from 54567 to 51219 non-zeros and stops tracking the contact count).

`mjd_effMulAdd` already had a `flg_contact` argument reserved for this, inert until now. It works now: the linear solves include contact, and a caller that accounts for contact energy separately can leave it out. It is also the entry point for a different contact producer: fill the same rows and every consumer sees them, without touching the metric's internals.

### Behaviour change

Because contact was in the CSR, the block preconditioner picked up its diagonal entries. It no longer does, and that is an improvement: a contact pair with `w ≈ [n, −n]` only resists the **relative** motion of its two vertices, but its diagonal entries resist each vertex's **absolute** motion, so the preconditioner was stiffening directions the operator leaves free.

The rows carry the full contact Jacobian chain, where the CSR path kept only flex-vertex dofs. Against static geometry the two agree; for a flex vertex on a moving body the metric now carries the coupling to that body that the shift already had.

Trajectory-level numbers on `drape.xml` are deliberately not quoted: a 1e-15 perturbation to one coordinate changes the contact count by ~20 and the peak speed by 20–50% within 3 s, so they measure rounding, not the change. From an identical loaded state (286 passive contacts, no constraint-solver iterations), 30 single steps take 8.9 ms instead of 10.9, the same under CG/implicitfast, CG/discrete and Newton/discrete; a third of the difference is the assembly, two thirds the `qacc_smooth` PCG.

### API

- `mjData.efm_con_ind` / `mjData.efm_con_val`: the contact rows packed `[nnz, colind…]` and `[scale, val…]`, sized by one `nefmcon` count; reset in `mj_resetData` and `mj_clearEfc`
- `mjd_effRank1Next` takes `flg_contact`; `mjd_flexStiff_assemble` loses it; `mjd_flexContact_mul` is removed
- generated files updated: `references.h`, `introspect/structs.py`, wasm and C# bindings


src/engine/engine_derivative.c
```diff
@@ -4016,13 +3879,82 @@ void mjd_effMulAddIsland(const mjModel* m, const mjData* d, mjtNum* res, const m
...
 
 
+// Publish passive flex contact as the metric's rank-1 contact class: pair c contributes
+// scale*k_c * J_c' J_c, one term per contact. Rows are packed back to back, [nnz, colind...] in
+// efm_con_ind and [scale, val...] in efm_con_val, so one count sizes both arrays. Contact is a sum of
+// rank-1 terms, so applying it matrix-free costs O(nnz) per pair where assembling it into the
+// stiffness CSR costs O(nnz^2) both to store and to apply -- and the CSR's sparsity stops
+// growing with the contact set, becoming a function of the mesh alone.
+//
+// The rows are the same contact-frame normal Jacobians mjd_flexContact_mul applies, so the
```
yuvaltassa: Stale self-reference: this comment cites `mjd_flexContact_mul`, which this PR deletes. Suggest e.g. "The rows are the contact-frame normal Jacobians the deleted CSR path assembled, so the operator and the shift -h*K*v see one definition of contact."

src/engine/engine_derivative.c
```diff
@@ -1804,8 +1763,7 @@ mjtBool mjd_flexStiff_any(const mjModel* m, int flg_interp) {
 // does this standard flex contribute implicit stiffness under the given term flags?
 // A flex participates if it has elasticity OR passive contacts: the contact stiffness may be the
 // only stiffness, so vertex slots must exist either way.
-static mjtBool flexMetric_participates(const mjModel* m, int f, int flg_bend, int flg_stretch,
-                                       int flg_contact);
+static mjtBool flexMetric_participates(const mjModel* m, int f, int flg_bend, int flg_stretch);
```
yuvaltassa: With the contact arm gone, `flexMetric_participates` is `flexStiff_active` verbatim, and this comment describes the deleted behavior ("the contact stiffness may be the only stiffness, so vertex slots must exist either way" — no longer true: contact-only flexes now carry no vertex slots, by design). Suggest deleting the wrapper (forward decl here, definition at L1816) and calling `flexStiff_active` directly at L1845, with a one-line note that contact no longer forces slots.

src/engine/engine_derivative.c
```diff
@@ -3394,6 +3225,22 @@ int mjd_effRank1Next(const mjModel* m, const mjData* d, mjEffRank1Iter* it, mjEf
...
+      break;
+    }
+    int adr = it->k, nnz = d->efm_con_ind[adr];   // packed row: header then entries
+    e->nnz = nnz;
+    e->scale = d->efm_con_val[adr];
+    e->colind = d->efm_con_ind + adr + 1;
+    e->val = d->efm_con_val + adr + 1;
+    it->k = adr + 1 + nnz;
+    return 1;
+  }
```
yuvaltassa, line 3234: Non-blocking, for the future: the island paths iterate this class with `flg_contact=1`, but the island matvec invariant ("union-find merges the support, so the first column decides membership") has no enforcer for contact rows. Tendons get `unionTendonTrees` and one flex gets the star-merge in engine_island.c, but a passive contact between two *different* flexes has no efc row and nothing merges their trees. Unreachable today (`effFlexAny` forces a monolithic solve; sleep+flex is an mjERROR), but when the TODO on island-local flex operators is tackled, findEdges will need a contact-driven merge (walk exclude==4 contacts, union the two sides). Worth a breadcrumb next to the star-merge.

src/engine/engine_derivative.c
```diff
@@ -4016,13 +3879,82 @@ void mjd_effMulAddIsland(const mjModel* m, const mjData* d, mjtNum* res, const m
...
+    }
+    int NV = mj_contactJacobian(m, d, con, con->dim, jacdif, NULL, jac1, jac2, NULL, NULL, chain);
+    if (!NV) {
+      continue;
+    }
+    mju_mulMatMat(jacn, con->frame, jacdif, con->dim > 1 ? 3 : 1, 3, NV);
+    d->efm_con_ind[adr] = NV;
+    d->efm_con_val[adr] = scale * k;
+    for (int a = 0; a < NV; a++) {
+      d->efm_con_ind[adr + 1 + a] = chain[a];
```
yuvaltassa, line 3932: The behavior change advertised in the description — full Jacobian chain, so a flex pinned to a moving body now has its base coupling in the metric, not just in the shift — has no test: the existing passive-contact tests pin to the world or leave the flex free. A moving-base variant (e.g. `PassiveFlexContactInMetric` with the lower sheet pinned to a free or hinged body, asserting no BADQACC and bounded vmax) would pin it.

src/engine/engine_derivative.c
```diff
@@ -4128,15 +4060,17 @@ void mjd_effBuild(const mjModel* m, mjData* d, int active, int flg_factor) {
   }
   if (assemble_any) {
     d->nefmK = mjd_flexStiff_assemble(m, d, d->efm_K_rownnz, d->efm_K_rowadr,
-                                      NULL, NULL, h*h, h, /*bend*/ 1, /*stretch*/ 1,
-                                      /*contact*/ 1, krot);
+                                      NULL, NULL, h*h, h, /*bend*/ 1, /*stretch*/ 1, krot);
   }
+
+  // passive flex contact is a rank-1 class, not CSR entries (see effContactBuild)
+  effContactBuild(m, d, h*h);
```
yuvaltassa, line 4058: Question on the preconditioner argument: it covers the flex-flex pair case (w ~ [n, -n]), where the diagonal entries stiffen directions the operator leaves free. For flex-vs-static contacts the dropped diagonal k*nn^T is a faithful diagonal of the operator, so the block preconditioner is strictly weaker there. Your drape timing still wins (per-iteration CSR cost shrinks), but did PCG iteration counts move on drape? No action needed if iterations held.

include/mujoco/mjdata.h
```diff
@@ -424,6 +425,8 @@ typedef struct mjData_ {
   int*    efm_K_colind;      // effective-stiffness CSR column indices           (nefmK x 1)
   mjtNum* efm_K_val;         // effective-stiffness CSR values                   (nefmK x 1)
   int*    efm_dofid;         // block k -> dof address of its vertex triple      (nefmdof x 1)
+  int*    efm_con_ind;       // contact rank-1 rows, packed [nnz, colind...]   (nefmcon x 1)
+  mjtNum* efm_con_val;       // contact rank-1 rows, packed [scale, val...]    (nefmcon x 1)
```
yuvaltassa: Size annotations sit at column 80; the rest of the block aligns them at 82 — two more spaces before `(nefmcon x 1)`. references.h mirrors this.

src/engine/engine_memory.h
```diff
@@ -79,6 +79,7 @@ static inline void mj_clearEfc(mjData* d) {
   // activity flag and counts consumers gate on must clear with them
   d->efm_active = 0;
   d->nefmT = d->nefmA = d->nefmK = d->nefmL = d->nefmdof = 0;
+  d->nefmcon = 0;
```
yuvaltassa: Join the chain above: `d->nefmT = d->nefmA = d->nefmK = d->nefmL = d->nefmdof = d->nefmcon = 0;`

yuvaltassa (commented):
Reviewed locally at 2e57cdd82: full test suite green, plus float32 and ASAN spot-runs of the passive-contact tests. Verified that the monolithic gate (`effFlexAny` = stiffness or contact possible) keeps the new rank-1 class out of the island paths, and that the shift's `-scale/h` application matches the deleted `mjd_flexContact_mul` exactly. Nothing blocking — comments inline: two stale comments, two nits, a future-islands note, a test suggestion, and one question on the preconditioner argument.

src/engine/engine_derivative.c
smallquail: Done, with your wording.

src/engine/engine_derivative.c
smallquail: Done: wrapper and forward declaration removed, `flexStiff_active` called directly, with a one-line note that contact no longer forces vertex slots.

src/engine/engine_derivative.c
smallquail: Added a note in engine_island.c next to the loop that unions the trees of every flex with metric terms: passive contact between two different flexes has no efc row and nothing merges their trees, so island-local flex operators will need a contact-driven merge there (walk the exclude==4 contacts, union the two sides).

src/engine/engine_derivative.c
smallquail: Added `PassiveFlexContactMovingBase`: the lower sheet pinned to a hinged 2 kg body, the upper sheet dropped off-centre; it asserts no BADQACC, peak speed under 4 m/s, a nonzero tilt of the base, and no pass-through. Two scene constraints worth knowing: pins on a jointed body are refused with bending, so the pinned sheet is stretch-only, and it is soft because its elastic coupling to the hinge is still explicit. With contact restricted to flex dofs this scene diverges within about 100 steps of contact onset; with the rows it holds.

src/engine/engine_derivative.c
smallquail: Measured from an identical loaded state, PCG iterations of the `qacc_smooth` solve per step: drape 215 → 165. On a cloth dropped on a static box, where every row is flex-vs-static, 43 → 68, so the point stands for static partners: there the diagonal was faithful and the solve is weaker without it. Left as is here; contact preconditioning gets a fuller treatment in the IPC integrator work, where the same rows are folded into the solver's preconditioner and the per-class choice can be measured properly.

include/mujoco/mjdata.h
smallquail: Done, `references.h` regenerated.

src/engine/engine_memory.h
smallquail: Done.

quagla (approved):

PR #3473, opened 2026-08-13
sysid: avoid mutating input in apply_delay

anxkhn:
## Summary

Copy time-series data before writing delayed sensor values.

`apply_delay()` constructed its result from the input array and then modified it in place, which also changed the caller's `TimeSeries`. The result now owns a copy before delayed samples are assigned.

## Tests

- Added a regression test confirming the input remains unchanged
- Verified the returned delayed signal differs as expected


anxkhn:
hey @yuvaltassa, could you please review this pr when you have a chance?

Issue #3472, opened 2026-08-12
USD: four defects in the Newton schemas v0.4.0 adoption (units, semantics, unread attribute)


andrewkaufman:
### Summary

`39e44588` ("Update USD support in MuJoCo to Newton USD schemas v0.4.0", unreleased — it sits above the `Version 3.11.0` heading in `doc/changelog.rst`) introduces four defects in the new `newton:` read paths. Two are angle-unit conversions that are simply missing; the third maps a distance-valued attribute onto a force-valued field; the fourth silently ignores an attribute that controls whether a constraint is active at all.

I wrote the original adoption in #3156, so some of this is my mess to begin with — flagging it now while the change is still unreleased and cheap to correct.

Schema references below are [newton-usd-schemas v0.4.0](https://github.com/newton-physics/newton-usd-schemas), the version `src/experimental/usd/CMakeLists.txt` now pins.

---

### 1. `newton:mimicCoef0` is authored in degrees, read as radians

The schema declares angle-valued units:

> `newton:mimicCoef0` — *"Offset added after scaling the leader joint's position/angle. […] Units: distance or degrees (matches the joint type for single-DOF joints)"*

`plugin/usd_decoder/usd_decoder.cc:2166-2172` assigns it straight into `mjEQ_JOINT` `data[0]`, which is radians for hinge and ball joints:

```cpp
if (newton_coef0 && newton_coef0.HasAuthoredValue()) {
  float val;
  newton_coef0.Get(&val);
  eq->data[0] = val;        // no degrees -> radians conversion
}
```

For an angular follower the offset is therefore wrong by a factor of 180/π ≈ 57.3.

Two details make this bite harder than the others:

- The `newton:` attribute takes **precedence** over `mjc:coef0` here, so authoring the deprecated attribute is not a workaround.
- `mujoco-usd-converter` writes `newton:mimicCoef0` with an explicit `np.degrees()` for hinge/ball followers and does not author `mjc:coef0` at all, so there is no fallback path for assets it produces.

Newton hit the identical bug on its own importer and fixed it in [newton-physics/newton#3728](https://github.com/newton-physics/newton/pull/3728).

`newton:mimicCoef1` is declared dimensionless and is correctly read as-is.

### 2. `newton:damping` is authored per-degree, read per-radian

> `newton:damping` — *"Units: effort * seconds / degrees (angular DOFs) or effort * seconds / distance (linear DOFs)"*

`usd_decoder.cc:1832-1836`:

[...]

PR #3477, opened 2026-08-13
Honor Newton mimic enabled state

Miko997:
## Summary

Read `newton:mimicEnabled` when importing Newton mimic constraints instead of using the follower joint's `physics:jointEnabled` value. The deprecated MJC equality path keeps its existing behavior.

Wire the existing Newton token definition into the standalone USD decoder target.

This addresses the `mimicEnabled` part of #3472. The other findings are outside this change.

## Testing

- pre-commit hooks
- `usd_decoder_plugin` with `MUJOCO_WITH_USD=ON`
- same target after merging current `main`


plugin/usd_decoder/usd_decoder.cc
havess, line 1: Something is weird with this file, git thinks it's binary on my end?

test/experimental/usd/plugins/decoder/CMakeLists.txt
havess, line 1: I appreciate adding tests! We have a suite of internal tests that I haven't exposed yet since they're in python. Can I ask you to split this change into the core fix + the tests in a follow up? Happy to verify the fix on our end, and defer testing to when we publicise our test suite.

havess (changes requested):

plugin/usd_decoder/usd_decoder.cc
Miko997: Fixed. I restored the file as text and reapplied only the core change.

test/experimental/usd/plugins/decoder/CMakeLists.txt
Miko997: Done. I removed the test target and fixture, so this PR is core fix only.

plugin/usd_decoder/usd_decoder.cc
```diff
@@ -2193,8 +2193,14 @@ void ParseConstraint(mjSpec* spec, const pxr::UsdPrim& prim, mjsBody* body,
     eq_joint_api.GetCoef3Attr().Get(&eq->data[3]);
     eq_joint_api.GetCoef4Attr().Get(&eq->data[4]);
 
-    pxr::UsdPhysicsJoint joint(prim);
-    ParseJointEnabled(eq, joint);
+    if (prim.HasAPI<pxr::MjcPhysicsEqualityJointAPI>()) {
```
havess: Can we push this logic into ParseJointEnabled?

havess (changes requested):

plugin/usd_decoder/usd_decoder.cc
Miko997: Done. `ParseJointEnabled` now handles the Newton-only mimic case while preserving the existing joint paths.

sylvesterkaczmarek (commented):
What should happen on a transition prim carrying both `NewtonMimicAPI` and `MjcPhysicsEqualityJointAPI`? Targets and coefficients in this parser prefer the Newton attributes and fall back to Mjc, but `ParseJointEnabled` does the opposite: it ignores `newton:mimicEnabled` whenever both APIs are present. Should enabled-state use the same Newton-first precedence?

Miko997:
Yes. Newton mimic state now takes precedence whenever NewtonMimicAPI is present, including transition prims with both APIs.

sylvesterkaczmarek (commented):
Rechecked the current diff. `ParseJointEnabled` now uses Newton-first precedence when `NewtonMimicAPI` is present and falls back to the regular joint attribute otherwise, which resolves the mixed-API precedence question I raised earlier. No blocker from my side on that point.

Miko997:
@havess, the requested logic is now in ParseJointEnabled, and the PR contains only the core fix. The mixed-API precedence concern has also been addressed and rechecked. Could you take another look and approve the pending CI runs when convenient?

Issue #3476, opened 2026-08-13
Box-box collider returns a phantom deep contact (dist = -box_width) for boxes resting face-to-face at near-zero penetration


kmoy007:
### Intro

Hi!

I am an engineer at Leap Automation, where we use MuJoCo to build digital twins of vision-guided pick-and-place machines for food packing.

### My setup

MuJoCo 3.10.0 and 3.11.0 (both reproduce), Python bindings, macOS arm64 (Apple Silicon). Default solver options; margin/gap are 0. Reproduces with and without `nativeccd` and `multiccd` (box-box pairs take the analytic primitive collider in all combinations).

### What's happening? What did you expect?

The box-box collider returns a **phantom deep contact** for two identical boxes resting face-to-face at near-zero penetration: alongside the correct contact (`dist = -0.056` mm at a corner), the manifold contains a second point with **`dist = -71.056`** — the full box width along the flush axis **plus** the real depth — with the same +y normal. The true configuration is two boxes whose faces kiss with ~0.06 mm overlap at one corner; a 71 mm overlap along the contact normal is geometrically impossible for these poses.

Consequences in a full simulation: the constraint solver resolves the phantom penetration by ejecting the two resting bodies at ±17.9 m/s in a single 0.5 ms step. In our digital twin this detonated a settled 3×3 stack of packed products inside a tray (products at rest, zero velocity, zero applied force for seconds beforehand) — the stacked rows had drifted into flush face-to-face contact, which is a resident configuration for packed goods.

The degenerate branch appears to be a **point singularity**: perturbing either box by ±10 µm in x, y, or z, or ±0.1° of yaw — in *any* direction — yields correct contacts (either none, or shallow ~-0.06 depths). Deeper real penetration (-0.01 mm and beyond) is also handled correctly. Resting stacks still find the branch because a settled body's qpos is frozen (qvel exactly zero) while its quaternion creeps at ~1e-13/substep, sliding the pair along the singular manifold until a step lands on it; in our sim the same pair hit it three times in 7 seconds.

The phantom depth is exactly `real_depth - 2*size[1]` (the flush axis' full extent), which smells like the face-clipping code measuring depth against the wrong (far) face for one manifold point when the boxes are near-identical in pose (relative yaw here is ~0.5°, bottom faces coplanar, side edges flush to within 0.07 mm).

[...]

PR #3478, opened 2026-08-13
Add phantom box contact regression

Miko997:
Adds the exact box pose from #3476 as a regression test for the separating-axis depth bound.

Validation:
- fails before `8655446f` with depth `-71.055559939...`
- focused regression passes on current `main`
- `engine_collision_box_test`: 10 passed
- pre-commit hooks pass


yuvaltassa:
Thanks, I don't think we need this.

PR #3486, opened 2026-08-16
Reject negative `sigma1` in LuGre bristle model

giusenso:
## Summary
Reject non-finite or negative LuGre sigma1 (see https://github.com/google-deepmind/mujoco/issues/3487). Negative values correspond to anti-damping, which causes the friction model to inject mechanical energy rather than dissipate it.
Validation covers both `<dcmotor>` and raw `dyntype="dcmotor"` configurations, including inherited defaults. Documentation is updated accordingly.

## Testing
Added C API and MJCF compilation tests for negative, non-finite, inherited, and raw dynprm values.


yuvaltassa:
Thanks, but I closed #3487 as WAI: MuJoCo is liberal with parameter values throughout and I don't think LuGre should be special. The line we do enforce is structural: a configuration that's meaningless (a gain with no input to act on, or a τ→V map with K=0) is rejected; a value that's merely unphysical is yours to choose.

Separately, this branch has picked up unrelated changes (engine_collision_box.c, the sensor test and contact_net.xml), so it isn't mergeable as-is. Closing.

PR #3488, opened 2026-08-16
Fix LuGre passivity in DC motor doc

giusenso:
Correct the LuGre passivity description in the DC motor technical documentation.
Since `tau` is defined as the torque applied to the mechanism (eq. 9b of the pdf), the passive resisting effort is `-tau`. The documented map is therefore corrected from `omega -> tau` to `omega -> -tau`.


giusenso:
For a more formal proof

## Proof

The document defines the torque applied to the mechanism as

$$
\tau=-(\sigma_0z+\sigma_1\dot z+\sigma_2\omega).
$$

Let $F=-\tau$ be the resisting friction effort and

$$
W(z)=\frac{\sigma_0}{2}z^2
$$

the stored bristle energy. Using

$$
\dot z=\omega-\sigma_0\frac{|\omega|}{g(\omega)}z,
$$

we obtain

$$
F\omega-\dot W = \sigma_2\omega^2 \quad+\sigma_1\omega^2 \left(1-\frac{\sigma_0z\,\mathrm{sgn}(\omega)}{g(\omega)}\right)\quad+\frac{|\omega|}{g(\omega)}(\sigma_0z)^2.
$$

For the standard LuGre domain, $\sigma_1\ge0$, $\tau_s\ge\tau_c>0$, $|\sigma_0z|\le\tau_s$, and $\tau_c\le g(\omega)\le\tau_s$. Therefore,

$$
F\omega-\dot W
\ge
\left(
\sigma_2-\sigma_1\frac{\tau_s-\tau_c}{\tau_c}
\right)\omega^2.
$$

If

$$
\rho=
\sigma_2-\sigma_1\frac{\tau_s-\tau_c}{\tau_c}>0,
$$

then

$$
\dot W\le(-\tau)\omega-\rho\omega^2,
$$

which is the input-strict passivity inequality for the map
$\omega\mapsto-\tau$.

PR #3493, opened 2026-08-20
Validate legacy MSH headers before parsing

CoffeeDrivenCoder:
## Summary
- read the four legacy MSH header counts together and reject truncated headers
- reject negative counts before passing them to NumPy data reads
- report the actual expected texcoord count instead of the literal placeholder
- add regression coverage for empty, truncated, negative-count, and incomplete-texcoord files

Previously, empty or truncated files leaked an internal `IndexError`, while negative counts reached `np.fromfile` where `-1` has special read-to-EOF semantics. These inputs now fail consistently with a descriptive `ValueError`.

## Testing
- `msh2obj_test.py`: 4 tests passed, including the existing MSH-to-OBJ model equivalence test
- `pyink --check` on both changed Python files
- `isort --check-only` on both changed Python files
- `git diff --check`

### DIFF
--- a/python/mujoco/msh2obj.py
+++ b/python/mujoco/msh2obj.py
@@ -42,10 +42,16 @@ class Msh:
       raise FileNotFoundError(f"{file} does not exist.")
 
     with open(file, "rb") as f:
-      nvertex = np.fromfile(f, dtype=np.int32, count=1)[0]
-      nnormal = np.fromfile(f, dtype=np.int32, count=1)[0]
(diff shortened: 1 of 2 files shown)


CoffeeDrivenCoder:
Hi maintainers, a quick follow-up on the CI for this PR.

The fork-triggered build and lint workflows were awaiting approval and have now expired after 30 days without running any jobs. Could someone with write access please approve and re-trigger the workflows when convenient? I am happy to update the branch if a new commit is needed to start a fresh run.

Thank you!

PR #3495, opened 2026-08-21
Fix flex contact filtering compaction

### LINKED ISSUE #3297: filterFlexContacts can keep the wrong final contact after FPS compaction
I think there may be a bug in `filterFlexContacts` introduced in `8ba68cee` (`Limit the number of flex contacts per collision pair`).

The short version is that the FPS selection loop can end up keeping a different final contact than the one it actually selected.

The relevant code is here:
- `src/engine/engine_collision_driver.c:401`
- current `main` at `5d782a2b`

What I think is happening:

- `filterFlexContacts` selects contacts with farthest-point sampling.
- During the loop it swaps `contacts[best]` into the kept prefix at `contacts[nselected]`.
- But that swap only happens when `nselected < mjMAXCONPAIR - 1`.
- On the final selected contact, the function increments `nselected` and then truncates `d->ncon`, without compacting `contacts[best]` into the kept prefix first.

So the last selected contact can be dropped, and the truncated prefix can contain a different contact instead.

A small toy translation of the loop reproduces that behavior: the selected sequence is `{A, B, C}` but the kept prefix after compaction/truncation becomes `{A, B, D}`.

I think there is also a second invariant problem here: because `contacts[]` is swapped in place during FPS, `selected[]` / `min_dist[]` no longer stay attached to the same logical candidate after the first swap. But even without leaning on that, the missing final compaction step already looks sufficient to make the retained set wrong.

Expected behavior:
- the first `nselected` contacts after filtering should be exactly the contacts selected by FPS.

Actual behavior:
- the final kept prefix can differ from the selected set.

If this diagnosis looks right, I'm happy to send a small PR with a regression test.

rootkiller6788:
Fixes #3297.

`filterFlexContacts` limits the number of contacts kept per flex collision pair using farthest-point sampling. The selection loop swapped `contacts[best]` into the kept prefix while the loop was still running, so the `selected[]` / `min_dist[]` bookkeeping (which is indexed by array position) no longer referred to the same logical contacts after the first swap. In practice this can re-select an already-picked contact (and skip a valid one), and because the swap was intentionally skipped for the final selected contact (when `nselected == mjMAXCONPAIR - 1`), the truncated prefix could contain a different, unselected contact instead.

This change keeps the selected contacts in a temporary buffer and compacts them into the contact prefix only after the selection loop finishes, so the first `nselected` contacts after filtering are exactly the contacts selected by farthest-point sampling.

Verification:
- `engine_collision_driver.c` compiles cleanly (single-file `gcc -c` against the in-repo headers).
- A standalone translation of the loop logic reproduces the discrepancy on the original code (the kept prefix differs from the selected set and contains duplicate selections) and shows the fixed logic keeps exactly the selected set.

### DIFF
--- a/src/engine/engine_collision_driver.c
+++ b/src/engine/engine_collision_driver.c
@@ -455,6 +455,7 @@ static void filterFlexContacts(mjData* d, int ncon_before) {
   mj_markStack(d);
   mjtByte* selected = mjSTACKALLOC(d, n, mjtByte);
   mjtNum* min_dist = mjSTACKALLOC(d, n, mjtNum);
+  mjContact* kept = mjSTACKALLOC(d, mjMAXCONPAIR, mjContact);
   memset(selected, 0, n);
 
   for (int i = 0; i < n; i++) {
@@ -472,8 +473,15 @@ static void filterFlexContacts(mjData* d, int ncon_before) {
     }
   }
 
+  // Select contacts by farthest point sampling.  The selected contacts are
+  // collected in `kept` and compacted into the contact prefix only after the
+  // selection loop finishes: swapping contacts in place during the loop would
+  // leave `selected`/`min_dist` indexed against a permuted array and corrupt
+  // the selection (and also drop the last selected contact when its index is
+  // not moved into the prefix).
   while (nselected < mjMAXCONPAIR && best >= 0) {
     selected[best] = 1;
+    kept[nselected] = contacts[best];
     mjtNum* bestpos = contacts[best].pos;
 
     int nextbest = -1;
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
This fixes the indexing problem cleanly. `selected` and `min_dist` now stay tied to the original contact array for the whole farthest-point loop, and compaction only happens after selection is finished, so the retained prefix matches the actual selected set, including the final pick. No blocker from me.

sylvesterkaczmarek (commented):
I think this needs a repository regression test. The bug is specifically in selection/compaction bookkeeping, but the PR currently validates it only with a standalone translation of the loop, so CI won't protect it later. A crafted case with more than `mjMAXCONPAIR` candidates should assert that the retained prefix is exactly the farthest-point selected set, including the final selected contact.

rootkiller6788:
Thanks, you're right that the fix shouldn't ship without CI coverage for the bookkeeping. I added a regression test to test/engine/engine_collision_driver_test.cc (commit 78aad81). It puts an 8x8 flex grid (64 vertices) just above a plane with midphase disabled, so the flex-plane pair goes through the direct narrowphase branch and all 64 vertices become candidates. The test reconstructs those candidates from the engine's own vertex positions using the same contact-point expressions as mj_collidePlaneFlex, replays the farthest-point selection, and asserts that the 50 retained contacts match the selected set one by one, in selection order — including the final pick that the old code dropped. Please take another look when you get a chance.

PR #3501, opened 2026-08-22
Fix BigInt/mjtSize WASM bindings and material transparency in demo viewer (#3496)

### LINKED ISSUE #3496: [mujoco stage: geom[0] texture] can't convert BigInt to number
### Intro

Hi!
 https://github.com/google-deepmind/mujoco/tree/main/wasm - throws an error in the viewer (not in console)
can't convert BigInt to number

### My setup

https://github.com/google-deepmind/mujoco/tree/main/wasm , React app integration

Setting `rgba="1 0 0 1"` renders as **green**, I expected it to render as **red**.

Here is a screen-shot showing a green sphere:

### Steps for reproduction

1. Load the model below.
2. Run the code below.
3. See green sphere (should be red).

<details>

```XML

<mujoco>
  <worldbody>
    <light pos="0 0 1"/>
  </worldbody>
</mujoco>

```
</details>

```python
import mujoco
import mediapy as media
data = mujoco.MjData(model)

with mujoco.Renderer(model) as renderer:
  renderer.update_scene(data)

  media.show_image(renderer.render())
```

### Confirmations

Adityakk9031:
Fixes #3496

## Root Cause

`MjModel.tex_adr` is `mjtSize*` (`int64_t*`). Emscripten's `typed_memory_view` over a 64-bit integer pointer surfaces as a **`BigInt64Array`** in JavaScript. Any arithmetic on its elements (e.g. `tex_adr[i] * 3` to compute byte offsets) throws:

- Chrome: `Cannot mix BigInt and other types, use explicit conversions`
- Firefox: `can't convert BigInt to number`

This is the root cause of the viewer error `[mujoco stage: geom[0] texture] can't convert BigInt to number`: the texture staging code reads `tex_adr`, hits `BigInt`, and the material setup aborts — leaving the geom with the wrong color.

Scalar `mjtSize` fields already avoid this by casting to `int` in the code generator. This PR extends the same policy to **pointer (array) fields**.

### 1. `wasm/codegen/generators/structs.py` — Binding generator (root cause fix)
When generating an accessor for a dynamically-sized pointer field whose inner type is `mjtSize`, emit an `Int32Array` copy instead of a `BigInt64Array` typed_memory_view. The WASM heap is 32-bit addressed so values always fit an `int`.

### 2. `wasm/codegen/generated/bindings.h` — Regenerated output
`tex_adr()` now returns an `Int32Array` copy, consistent with `tex_height`, `tex_width`, and `tex_nchannel`.

### 3. `wasm/tests/bindings_test.ts` — Test coverage
Added assertions that:
- `tex_adr` is an `Int32Array` (not `BigInt64Array`)
- Arithmetic on `tex_adr[0]` does not throw

### 4. `wasm/demo_app/app.ts` — Defensive consumer-side fixes
- Explicit `Number()` coercion on all `mjvGeom` fields (`type`, `dataid`, `size`, `rgba`, `mat`, `pos`) and `geoms.size()` / `mjData.time`
- Removed duplicate `geoms.get(i)` call (memory leak)
- Fixed `transparent` condition: `rgba[3] < 1.0` instead of `rgba[3] !== 0`
- Material color/transparency now updates dynamically every frame

## Before / After

```js
// Before
model.tex_adr          // BigInt64Array [ 0n, 786432n ]
model.tex_adr[1] * 3   // throws: can't convert BigInt to number

// After
model.tex_adr          // Int32Array [ 0, 786432 ]
model.tex_adr[1] * 3   // 2359296
```

### DIFF
--- a/wasm/codegen/generated/bindings.h
+++ b/wasm/codegen/generated/bindings.h
@@ -5263,7 +5263,12 @@ struct MjModel {
     return emscripten::val(emscripten::typed_memory_view(ptr_->ntex, ptr_->tex_nchannel));
   }
   emscripten::val tex_adr() const {
-    return emscripten::val(emscripten::typed_memory_view(ptr_->ntex, ptr_->tex_adr));
+    int size = ptr_->ntex;
+    emscripten::val result = emscripten::val::global("Int32Array").new_(size);
+    for (int i = 0; i < size; ++i) {
+      result.set(i, static_cast<int>(ptr_->tex_adr[i]));
+    }
+    return result;
   }
   emscripten::val tex_data() const {
     return emscripten::val(emscripten::typed_memory_view(ptr_->ntexdata, ptr_->tex_data));
--- a/wasm/codegen/generators/structs.py
+++ b/wasm/codegen/generators/structs.py
@@ -295,12 +295,29 @@ def _generate_field_data(
         array_size_str = parse_array_extent(extent, w, f.name)
 
       builder = code_builder.CodeBuilder()
-      with builder.function(f"emscripten::val {f.name}() const"):
-        builder.line(
-            "return"
-            f" emscripten::val(emscripten::typed_memory_view({array_size_str},"
-            f" {ptr_field_expr}));"
-        )
+      if inner_type_name == "mjtSize":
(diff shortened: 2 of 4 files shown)


Adityakk9031:
@yuvaltassa have a look

PR #3502, opened 2026-08-22
Expose mjtSize array fields as Int32Array in the WASM bindings

### LINKED ISSUE #3496: [mujoco stage: geom[0] texture] can't convert BigInt to number

VihaanAgarwal:
`MjModel.tex_adr` is `mjtSize*` (int64_t), and the generated binding exposes it through `typed_memory_view`, which surfaces it in JS as a `BigInt64Array`. Any arithmetic on its elements then throws: `Cannot mix BigInt and other types` in Chrome, `can't convert BigInt to number` in Firefox. That is the error behind #3496: the reporter's texture stage reads `tex_adr`, throws, and their material setup aborts, so the geom renders with the wrong color.

Scalar `mjtSize` fields already avoid this by casting to int in the generator, with a comment explaining that bigint does not fit the JS number model. This extends the same policy to pointer fields: the generator now emits an accessor that returns an `Int32Array` copy instead of a 64-bit view. The wasm heap is 32-bit addressed, so the values always fit. `tex_adr` is currently the only `mjtSize*` field in the public headers, so the regenerated `bindings.h` changes just that one accessor.

Before, against a two-texture model:

```
model.tex_adr          // BigInt64Array [ 0n, 786432n ]
model.tex_adr[1] * 3   // throws
```

After:

```
model.tex_adr          // Int32Array [ 0, 786432 ]
model.tex_adr[1] * 3   // 2359296
```

Added an assertion to the existing texture test that fails on main and passes with this change. Full wasm suite: 175 specs, 0 failures.

Fixes #3496

### DIFF
--- a/wasm/codegen/generated/bindings.h
+++ b/wasm/codegen/generated/bindings.h
@@ -5263,7 +5263,12 @@ struct MjModel {
     return emscripten::val(emscripten::typed_memory_view(ptr_->ntex, ptr_->tex_nchannel));
   }
   emscripten::val tex_adr() const {
-    return emscripten::val(emscripten::typed_memory_view(ptr_->ntex, ptr_->tex_adr));
+    int size = ptr_->ntex;
+    emscripten::val result = emscripten::val::global("Int32Array").new_(size);
+    for (int i = 0; i < size; ++i) {
+      result.set(i, static_cast<int>(ptr_->tex_adr[i]));
+    }
+    return result;
   }
   emscripten::val tex_data() const {
     return emscripten::val(emscripten::typed_memory_view(ptr_->ntexdata, ptr_->tex_data));
--- a/wasm/codegen/generators/structs.py
+++ b/wasm/codegen/generators/structs.py
@@ -295,12 +295,29 @@ def _generate_field_data(
         array_size_str = parse_array_extent(extent, w, f.name)
 
       builder = code_builder.CodeBuilder()
-      with builder.function(f"emscripten::val {f.name}() const"):
-        builder.line(
-            "return"
-            f" emscripten::val(emscripten::typed_memory_view({array_size_str},"
-            f" {ptr_field_expr}));"
-        )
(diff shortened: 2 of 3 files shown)


sylvesterkaczmarek (commented):
One API-semantics concern: `tex_adr` changes from a live WASM-backed `typed_memory_view` to a detached `Int32Array` copy. Other generated pointer fields still expose live views, so a write like `model.tex_adr[0] = ...` would now silently modify only the copy. If these bindings are meant to remain mutable, can we preserve the live-view behavior somehow, or explicitly make this field read-only and pin that contract in a test?

VihaanAgarwal:
Fair point on the semantics change. The mjtSize array fields are buffer offsets the model compiler computes (tex_adr, mesh_vertadr and friends), so writing them from JS was never a supported operation. The old live view allowed the write but threw on the read side the moment you did arithmetic, which is what #3496 hit. I see two clean options: pin the copy as explicitly read-only with a test, or keep a live view by teaching the generator to emit a wrapper that converts on access. I lean toward the first since these are derived quantities, but I will follow whatever the maintainers prefer here.

PR #3503, opened 2026-08-22
Fix DC motor Vmax voltage limiting in voltage input mode (#3489)

### LINKED ISSUE #3489: `Vmax` silently ignored for voltage-controlled dcmotors
### Intro

Hi! I am a robotics engineer looking into actuators modeling and friction models.

### My setup

* MuJoCo: 3.11.1 (3.11.0-105-g24b781ab)
* Architecture: x86-64, 64-bit
* OS: Ubuntu 24.04.4 LTS
* Kernel: Linux 7.0.0-28-generic

`Vmax` is silently ignored for voltage-controlled motors without an inductance/current state.

Consequences:

* Motor torque can exceed the configured supply-voltage capability.
* Simulated acceleration, contact forces, and energy consumption become unrealistic.
* Thermal-only motors are inconsistent: heating uses clamped voltage while torque uses unclamped voltage.
* Adding a small inductance unexpectedly changes behavior because stateful motors apply `Vmax`, while stateless motors do not.
* Controller tuning and safety validation based on `Vmax` become unreliable.

Example with `K=0.05`, `R=2`, `Vmax=10`, and `ctrl=20`:

Expected torque: `0.05 / 2 * 10 = 0.25`
Actual torque:   `0.05 / 2 * 20 = 0.50`

A separate `ctrlrange` or torque limit can mask the bug, but `Vmax` itself does not work as documented.

### Steps for reproduction

See tests in https://github.com/google-deepmind/mujoco/pull/3490.

### Confirmations

Adityakk9031:
Fixes #3489

### Problem
When a `<dcmotor>` is configured with `Vmax` in voltage-commanded mode (`input="voltage"`) without an inductance state:
1. `mjGAIN_DCMOTOR` skipped `dcmotorVoltage()` due to checking `ctrlspec != mjINPUT_VOLTAGE`, leaving `ctrl[uadr]` unclamped.
2. Inside `dcmotorVoltage()`, the raw terminal voltage input `u4[3]` was added after the `Vmax` clamp, bypassing the supply voltage limit.

### Solution
1. In `dcmotorVoltage()`, add raw terminal voltage `u4[3]` and apply the `Vmax` clamp to the resulting total terminal voltage.
2. In `mjGAIN_DCMOTOR`, invoke `dcmotorVoltage()` whenever `ctrlnum > 0` so that `ctrl[uadr]` is consistently clamped by `Vmax`.
3. Added unit test `TEST_F(DCMotorTest, VmaxVoltageInput)` to verify `Vmax` enforcement for voltage-controlled DC motors.

### DIFF
--- a/src/engine/engine_forward.c
+++ b/src/engine/engine_forward.c
@@ -263,8 +263,14 @@ static mjtNum dcmotorVoltage(const mjtNum* u, int spec, mjtNum length, mjtNum ve
     if (Vmax > 0) voltage = mju_clip(voltage, -Vmax, Vmax);
   }
 
-  // raw terminal voltage input: downstream of the controller, unclamped
-  return voltage + u4[3];
+  // raw terminal voltage input
+  voltage += u4[3];
+
+  // driver supply limit on total terminal voltage
+  mjtNum Vmax = gainprm[7];
+  if (Vmax > 0) voltage = mju_clip(voltage, -Vmax, Vmax);
+
+  return voltage;
 }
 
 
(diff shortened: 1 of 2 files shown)


Adityakk9031:
@yuvaltassa have a look

yuvaltassa:
This is working as intended: Vmax models the driver's limit on the controller output, and for a voltage-commanded motor the control is the voltage, so ctrlrange is the limit — I'd rather not have two mechanisms for one bound, especially since ctrlrange can be asymmetric. Clamping the sum would also silently clip a raw command that a user deliberately set outside the controller's supply limit.

The real problem here is the documentation, which I'm fixing in #3489. Closing this one.

Issue #3357, opened 2026-06-23
Allow running Simulate app with plugin libraries from any folder


scastro-nasa:
### The feature, motivation and pitch

Hi all -- I'm an engineer working with NASA on building out tools/applications with [`mujoco_ros2_control`](https://github.com/ros-controls/mujoco_ros2_control).

One thing that we have been working on is making new plugins, which works great with our custom entrypoint in `mujoco_ros2_control`.

However, we run into limitations running the `simulate` app standalone with our custom plugins if MuJoCo happens to be installed in a root-owned folder. This is due to the logic below, which assumes plugins are placed in a specific location which is tunable via preprocessor directive, but always is _relative to the mujoco executable directory_ -- meaning that users _have_ to copy their plugin libraries to this folder when they may not necessarily have permissions to do so:

https://github.com/google-deepmind/mujoco/blob/080668df42e2258b5c84f9630816f0a43bdcb562/simulate/main.cc#L189-L197

Would the maintainers be amenable to providing some kind of environment variable, command-line args, or other mechanism where users could specify additional paths to scan for plugins that may be anywhere on a user's file system? Or maybe we're unaware of an existing mechanism to do this.

We're happy to contribute changes, but wanted to get alignment from maintainers. Thanks for your help!

### Alternatives

Our only alternative, really, is what we have working now: Automation that tries to copy the built plugin libraries to the install location, but gives users a warning if there is a permissions-related failure or their `simulate` binary is not found, prompting them to manually do the copying if possible.

saran-t:
The reason that it's done this way was because I'm a bit uneasy about shopping a binary that's capable of silently running arbitrary code (it doesn't even need execute permission to do this) that's dumped in the user's local directory.

I have no objection in principle to you introducing a CMake option that injects additional lookup paths for custom builds. Free free to send a PR.

Do you strongly prefer environment variables to a compile time change?

scastro-nasa:
Thank you! I think we were looking for something that allowed directly using pre-installed mujoco binaries in addition to building from source, which is why I mentioned environment variables / CLI args. So unfortunately, we can't rely on a compile-time change in all our envisioned use cases.

But if there is a significant security concern that eliminates that option, it's good for us to know so we can explore alternatives.

saran-t:
Happy to brainstorm on this.

[...]

PR #3504, opened 2026-08-23
Allow a compile-time additional Simulate plugin directory

sylvesterkaczmarek:
## Summary

Add an opt-in CMake path that lets custom Simulate builds scan one additional plugin directory outside `${EXECDIR}/mujoco_plugin`.

This follows the approach discussed in #3357. The issue asked for runtime configurability, but the maintainer raised a security concern about allowing an environment variable or command-line value to make the distributed `simulate` binary silently load arbitrary DSOs, and explicitly indicated that a compile-time CMake option would be acceptable.

This change:
- adds `SIMULATE_ADDITIONAL_PLUGIN_DIR` as an empty-by-default CMake cache `PATH`
- injects the path only into custom builds where it is explicitly configured
- keeps the existing `${EXECDIR}/mujoco_plugin` scan unchanged
- scans the configured directory in addition to the built-in location
- does not add an environment variable or runtime command-line DSO search path

Example:
```sh
cmake -S simulate -B build \
  -DSIMULATE_ADDITIONAL_PLUGIN_DIR=/opt/my_mujoco_plugins
```

Addresses #3357.

## Validation

The final branch is based directly on current `main` and contains one focused commit rebased onto current `main`.

Final diff:
- `simulate/CMakeLists.txt`: 8 additions
- `simulate/main.cc`: 13 additions, 10 deletions

A full native Simulate build was not available in this execution environment, so no local build/test pass is claimed. The default build path is unchanged because the new CMake option is empty unless explicitly configured.

### DIFF
--- a/simulate/CMakeLists.txt
+++ b/simulate/CMakeLists.txt
@@ -39,6 +39,9 @@ endif()
 
 option(SIMULATE_BUILD_EXECUTABLE "Build the simulate executable binary." ON)
 option(SIMULATE_GLFW_DYNAMIC_SYMBOLS "Whether to resolve GLFW symbols dynamically." OFF)
+set(SIMULATE_ADDITIONAL_PLUGIN_DIR "" CACHE PATH
+    "Additional directory scanned by the simulate executable for plugin libraries."
+)
 
 # Check if we are building as standalone project.
 set(SIMULATE_STANDALONE OFF)
@@ -140,6 +143,11 @@ if(SIMULATE_BUILD_EXECUTABLE)
 
   add_executable(simulate main.cc array_safety.h ${SIMULATE_RESOURCE_FILES})
   target_compile_options(simulate PUBLIC ${MUJOCO_SIMULATE_COMPILE_OPTIONS})
+  if(SIMULATE_ADDITIONAL_PLUGIN_DIR)
+    target_compile_definitions(
+      simulate PRIVATE MUJOCO_ADDITIONAL_PLUGIN_DIR="${SIMULATE_ADDITIONAL_PLUGIN_DIR}"
+    )
+  endif()
   if(WIN32)
     add_custom_command(
       TARGET simulate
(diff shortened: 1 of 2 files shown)


yuvaltassa:
Meh, IDK. People that write their own plugins can tweak their own cmake?

Ping the original requester here and see if they are still enthusiastic about this?

sylvesterkaczmarek:
> Meh, IDK. People that write their own plugins can tweak their own cmake?
> 
> Ping the original requester here and see if they are still enthusiastic about this?

Hi @scastro-nasa, the maintainer asked whether this compile-time option would still be useful for your use case.

You previously mentioned that you also need to support preinstalled `simulate` binaries, so a compile-time path would not cover everything you originally wanted. Would this CMake option still materially help your workflow, or would you prefer that we close this and pursue a different approach?

scastro-nasa:
Tagging @eholum-nasa and @ndunkelb-nasa as well.

I think this would indeed address the workflow of building `simulate` + our own plugins from source, but not the binaries workflow.

Our path going forward with [`mujoco_ros2_control`](https://github.com/ros-controls/mujoco_ros2_control) is to distribute via binaries, and then additionally let people build their own plugins from source in additional CMake projects / possibly distribute them via binaries via the ROS ecosystem as well.

I wouldn't let this stop merging this PR, as it is helpful regardless for other users, but I also wouldn't object to finding an alternative that works for everyone if you have time to explore that.

sylvesterkaczmarek:
> Tagging @eholum-nasa and @ndunkelb-nasa as well.
> 
> I think this would indeed address the workflow of building `simulate` + our own plugins from source, but not the binaries workflow.
> 
> Our path going forward with [`mujoco_ros2_control`](https://github.com/ros-controls/mujoco_ros2_control) is to distribute via binaries, and then additionally let people build their own plugins from source in additional CMake projects / possibly distribute them via binaries via the ROS ecosystem as well.
> 
> I wouldn't let this stop merging this PR, as it is helpful regardless for other users, but I also wouldn't object to finding an alternative that works for everyone if you have time to explore that.

Thanks, that clarifies it. @yuvaltassa, this confirms the compile-time option is useful for source-built Simulate/plugin workflows, even though it does not solve binary distribution. I'd keep this PR scoped as-is and treat a binary-friendly mechanism as a separate follow-up if desired. Thanks a lot!

PR #3506, opened 2026-08-23
Fix swapped MJX ray empty-return types

sylvesterkaczmarek:
## Summary

When the ray filter excludes every geom, `mjx.ray` returns an integer in the distance slot and a float in the geom-id slot, contrary to its documented `(distance, geom id)` contract and its normal return path.

This change fixes the sentinel order, aligns the docstring, and adds a regression test asserting that the early-return dtypes match the normal path.

## Context

Closed, unmerged PR #3326 identified the same bug. Current `main` still contains it and no active replacement was found.

## Validation

- Based on current `main`
- Final diff changes only `ray.py` and `ray_test.py`
- Focused runtime validation was not run in this environment

### DIFF
--- a/mjx/mujoco/mjx/_src/ray.py
+++ b/mjx/mujoco/mjx/_src/ray.py
@@ -239,7 +239,7 @@ def ray(
     flg_static: bool = True,
     bodyexclude: Sequence[int] | int = -1,
 ) -> Tuple[jax.Array, jax.Array]:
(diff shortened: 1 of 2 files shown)


btaba (approved):

sylvesterkaczmarek:
@btaba thanks for taking the time to review and approve this. Much appreciated!

PR #3507, opened 2026-08-23
Restore CGL-to-GLFW fallback on macOS

sylvesterkaczmarek:
## Summary

On current `main`, Darwin always imports the CGL backend for classic rendering. If Apple's OpenGL framework does not expose `CGLSetCurrentContext`, that import fails even when the default backend selection could use GLFW.

This restores fallback only for default backend selection: explicit `MUJOCO_GL=cgl` still surfaces the CGL failure, while explicit `MUJOCO_GL=glfw` uses GLFW directly.

## Context

This revives the substance of closed, unmerged PRs #3312/#3313 with stricter explicit-backend semantics.

## Validation

- Based on current `main`
- Final diff changes one file
- macOS runtime validation was not run in this environment


haroonq (approved):

sylvesterkaczmarek:
Thanks for the review and merge. Much appreciated.

PR #3510, opened 2026-08-23
Fix msh2obj texcoord error formatting

sylvesterkaczmarek:
Fix the malformed texcoord-count diagnostic in `mujoco.msh2obj` by adding the missing f-string prefix. Invalid legacy MSH files now report the expected count instead of the literal `{ntexcoord}` placeholder.

The diff is one line in `python/mujoco/msh2obj.py`. The separate regression test was removed as requested in review.

This retains the fix from closed, unmerged PR #3302. GitHub CI passes on the current PR head; no local MuJoCo test run is claimed.


kevinzakka:
If you continue to spam the repo, you will be blocked @sylvesterkaczmarek

python/mujoco/msh2obj_error_test.py
yuvaltassa, line 1: we don't need the test, the fix is trivial

yuvaltassa (changes requested):

davidhozic:
> If you continue to spam the repo, you will be blocked @sylvesterkaczmarek

@kevinzakka @yuvaltassa 

These accounts are not controlled by humans, but by LLM bots. I doubt a human even reviews the code considering how many PRs they open. I know I don't really have a say in this but I'd urge you to create some agent instructions file (AGENTS.md) that gives these bots false directions so that they don't create slop like this or to include some specific "promp-injection" phrase so that the bots clearly identify themselves.

I really don't get what people have to gain by spamming open source projects with autonomous bots. I'm not really against people using agents when properly used, but such use is really bad. 

image

yuvaltassa:
On the one hand I agree, OTOH they are actually finding some things 🤷‍♂️

We are working out how to deal with this. I'm not sure what you're proposing is the way to go, but thanks for the suggestion.

Your humanity is appreciated @davidhozic 😇

python/mujoco/msh2obj_error_test.py
sylvesterkaczmarek: Yep, fair enough. Removed the test. Thanks a lot.

sylvesterkaczmarek:
> On the one hand I agree, OTOH they are actually finding some things 🤷‍♂️
> 
> We are working out how to deal with this. I'm not sure what you're proposing is the way to go, but thanks for the suggestion.
> 
> Your humanity is appreciated @davidhozic 😇

Just to clarify, I am a real person, not an autonomous bot. Like most people these days, I do use AI to help me test/verify and prepare patch descriptions, but I decide what to implement and submit, and I review and manage the PRs myself.

I understand the concern about volume though. Kevin raised it first and David echoed the same concern, so if having several PRs open at once creates too much review overhead for you @yuvaltassa, I'll slow down and be more selective about what I submit here. Thanks.

kevinzakka:
@sylvesterkaczmarek you were submitting hundreds of PRs a day, totally believable you were  reviewing and managing the PRs yourself. And on top of that, you made your github activity private which is not suspicious at all ;)

PR #3511, opened 2026-08-23
Accept NumPy integers for rollout sizing

sylvesterkaczmarek:
## Summary

Allow `rollout` to accept NumPy integer scalar values for `nstep` and `chunk_size`.

These values commonly come from NumPy computations. `np.int64(3)` is a valid integral value but is not an instance of Python's built-in `int`, so current validation rejects it before the rollout reaches the native implementation.

This change:
- accepts `np.integer` alongside Python `int` for `nstep`
- accepts `np.integer` alongside Python `int` for `chunk_size`
- adds focused regression coverage for both arguments
- keeps all existing dimension and value checks unchanged

This revives the still-relevant fix from closed, unmerged PR #3300 against current `main`.

## Validation

The final branch is based directly on current `main` and contains one focused commit changing only `python/mujoco/rollout.py` plus a regression test.

The MuJoCo Python extension was not built in this execution environment, so the new runtime regression test was not executed locally and no local test pass is claimed.


python/mujoco/rollout_numpy_integer_test.py
yuvaltassa, line 1: I don't think we need the test, though thanks for adding it.

Delete and i will approve

sylvesterkaczmarek: Done, thanks. I've removed the dedicated test as requested, so the PR diff is now limited to the two `np.integer` type-check changes in `rollout.py`.

PR #3513, opened 2026-08-24
Avoid sqrt of negative discriminants in MJX ray quadratics

sylvesterkaczmarek:
## Summary

Prevent `_ray_quad` from taking `sqrt` of a negative discriminant before no-intersection masking is applied.

The current code computes `jp.sqrt(det)` first and only afterward replaces results with infinity when `det < mjMINVAL`. For negative discriminants, the square root therefore produces NaNs unnecessarily.

Clamp only the square-root input to zero:

`jp.sqrt(jp.maximum(det, 0))`

The existing `det < mjMINVAL` condition still controls the returned no-intersection result, so positive-discriminant behavior and the existing threshold are unchanged.

Per maintainer feedback, the regression test was removed. The PR diff is now just the one-line _ray_quad change. GitHub Actions build and lint checks pass.

The original diagnosis was reported in closed, unmerged PR #2895 by @zhangbaozhe. This is a narrower reimplementation on current `main`.

## Validation

Final branch is based on current `main`, with one implementation line and one focused regression test.

The MJX runtime test was not executed locally because the required MuJoCo/JAX Python environment is not installed here.


mjx/mujoco/mjx/_src/ray_quad_test.py
yuvaltassa, line 1: We don't need the test (though thanks for adding it initially!)

Delete the test and I will approve.

sylvesterkaczmarek: Done, thanks. I've removed the test as requested, so the PR diff is now just the one-line `_ray_quad` change. Lint is green and the build is currently running.

fallenmi (approved):
I reviewed exact head `03c21c5dc457fc95cb4ee10a9c7b35f48cda7c48` and the current GitHub merge `5e1f9967c8772a828b71652506c399bc9a945ba1`.

The clamp is placed correctly: it changes only the square-root operand, while the existing `det < mjMINVAL` mask remains the authority for returning no intersection. I ran the exact production `_ray_quad` across 250,000 deterministic float32 coefficient triplets, including 55,276 negative discriminants, zero denominators, and positive/double-root controls. Base, head, and current merge produced the same output digest with zero NaN roots; the head and merge StableHLO add `maximum(det, 0)` immediately before `sqrt`, while the base does not. Thus the invalid intermediate is removed without changing finite roots, the no-hit `inf` contract, or threshold behavior.

The exact-head ray suite passed 11/11, the current merge passed 12/12 (including a newer main-branch contract test), and `git diff --check` is clean for both the submitted and current-merge deltas. All exact-head build and lint checks are green. I found no blocking issue.

Disclosure: I used OpenAI Codex to inspect the exact revisions, repository policy and interaction history, run the production JAX/StableHLO oracle and ray suites, and draft this review. I verified the evidence and conclusion.

PR #3514, opened 2026-08-24
Fix strict C11 compilation of mjsan.h

sylvesterkaczmarek:
## Summary

Make the AddressSanitizer stack instrumentation in `mjsan.h` compile under GCC's strict C11 mode.

The sanitizer-only definitions currently use two constructs rejected by the project's `-std=c11 -Wpedantic -Werror` configuration:
- bare `asm`, which is not a C11 keyword
- `__attribute__((always_inline))` after the declarator on a function definition, which GCC rejects

This moves `always_inline` before the definition and uses `__asm__ volatile` for the memory barriers. The generated semantics are unchanged, and the non-sanitizer path is untouched.

This revives the still-relevant fix from closed, unmerged PR #3458 by @teerthsharma, rebased onto current `main`.

## Validation

Focused GCC 14.2.0 compile passed:

`cc -std=c11 -Wpedantic -Werror -fsanitize=address -c <minimal translation unit including mjsan.h>`

Final diff: one file, six line substitutions, one commit.

A full MuJoCo sanitizer build was not run in this environment.

### DIFF
--- a/include/mujoco/mjsan.h
+++ b/include/mujoco/mjsan.h
@@ -55,17 +55,17 @@ extern "C" {
 #endif
 
 void mj__markStack(mjData*) __attribute__((noinline));
-static inline void mj_markStack(mjData* d) __attribute__((always_inline)) {
-  asm volatile("" ::: "memory");
+__attribute__((always_inline)) static inline void mj_markStack(mjData* d) {
+  __asm__ volatile("" ::: "memory");
   mj__markStack(d);
-  asm volatile("" ::: "memory");
+  __asm__ volatile("" ::: "memory");
 }
 
 void mj__freeStack(mjData*) __attribute__((noinline));
-static inline void mj_freeStack(mjData* d) __attribute__((always_inline)) {
-  asm volatile("" ::: "memory");
+__attribute__((always_inline)) static inline void mj_freeStack(mjData* d) {
+  __asm__ volatile("" ::: "memory");
   mj__freeStack(d);
-  asm volatile("" ::: "memory");
+  __asm__ volatile("" ::: "memory");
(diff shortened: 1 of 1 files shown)


yuvaltassa:
Can you confirm this works under the full matrix of supported compilers / arches? 

Unfortunately we don't have ASAN running in CI (perhaps we should?)... hmmm.

I feel like this calls for a pre-PR. The issue with ASAN is that it's slow, we definitely don't want to run everything under ASAN. However bulding with ASAN is not terribly slow.

How about we build the library with ASAN and just run a small number of tests which are known to be fast (pipeline_test ?)

The main thing I'd like to see is is CI timing before/after

sylvesterkaczmarek:
> Can you confirm this works under the full matrix of supported compilers / arches?
> 
> Unfortunately we don't have ASAN running in CI (perhaps we should?)... hmmm.
> 
> I feel like this calls for a pre-PR. The issue with ASAN is that it's slow, we definitely don't want to run everything under ASAN. However bulding with ASAN is not terribly slow.
> 
> How about we build the library with ASAN and just run a small number of tests which are known to be fast (pipeline_test ?)
> 
> The main thing I'd like to see is is CI timing before/after

Thanks. Since this comment, the same strict-C11 fix has landed directly in `main` as `7eb779a33bae4eef7a874584706d069ae7027642`.

That supersedes this PR, so I'll close it rather than add CI work around a fix that is already upstream. The ASAN CI idea still seems useful as a separate improvement if you want it pursued independently. :) Please let me know @yuvaltassa

yuvaltassa:
The problem with ASAN in CI is that it's slooooooow

PR #3515, opened 2026-08-24
Fix Unity hinge reference import for radian models

sylvesterkaczmarek:
## Summary

Fix Unity import of hinge joint `ref` values when the MJCF compiler uses radians.

`MjHingeJoint` already converts hinge ranges from radians to the degree-valued Unity fields, but `ref` was copied directly into `Configuration`, which is also stored and exposed in degrees. A model using `angle="radian"` therefore imported the reference angle with the wrong unit.

This change:
- converts hinge `ref` from radians to degrees when `MjSceneImportSettings.AnglesInDegrees` is false
- preserves existing degree-mode behaviour
- adds focused regression coverage using a pi/2 reference and checking for 90 degrees

## Validation

The branch is based directly on current `main`, is one commit ahead, and changes only `MjHingeJoint.cs` and its existing test file.

The Unity EditMode suite was not executed in this environment, so no local Unity test pass is claimed.

### DIFF
--- a/unity/Runtime/Components/Joints/MjHingeJoint.cs
+++ b/unity/Runtime/Components/Joints/MjHingeJoint.cs
@@ -76,6 +76,9 @@ namespace Mujoco {
       }
 
       Configuration = mjcf.GetFloatAttribute("ref", 0.0f);
+      if (!MjSceneImportSettings.AnglesInDegrees) {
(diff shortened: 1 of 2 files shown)


Balint-H (approved):
This is a necessary fix, lgtm.

PR #3516, opened 2026-08-24
Keep Unity slide spring references in linear units

sylvesterkaczmarek:
## Summary

Fix Unity import of `springref` for slide joints when the MJCF compiler uses radians.

The shared `SpringSettings.FromMjcf` path converted every `springref` from radians to degrees whenever `MjSceneImportSettings.AnglesInDegrees` was false. That is correct for angular joints, but a slide joint's `springref` is a linear position and must remain in metres.

This change:
- makes the shared joint-settings parser distinguish angular from linear spring references
- keeps the existing angular conversion as the default
- has `MjSlideJoint` explicitly request linear spring-reference handling
- adds regression coverage showing `springref="0.25"` remains `0.25` in a radian-model import

## Validation

Rebased onto current `main`. The diff remains limited to the shared joint settings, `MjSlideJoint`, and its existing test file. Pre-commit and change-scanner checks pass on the refreshed branch.

The Unity EditMode suite was not executed locally, so upstream CI provides the full platform validation.


Balint-H (approved):
Thanks for fixing the end of file newline issue. LGTM

sylvesterkaczmarek:
Thanks @Balint-H for the review and approval. Appreciate you checking the final newline fix as well.

PR #3517, opened 2026-08-24
Serialize Unity hinge reference with invariant culture

sylvesterkaczmarek:
## Summary

Make Unity-generated hinge joint `ref` values locale-independent.

`MjHingeJoint.OnGenerateMjcf` already uses `MjEngineTool.MakeLocaleInvariant` for `range`, but formats `ref` with the current process culture. On locales with a comma decimal separator, for example `fr-FR`, a reference value such as 1.5 can therefore be written as `1,5`, which is not the intended MJCF numeric representation.

This change:
- serializes hinge `ref` through `MjEngineTool.MakeLocaleInvariant`
- adds a regression test that switches the current culture to `fr-FR` and verifies the generated value remains `1.5`
- does not change parsing or runtime joint behaviour

## Validation

The branch is based directly on current `main`, is one commit ahead, and changes only `MjHingeJoint.cs` and its existing test file.

The Unity EditMode suite was not executed in this environment, so no local Unity test pass is claimed.

### DIFF
--- a/unity/Runtime/Components/Joints/MjHingeJoint.cs
+++ b/unity/Runtime/Components/Joints/MjHingeJoint.cs
@@ -91,7 +91,7 @@ namespace Mujoco {
         throw new ArgumentException("Lower range value can't be bigger than Higher");
       }
       mjcf.SetAttribute("range", MjEngineTool.MakeLocaleInvariant($"{RangeLower} {RangeUpper}"));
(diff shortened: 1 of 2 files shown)


Balint-H (approved):
Simple minor tweak. Having a test just for this feels excessive, but not problematic.

Issue #3527, opened 2026-08-27
URDF <mimic> is dropped silently, while <dynamics> and <limit effort> in the same parser are honoured


omnilink-tech:
`src/xml/xml_urdf.cc` contains no handling for `<mimic>` -- `grep -ci mimic`
over the file returns 0, and a repo-wide search under `src/` finds it only in
two unrelated USD schema files.

The consequence is quiet. A parallel-jaw gripper URDF -- Robotiq 2F85/2F140,
Franka Hand, Schunk, and most multi-finger hands -- declares its coupled finger
joints with `<mimic>`, loads without error, and comes back with those joints as
independent uncontrolled DOFs. Nothing warns, so the first symptom is a gripper
that will not close symmetrically, or an `nq` that does not match expectations.

I want to be careful about what I am claiming, because "URDF is intentionally a
subset of MJCF" is a reasonable position and I would not argue with it in
general. What makes this one look like an inconsistency rather than a policy is
its immediate neighbours in the same file: `xml_urdf.cc:463` reads
`<dynamics>`, and `:485-490` maps `<limit effort>` onto `actfrcrange`. Those are
physical joint properties honoured a few lines from one that is discarded in
silence.

MuJoCo also already has the target construct -- `<equality><joint>` with
`polycoef` -- so a linear `<mimic>` is representable. Genesis performs exactly
that translation in its URDF importer, which is at least an existence proof that
the mapping is workable.

**The ask I would actually make is the smaller one: warn.** A single line naming
the dropped `<mimic>` joints at parse time removes the silence, costs nothing in
behaviour, and does not commit the parser to supporting a construct you may not
want. Translating to an equality is the larger option and I would not press for
it.

Happy to send either as a PR if there is appetite -- warning-only, or
warning plus the `polycoef` translation with tests. If the answer is "URDF is a
subset and that is deliberate", a note to that effect in the URDF extensions
section of `doc/modeling.rst` would still save the next person the debugging;
`mimic` currently appears nowhere in that document.

Context: found while importing gripper URDFs into another simulator, where the
same omission on our side produced a gripper whose fingers drifted apart under
asymmetric load. Ours warns now.

Abhayindia:
@omnilink-tech  I'd like to take this. Planning the warn-only version — a line in `mjXURDF::Joint` naming the dropped `<mimic>` joint at parse time — unless you'd prefer translating it to an `<equality><joint>` with polycoef, which is also doable. Which do you want?

[...]

PR #3530, opened 2026-08-28
translate URDF <mimic> joints to joint equality constraints

Abhayindia:
per #3527. urdf `<mimic multiplier="M" offset="O">` on a follower joint becomes an `<equality><joint>` with `polycoef = [O M 0 0 0]`. couplings are collected during joint parsing and emitted after, so the target joint can be declared later in the file.

also wired `xml_urdf_test.cc` into cmake - it wasn't referenced by any target so its tests weren't building.

### DIFF
--- a/src/xml/xml_urdf.cc
+++ b/src/xml/xml_urdf.cc
@@ -189,9 +189,21 @@ void mjXURDF::Parse(XMLElement*        root,
   }
 
   // <mimic> -> joint equality: q_follower - q0 = offset + multiplier*(q_driver - q0)
+  auto hinge_or_slide = [&](const std::string& jnt_name) {
+    mjsElement* el  = mjs_findElement(spec, mjOBJ_JOINT, jnt_name.c_str());
+    mjsJoint*   jnt = el ? mjs_asJoint(el) : nullptr;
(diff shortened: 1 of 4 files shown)


Abhayindia:
@googlebot I signed it!

Abhayindia:
added the guard - a mimic pointing at a fixed/planar/floating/missing joint now warns and is skipped rather than failing the load, with a test for the fixed-target case. thanks for catching that.

sylvesterkaczmarek (commented):
Could we make the mimic regression behavioral rather than only checking the emitted IDs and `polycoef`? The current test would still pass if the equality were wired in the wrong direction or interpreted incorrectly relative to the joint references. A small case with a nonzero driver value, `multiplier=-1`, and nonzero offset that checks the resulting equality residual/tracking would pin the actual URDF `q_follower = M*q_driver + O` semantics.

Abhayindia:
done - MimicBecomesJointEquality now simulates from a driver displacement (multiplier=-1, offset=0.02, damping so it settles) and checks q_right == -q_left + 0.02 to 1e-4, so it'd catch a wrong direction or a dropped multiplier/offset.

PR #3531, opened 2026-08-28
Fix thermal DC motor velocity derivatives

giusenso:
## Summary

DC motor forward dynamics use temperature-adjusted winding resistance, while the analytic velocity derivative uses the unadjusted configured resistance.

This PR makes the derivative use the same effective resistance as the forward model for physical motor terms. The controller’s torque-to-voltage conversion continues using configured resistance, matching current forward behavior.

## Motivation

The analytic velocity derivative can disagree with finite differences when thermal dynamics change winding resistance. This affects implicit integration and other consumers of qDeriv.

## Intent Question

This difference may be intentional (if so, just close this PR). MuJoCo sometimes uses approximate derivatives, and configured resistance may have been selected deliberately for stability or controller semantics.

The RST documentation mentions thermal resistance variation but does not specify whether:

* resistance is defined at reftemp.
* Analytic derivatives use configured or temperature-adjusted resistance.
* The controller compensates for temperature-adjusted resistance.

If configured-resistance linearization is intentional, the appropriate resolution may instead be a documentation update and explanatory source comment. If the controller should use temperature-adjusted resistance, a broader forward-model change may be needed.

This PR follows the narrower interpretation that analytic derivatives should match the existing forward implementation.

## Changes

* Compute winding resistance from the current temperature state.
* Use temperature-adjusted resistance for stateless back-EMF and stateful current derivatives.
* Retain configured resistance in the controller’s torque-to-voltage mapping.
* Leave motors without thermal dynamics unchanged.
* Add regression coverage comparing analytic and centered finite-difference derivatives for thermal stateless and stateful motors.


fallenmi (approved):
I reviewed exact head `969f85943f3f59f765083db0a5973b1edd6d04e2`.

The derivative now uses the same temperature-adjusted winding resistance as the forward actuator path, while correctly retaining the configured/nameplate resistance in the torque-to-voltage controller mapping. That preserves the existing controller semantics and makes the resulting damping scale by `R0 / R(T)` for stateless motors and by the same physical `R(T)` in the stateful current update.

I checked this with an independent central finite-difference oracle over eight configurations: raw-voltage and controller inputs, stateless and stateful current, preceding slew/integral activation slots, non-default ambient/reference temperatures, and a non-thermal control. Exact base disagrees in six thermal cases; this head and a clean merge into current `main` pass all eight with a worst absolute error of `1.6e-10`. The full owning derivative test binary passes 32/32 on the current merge, the three DC-motor tests pass three repetitions, `git diff --check` is clean, and all exact-head build/lint workflows are green.

Disclosure: I used OpenAI Codex to inspect and build the exact revisions, construct and run the finite-difference matrix, execute the focused/full tests, and draft this review. I verified the code paths, outputs, exact commits, CI, and conclusion.

PR #3536, opened 2026-08-30
Store broadphase pair indices as uint32_t to fix out-of-bounds read

### LINKED ISSUE #3535: Out-of-bounds read in sweep-and-prune collision broadphase with more than 32768 items
### Intro

Hi!

I am a student at Aalto university, I plan to use Mujoco for physical simulations and studying the speed/accuracy tradeoff of environments.

### My setup

MuJoCo 3.12.1 (`mjVERSION_HEADER 3012001`), built from source at commit `47ebfe48`. C API. Linux x86_64, Ubuntu 24.04.4 LTS, kernel 6.8.0-138-generic.

A 2D `flexcomp` grid with more than 32768 elements and `selfcollide="auto"` crashes the process with a segfault on the first `mj_step`. The model compiles with no error or warning. I expected either a successful step, or a compile-time error/limit if this element count isn't meant to be supported.
A 130×130 cloth crosses the threshold, so the input is large for MuJoCo simulation but not inherently malformed or absurd. The main impact is a non-documented, likely unintended size boundary and a non-informative crash.

Cause: `mj_SAP` packs a pair of item indices into a signed 32-bit `int`:

```c
// engine_collision_driver.c:1505 (mj_SAP)
pair[npair++] = (id1<<16) + id2;
```

`mj_SAP` only rejects `n >= 0x10000`, so `id1` up to 65534 is accepted. Once `id1 >= 32768`, `(id1<<16)` sets bit 31, so the packed `int` becomes negative. Two call sites decode the high half of the pair without masking it back down:

```c
// engine_collision_driver.c:1711 (body broadphase)
int bf1 = bfid[sappair[i] >> 16];

// engine_collision_driver.c:2381 (flex self-collision)
int e1 = elid[sappair[i] >> 16];
```

The arithmetic right shift sign-extends, so this reads `bfid[]`/`elid[]` at a negative index. A third call site decodes the exact same packing correctly, with a mask — this is the pattern the other two are missing:

```c
// engine_collision_driver.c:639 (also body broadphase, different loop)
int bf1 = (broadphasepair[i]>>16) & 0xFFFF;
```

In the flex path, the garbage value read from `elid[]` becomes an element id used directly in unchecked pointer arithmetic (`d->flexelem_aabb + 6*(m->flex_elemadr[f1] + e1)`, `m->flex_elem + m->flex_elemdataadr[f1] + e1*(dim1+1)`), which is what causes the segfault.

Two ways this element count is reached that I found:

- `mj_isElemActive` returns 1 for every element when `flex_dim < 3`, so a 2D flex has `nactive == flex_elemnum`. With `flex_elemnum > 32768` and `selfcollide="auto"`, collision routes to `mj_collideFlexSAP` (the flex does have a BVH; `auto` picks SAP over the BVH-tree path whenever `flex_dim != 3`).
- The same unmasked decode at `:1711` is also reachable with a plain model containing more than 32768 mutually-collidable bodies and no flex at all — I reproduced a segfault this way too, confirming the bug isn't flex-specific.

The compiler doesn't guard against this: it only checks `bodies + flexes < 65534` (`user_model.cc:4965`), and doesn't bound per-flex element counts at all.

Fix: mask both decode sites the same way `:639` already does — `(sappair[i] >> 16) & 0xFFFF`.

### Steps for reproduction

1. Load the model below with `mj_loadXML`.
2. Call `mj_makeData`, then `mj_step` once.
3. Process crashes with SIGSEGV (exit code 139) inside the first `mj_step` call.

Model compilation takes a few minutes.

<details>
<summary>poc_flex.xml</summary>

```XML
<mujoco model="big cloth">
  <compiler angle="radian" autolimits="true"/>
  <option timestep="0.001" integrator="implicitfast" solver="CG" cone="pyramidal"/>
  <size memory="3G"/>
  <worldbody>
    <!-- 130x130 grid, dim=2 -> 2*129^2 = 33282 elements (> 32768) -->
    <flexcomp type="grid" count="130 130 1" spacing=".01 .01 .01" pos="0 0 1"
              radius=".002" mass="1" name="cloth" dim="2">
      <contact selfcollide="auto" passive="true" solref="0.01 1" solimp=".95 .99 .0001"/>
      <elasticity young="2e4" poisson=".2" thickness="1e-3" elastic2d="both" damping="1e-2"/>
    </flexcomp>
  </worldbody>
</mujoco>
```

</details>

```cpp
#include <cstdio>
#include <mujoco/mujoco.h>

int main(int argc, char** argv) {
  char err[1024];
  mjModel* m = mj_loadXML(argv[1], nullptr, err, sizeof(err));
  if (!m) { std::printf("LOAD FAILED: %s\n", err); return 1; }
  std::printf("model compiled OK, flex_elemnum[0]=%d\n", (int)m->flex_elemnum[0]);
  std::fflush(stdout);

  mjData* d = mj_makeData(m);
  std::printf("calling mj_step...\n");
  std::fflush(stdout);

  mj_step(m, d);  // segfaults here

  std::printf("survived, ncon=%d\n", d->ncon);
  mj_deleteData(d);
  mj_deleteModel(m);
  return 0;
}
```

Build and run:

```
g++ -O2 -g -std=c++17 poc.cc -I<mujoco>/include \
    -L<mujoco>/build/lib -lmujoco -Wl,-rpath,<mujoco>/build/lib -o poc
./poc poc_flex.xml
```

With no sanitizer: prints up to `calling mj_step...`, then the process dies with SIGSEGV, exit code 139.

Under AddressSanitizer (library rebuilt with `-fsanitize=address -fno-omit-frame-pointer -g`):

```
==1291406==ERROR: AddressSanitizer: SEGV on unknown address 0x7bb1190f27f0
==1291406==The signal is caused by a READ memory access.
    #0 filterBox         engine_collision_driver.c:235
    #1 mj_collideElems   engine_collision_driver.c:2550
    #2 mj_collideFlexSAP engine_collision_driver.c:2383
    #3 mj_collision      engine_collision_driver.c:860
    #4 mj_fwdPosition    engine_forward.c:147
    #5 mj_forwardSkip    engine_forward.c:1794
    #6 mj_forward        engine_forward.c:1841
    #7 mj_step           engine_forward.c:1852
```

### Confirmations

ilmari99:
Fixes #3535

`mj_SAP` and `add_pair` both pack a pair of item indices into a 32-bit word as `(id1<<16) | id2`, but the buffers holding those words were signed `int`. Once `id1 >= 32768` the shift moves a set bit into the sign position, which is undefined behavior, and the packed value comes back out negative. Two call sites decode the high half with a plain arithmetic shift, which then sign-extends instead of recovering the original index:

```c
int bf1 = bfid[sappair[i] >> 16];   // body broadphase
int e1  = elid[sappair[i] >> 16];   // flex self-collision
```

producing a negative array index and an out-of-bounds read that reliably segfaults once a model has more than 32768 mutually-collidable bodies, or a 2D/1D flex with more than 32768 elements and `selfcollide="auto"`.

Masking the decode would fix the read but leave the shift that creates the bad value undefined, so this makes the packed representation `uint32_t` end-to-end instead. `mj_SAP` already rejects `n >= 0x10000`, so seems like the packing is meant to carry a full 16 bits per id.
The decode sites above are correct now that the buffers are unsigned, so they need no mask, and `uintcmp` no longer has to cast its operands to compare them.

Review found a second overflow in the same path: `n*(n-1)/2` is computed in `int` at three sites and overflows from `n = 46342`, although the result fits for every `n` that `mj_SAP` accepts. At the flex site this kills any 1D/2D flex with 46342–65535 elements and `selfcollide=auto`. The product is now computed in size_t.

A regression test is included, covering the flex self-collision path: a 46342-element `dim="1"` flex whose only colliding pair is last in the element list and leftmost in space, so element 46340 is packed into the high half. This covers both the original overflow, and the second overflow from `n*(n-1)/2`. It asserts the decoded element ids and the contact distance exactly, so a sign-extended index fails the test even where the out-of-bounds read does not fault. It fails on the parent commit and adds 0.8 s to the suite. The body broadphase path packs through the same line in `mj_SAP`, but reaching index 32768 there needs a model with more than 32768 collidable bodies, which is too slow to compile in a test.

Verified with UBSan, which on the old code with a 32770 element flex reports `left shift of 32768 by 16 places cannot be represented in type 'int'` at the pack site, and a follow-on `signed integer overflow` where the sign-extended index is scaled into an element offset. With the 46342-element test model it instead reports `signed integer overflow: 46341 * 46342` at the pair-count multiply, before reaching the packing.
Clean after the changes. Behavior and performance are unchanged: pairs keep the same unsigned sort order, `qpos`/`qvel` and contact output are identical across tests, and `mjTIMER_COL_BROAD` is unchanged on 100 humanoids (median 512.9 us, 3% spread, both before and after, over interleaved A/B runs).
Full suite passes (2091/2091).

### DIFF
--- a/src/engine/engine_collision_driver.c
+++ b/src/engine/engine_collision_driver.c
@@ -15,6 +15,7 @@
 #include "engine/engine_collision_driver.h"
 
 #include <stddef.h>
+#include <stdint.h>
 #include <string.h>
 
 #include <mujoco/mjdata.h>
@@ -618,7 +619,7 @@ void mj_collision(const mjModel* m, mjData* d) {
   // broadphase collision detector
   TM_START;
   int nmaxpairs = (nbodyflex*(nbodyflex - 1))/2;
-  int* broadphasepair = mjSTACKALLOC(d, nmaxpairs, int);
+  uint32_t* broadphasepair = mjSTACKALLOC(d, nmaxpairs, uint32_t);
   int nbfpair = mj_broadphase(m, d, broadphasepair, nmaxpairs);
   unsigned int last_signature = -1;
   TM_END(mjTIMER_COL_BROAD);
@@ -634,11 +635,11 @@ void mj_collision(const mjModel* m, mjData* d) {
 
   for (int i=0; i < nbfpair; i++) {
     // reconstruct bodyflex pair ids
-    int bf1 = (broadphasepair[i]>>16) & 0xFFFF;
+    int bf1 = broadphasepair[i] >> 16;
     int bf2 = broadphasepair[i] & 0xFFFF;
 
     // compute signature for this bodyflex pair
-    unsigned int signature = (bf1<<16) + bf2;
+    unsigned int signature = ((uint32_t)bf1<<16) + bf2;
     // pairs come sorted by signature, but may not be unique
     // if signature is repeated, skip it
     if (signature == last_signature) {
@@ -1350,7 +1351,7 @@ static void makeAAMM(const mjModel* m, mjData* d,
 
 // add bodyflex pair in buffer; do not filter if m is NULL
 static void add_pair(const mjModel* m, int bf1, int bf2,
-                     int* npair, int* pair, int maxpair) {
+                     int* npair, uint32_t* pair, int maxpair) {
   // add pair if there is room in buffer
   if ((*npair) < maxpair) {
     // contact filtering if m is not NULL
@@ -1394,9 +1395,9 @@ static void add_pair(const mjModel* m, int bf1, int bf2,
 
     // add pair
     if (bf1 < bf2) {
-      pair[*npair] = (bf1<<16) + bf2;
+      pair[*npair] = ((uint32_t)bf1 << 16) | (uint32_t)bf2;
     } else {
-      pair[*npair] = (bf2<<16) + bf1;
+      pair[*npair] = ((uint32_t)bf2 << 16) | (uint32_t)bf1;
     }
     (*npair)++;
   } else {
@@ -1432,9 +1433,9 @@ mjSORT(SAPsort, mjtSAP, SAPcmp);
 
 
 // given list of axis-aligned bounding boxes in AAMM (min[3], max[3]) format,
-// return list of pairs (i, j) in format (i<<16 + j) that can collide,
+// return list of pairs (i, j) in format (i<<16 | j) that can collide,
 // using sweep-and-prune along specified x axis (0, 1 or 2).
-static int mj_SAP(mjData* d, const mjtNum* aamm, int n, int axis_x, int* pair, int maxpair) {
+static int mj_SAP(mjData* d, const mjtNum* aamm, int n, int axis_x, uint32_t* pair, int maxpair) {
   // check inputs
   if (n >= 0x10000 || axis_x < 0 || axis_x > 2 || maxpair < 1) {
     return -1;
@@ -1500,7 +1501,7 @@ static int mj_SAP(mjData* d, const mjtNum* aamm, int n, int axis_x, int* pair, i
         }
 
         // add pair, check buffer size
-        pair[npair++] = (id1<<16) + id2;
+        pair[npair++] = ((uint32_t)id1 << 16) | (uint32_t)id2;
         if (npair >= maxpair) {
           return maxpair;
         }
@@ -1552,8 +1553,8 @@ static void updateCov(mjtNum cov[9], const mjtNum vec[3], const mjtNum cen[3]) {
 
 
 // comparison function for unsigned ints
-static inline int uintcmp(int* i, int* j, void* context) {
-  if ((unsigned) *i < (unsigned) *j) {
+static inline int uintcmp(uint32_t* i, uint32_t* j, void* context) {
+  if (*i < *j) {
     return -1;
   } else if (*i == *j) {
     return 0;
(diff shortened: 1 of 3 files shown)


sylvesterkaczmarek (commented):
The decode is fixed, but the same sign-bit case is still undefined at the packing sites: `(id1 << 16) + id2` operates on signed `int`, so `id1 >= 32768` shifts to a value outside the representable range. The new test also constructs `(32768 << 16) + 5` this way. Could the packed representation use `uint32_t` (or cast before shifting) end-to-end, then decode from that unsigned value?

ilmari99:
Great, thank you, you're right that the mask only addressed the read and left the packing itself undefined. Your suggestion to switch to uint32_t was good and simplified the fix. I have implemented that; end-to-end uint32_t. I updated the PR description to match.

Let me know if there is something else.

ilmari99:
Rebased onto current main (zizmor test was not passing) and added a regression test.

sylvesterkaczmarek (commented):
One high-index overflow remains in the pair-count arithmetic: `nactive*(nactive-1)/2` and `nbodyflex*(nbodyflex-1)/2` multiply in signed `int`. Above about 46,341 entries the intermediate product overflows even though the final halved value still fits `int` up to 65,535, which `mj_SAP` accepts. Could the multiplication be widened before taking the product?

ilmari99:
Thanks. Widened all three sites (`mj_collision, mj_broadphase, mj_collideFlexSAP`). The narrowing back to int is lossless here: the compiler caps bodies+flexes below 65534 and `mj_SAP` rejects `n >= 0x10000`. The test now uses 46342 elements to cover this case as well.

Unrelated, I noticed when the 1000000-pair cap in `mj_collideFlexSAP` is reached, `mj_SAP` returns the truncated list without a warning, so a very large folded 2D flex would silently lose contacts past that point.

kbayes:
@ilmari99 thank you for the PR. I think it's best to use a struct than to continue with the bit fiddling. Marking this closed due to https://github.com/google-deepmind/mujoco/commit/2944f5e70cf2f970ca6019338786ed38ba43da53.

If you want to reopen this or another PR with the repo unit test, then I'll be happy to merge it in.

PR #3539, opened 2026-09-02
Mask the high half when decoding SAP pairs to avoid a negative index

### LINKED ISSUE #3535: Out-of-bounds read in sweep-and-prune collision broadphase with more than 32768 items

Abhayindia:
Fixes #3535.

`mj_SAP` packs a pair of item indices as `(id1<<16) + id2` into a signed `int`. Once `id1 >= 0x8000` that sets bit 31, so the packed value is negative. Two sites decode the high half with a bare arithmetic shift:

```c
int bf1 = bfid[sappair[i] >> 16];   // engine_collision_driver.c, body broadphase
int e1  = elid[sappair[i] >> 16];   // mj_collideFlexSAP
```

so the shift sign-extends and the array is indexed at a negative offset - an out-of-bounds read (SEGV in the flex path once the garbage id feeds unchecked pointer arithmetic). A third site decoding the same packing already masks it:

```c
int bf1 = (broadphasepair[i]>>16) & 0xFFFF;   // line ~639
```

This brings the other two in line with `& 0xFFFF`. Reachable with a 2D flex over ~33k elements and `selfcollide="auto"`, or a plain model with more than 32768 mutually-collidable bodies (`mj_SAP` only rejects `n >= 0x10000`).

No unit test: triggering it needs >32768 collidable bodies, and a model that size takes minutes to compile. The issue has a full AddressSanitizer trace of the pre-fix crash.

### DIFF
--- a/src/engine/engine_collision_driver.c
+++ b/src/engine/engine_collision_driver.c
@@ -618,7 +618,7 @@ void mj_collision(const mjModel* m, mjData* d) {
   // broadphase collision detector
   TM_START;
   int nmaxpairs = (nbodyflex*(nbodyflex - 1))/2;
-  int* broadphasepair = mjSTACKALLOC(d, nmaxpairs, int);
+  unsigned* broadphasepair = mjSTACKALLOC(d, nmaxpairs, unsigned);
   int nbfpair = mj_broadphase(m, d, broadphasepair, nmaxpairs);
   unsigned int last_signature = -1;
   TM_END(mjTIMER_COL_BROAD);
@@ -633,12 +633,11 @@ void mj_collision(const mjModel* m, mjData* d) {
   size_t parena = alignArena(d, _Alignof(int));
 
   for (int i=0; i < nbfpair; i++) {
-    // reconstruct bodyflex pair ids
-    int bf1 = (broadphasepair[i]>>16) & 0xFFFF;
-    int bf2 = broadphasepair[i] & 0xFFFF;
+    // packed as (bf1 << 16) | bf2, bf1 < bf2
+    unsigned int signature = broadphasepair[i];
+    int bf1 = signature >> 16;
+    int bf2 = signature & 0xFFFF;
 
-    // compute signature for this bodyflex pair
-    unsigned int signature = (bf1<<16) + bf2;
     // pairs come sorted by signature, but may not be unique
     // if signature is repeated, skip it
     if (signature == last_signature) {
@@ -1350,7 +1349,7 @@ static void makeAAMM(const mjModel* m, mjData* d,
 
 // add bodyflex pair in buffer; do not filter if m is NULL
 static void add_pair(const mjModel* m, int bf1, int bf2,
-                     int* npair, int* pair, int maxpair) {
+                     int* npair, unsigned* pair, int maxpair) {
   // add pair if there is room in buffer
   if ((*npair) < maxpair) {
     // contact filtering if m is not NULL
@@ -1392,11 +1391,11 @@ static void add_pair(const mjModel* m, int bf1, int bf2,
       }
     }
 
-    // add pair
+    // add pair; pack as unsigned so the shift is defined for ids >= 0x8000
     if (bf1 < bf2) {
-      pair[*npair] = (bf1<<16) + bf2;
+      pair[*npair] = ((unsigned)bf1 << 16) | (unsigned)bf2;
     } else {
-      pair[*npair] = (bf2<<16) + bf1;
+      pair[*npair] = ((unsigned)bf2 << 16) | (unsigned)bf1;
     }
     (*npair)++;
   } else {
@@ -1434,7 +1433,7 @@ mjSORT(SAPsort, mjtSAP, SAPcmp);
 // given list of axis-aligned bounding boxes in AAMM (min[3], max[3]) format,
 // return list of pairs (i, j) in format (i<<16 + j) that can collide,
 // using sweep-and-prune along specified x axis (0, 1 or 2).
-static int mj_SAP(mjData* d, const mjtNum* aamm, int n, int axis_x, int* pair, int maxpair) {
+static int mj_SAP(mjData* d, const mjtNum* aamm, int n, int axis_x, unsigned* pair, int maxpair) {
   // check inputs
   if (n >= 0x10000 || axis_x < 0 || axis_x > 2 || maxpair < 1) {
     return -1;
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
Masking the high half fixes the negative decode, but the packed value is still created with signed (id1 << 16), which is undefined once id1 >= 0x8000. Please make the packed representation unsigned end-to-end rather than only masking the decode.

Abhayindia:
good call. reworked it: the pair buffers, `mj_SAP`/`mj_broadphase`/`add_pair` and the `bfsort` comparator are all `unsigned` now, so the `(id << 16)` pack is defined and the decodes go back to a plain `>> 16` / `& 0xFFFF`. no signed shift left in the path. force-pushed.

sylvesterkaczmarek (approved):
Rechecked the current head. Pair storage and sort buffers are unsigned end to end, and packing casts before the shift, so IDs at or above 0x8000 no longer invoke signed-shift undefined behaviour. My previous blocker is resolved.

kbayes:
@Abhayindia thank you for the PR. I think it's best to use a struct than to continue with the bit fiddling.  Marking this closed due to https://github.com/google-deepmind/mujoco/commit/2944f5e70cf2f970ca6019338786ed38ba43da53.

PR #3542, opened 2026-09-02
Fix Studio crash when reloading models with active decorations

nico-bohlinger:
Hey!
I ran into a repeatable Studio crash when loading a new model after enabling visualization options that add decorations:

```text
ERROR: Attempting to remove renderable from wrong scene.
```

The decoration renderables were destroyed without first being unregistered from their `SceneView`. That left registrations behind, so the later scene teardown tried to remove renderables that no longer belonged to it.

I simply added a call of `ModelDecorations::Clear()` in the destructor before freeing the scene. I also added a test that renders a decoration, reinitializes the renderer, and renders again.

### DIFF
--- a/src/experimental/studio/hal/renderer_test.cc
+++ b/src/experimental/studio/hal/renderer_test.cc
@@ -68,6 +68,26 @@ TEST_F(RendererTest, OpengGlSoftware) {
   }
 }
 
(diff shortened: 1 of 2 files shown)


yuvaltassa:
Fixed in 3d4447e1044ac99a4604653f74e97e79fd4ffd83. Thanks for reporting!

PR #3544, opened 2026-09-03
enable multiccd by default in Unity

Liuweixian:
Hi,

I am trying to use mujoco in Unity but found an issue about multiccd flag.

Set multiccd flag enable by default in Unity.

In my case, the multiccd flag was disable in GlobalSettings  when '<option> <flag multiccd="enable" /> </option>'  describe in xml. Because multiccd is enable by default, so after compiled, this flag will dismiss. but the default value in Unity still keep disable that cause problem.


SebasNorenaR:
solved in https://github.com/google-deepmind/mujoco/commit/a90ca41c00f43efc745dd32b2fd7cac2fa0f6004

PR #3554, opened 2026-09-06
Extend gyroscopic stabilization to free rigid subtrees

### LINKED ISSUE #3553: Extend implicitfast gyroscopic stabilization to free bodies with fixed descendants
### Problem

`implicitfast`'s local gyroscopic stabilization excludes a free body when it has massive fixed descendants. Consequently, two representations of the same rigid object can have substantially different numerical stability: moving an inertial element into an otherwise empty, fixed child body is enough to turn a stable free spin into `mjWARN_BADQACC`.

I understand that the [documentation](https://mujoco.readthedocs.io/en/3.12.0/computation/index.html#gyroscopic-derivatives-for-free-bodies) currently limits this treatment to standalone free bodies. This is a request to extend that treatment to free rigid subtrees, rather than a claim that the implementation contradicts the documented eligibility rule.

### Minimal reproduction

No contacts, gravity, controls, external assets, or plugins are involved.

```python
import mujoco

print(mujoco.__version__)
inertia = '''<inertial pos="0 0 0" mass="0.12"
             diaginertia="0.00017231 0.00000658 0.00017243"/>'''

for nested in (False, True):
    for integrator in ("implicitfast", "implicit"):
        payload = f"<body>{inertia}</body>" if nested else inertia
        model = mujoco.MjModel.from_xml_string(f"""
          <mujoco>
            <option gravity="0 0 0" timestep="0.001" integrator="{integrator}"/>
            <worldbody><body><freejoint/>{payload}</body></worldbody>
          </mujoco>
        """)
        data = mujoco.MjData(model)
        data.qvel[3:] = [100, 30, 100]
        for step in range(10000):
            mujoco.mj_step(model, data)
            if data.warning[mujoco.mjtWarning.mjWARN_BADQACC].number:
                print(integrator, nested, "BADQACC at step", step)
                break
        else:
            print(integrator, nested, "completed 10000 steps")
```

Observed with 3.12.0, and independently reproduced using the equivalent C API model construction and stepping loop against upstream commit `707f4735e248872a38c5040a21b510a5bd386459` (3.12.1 development), built in Release mode with GCC 11.5.0 on Linux x86_64:

| Integrator | Fixed child | Result |
| --- | --- | --- |
| implicitfast | No | Completes 10,000 steps |
| implicit | No | Completes 10,000 steps |
| implicitfast | Yes | BADQACC at step 4294, time 4.294 s, DOF 3 |
| implicit | Yes | Completes 10,000 steps |

### Possible direction

[`mj_isFreeBody`](https://github.com/google-deepmind/mujoco/blob/707f4735e248872a38c5040a21b510a5bd386459/src/engine/engine_derivative.c#L822) already checks that the tree has exactly six DOFs, but additionally requires `body_subtreemass == body_mass`. A free root with only fixed descendants still has a six-DOF inertia block, so it appears eligible for the same local solve.

Removing the mass check alone would be incorrect: `mjd_freeBias_vel` and `mjd_freeMhat` currently compute the bias derivative using only the root body's inertia. The derivative needs to include all fixed descendants, expressed about the free joint's origin and orientation. The fluid derivative symmetry exception also needs to recognize geoms on fixed descendants.

Would extending the existing treatment to these rigid subtrees be acceptable? I am preparing a draft PR with a regression test, derivative checks, and comparisons against `implicit`, so that the proposed scope and implementation can be reviewed together. Articulated trees would retain their existing behavior.

ruziniuuuuu:
Fixes #3553.

A free rigid object can diverge under `implicitfast` simply because its inertia is represented on a fixed child body. In the issue's contact-free reproduction, a massless free root with one inertial child emits BADQACC at step 4294; moving the same inertial element to the root, or using `implicit`, completes 10,000 steps.

This extends the existing local gyroscopic solve to free rigid subtrees. The six-DOF tree check remains, and the bias derivative sums each body's contribution about the free joint's origin and orientation. `mjd_freeMhat` reuses that aggregate derivative. Fluid geoms on fixed descendants retain the asymmetric derivatives required by the local solve. The shared helper also extends the existing `discrete` treatment to eligible rigid subtrees.

The implementation continues using the global Cholesky solve and a local 6-by-6 LU solve; descendant aggregation adds work proportional to the number of bodies in each eligible rigid subtree. No model rewriting, inertia regularization, new settings, or integrator fallback is introduced. Trees with articulated descendants remain excluded.

### Validation

Built against upstream `707f4735e248872a38c5040a21b510a5bd386459` using GCC 11.5.0, Linux x86_64, CMake Release, double precision.

- Before the fix: the minimal C API reproduction fails at step 4294; new energy-stability, derivative, and `implicit` comparison checks fail.
- After the fix: all four flat/nested and implicitfast/implicit reproduction combinations complete 10,000 steps without BADQACC.
- Full `engine_forward_test`: 129 tests pass.
- Full `engine_derivative_test`: 34 tests pass.
- Coverage includes a massless free root, multiple levels of rotated and offset inertias, another independent free body, fixed-child ellipsoid fluid forces, full RNE and central finite-difference derivative comparisons, `discrete` energy stability, and rejection of articulated descendants.
- Changed C++ lines formatted with the repository's clang-format configuration; `git diff --check` passes.

Build and test commands (run test binaries from the source `test/` directory so fixtures resolve):

```sh
cmake -S . -B ../build -DMUJOCO_BUILD_EXAMPLES=OFF -DMUJOCO_BUILD_SIMULATE=OFF \
  -DMUJOCO_BUILD_TESTS=ON -DCMAKE_BUILD_TYPE=Release
cmake --build ../build --target engine_forward_test engine_derivative_test -j 6
cd test
../../build/bin/engine_forward_test
../../build/bin/engine_derivative_test
```

This is a draft for maintainer feedback on extending the documented standalone-body eligibility rule. Validation is limited to the two complete test binaries above and the synthetic reproduction; other platforms, single precision, the full repository suite, and application workloads have not been tested.

### DIFF
--- a/doc/computation/index.rst
+++ b/doc/computation/index.rst
@@ -614,7 +614,8 @@ Gyroscopic derivatives for free bodies
    with the faster Cholesky decomposition. However integrating gyroscopic forces explicitly can lead to
    energy gain and divergence of fast-spinning free bodies with asymmetric inertia.
 
-   Therefore for *standalone free bodies* (free joints whose body has no children), these derivatives are reinstated.
+   Therefore for *standalone free bodies* (free joints whose body has no descendants with joints), these derivatives
+   are reinstated. Fixed descendants are included in the rigid body's inertia and bias derivatives.
    The rows of :math:`\widehat M` corresponding to such a body form a :math:`6\times 6` block which is decoupled from
    the rest of the system. After the global Cholesky solve, this block is re-assembled with the exact derivative of the
    body's bias force and re-solved with an optimized :math:`6\times 6` LU routine. For standalone free bodies,
--- a/src/engine/engine_derivative.c
+++ b/src/engine/engine_derivative.c
@@ -789,36 +789,34 @@ static void freeBias_vel_blocks(mjtNum mass, const mjtNum R[9], const mjtNum Xi[
 }
 
 
-// 6x6 block B = d qfrc_bias / d qvel for a standalone free body
+// 6x6 block B = d qfrc_bias / d qvel for a free rigid subtree
 //   assembles the full 6x6 from the 3x3 sub-blocks computed by freeBias_vel_blocks
 //   rows/cols ordered like the free joint dofs: [linear(3); rotational(3)]
 //   linear columns are zero: the bias force does not depend on linear velocity
 void mjd_freeBias_vel(const mjModel* m, const mjData* d, int jnt, mjtNum B[36]) {
   int body = m->jnt_bodyid[jnt];
   int adr = m->jnt_dofadr[jnt];
-  mjtNum mass = m->body_mass[body];
-  const mjtNum* R = d->xmat + 9*body;    // body  -> world
-  const mjtNum* Xi = d->ximat + 9*body;  // inertia -> world
-  const mjtNum* inertia = m->body_inertia + 3*body;
-
-  // CoM offset from joint origin, world frame
-  mjtNum s[3];
-  mji_sub3(s, d->xipos + 3*body, d->xpos + 3*body);
-
-  mjtNum lin[9], rot[9];
-  freeBias_vel_blocks(mass, R, Xi, inertia, s, d->qvel + adr + 3, lin, rot);
-
   mju_zero(B, 36);
-  for (int r=0; r < 3; r++) {
-    for (int c=0; c < 3; c++) {
-      B[6*r + 3+c] = -mass * lin[3*r+c];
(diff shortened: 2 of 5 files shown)


yuvaltassa:
ya, makes sense. do the CLA thing and we'll discuss?

ruziniuuuuu:
Hi Yuval, I've done the CLA thing.

yuvaltassa:
This was merged, but note I changed it quite a bit. Do you understand what I did? The composite inertia is already available. Anyway good idea, thanks for the PR!

PR #3555, opened 2026-09-06
Avoid catastrophic cancellation in mju_springDamper's overdamped roots

### LINKED ISSUE #3551: mju_springDamper loses slow decay under strong overdamping
### Intro

I'm an undergraduate engineering student at ASU using MuJoCo for robotics research. I found a separate numerical issue in `mju_springDamper` while checking its behavior after #3522.

### My setup

- MuJoCo `main` at `bbb71a28a4d16db0d42f98da84e9e56ff663df5f` (reports 3.12.1)
- C API, macOS 26.6.2, arm64, AppleClang 21
- Reproduced in double precision and in a separate build with `mjUSESINGLE`

For a strongly overdamped system, `mju_springDamper` can lose the slow decay mode entirely:

```c
mju_springDamper(1, 0, 1, 1e9, 1e9)
```

This returns `1.0`. The analytical solution is approximately `0.3678794411714423`, checked using 90-digit Decimal arithmetic. The discriminant is positive and the overdamped branch is selected correctly.

For `x'' + b*x' + k*x = 0`, this example has `k = 1`, `b = 1e9`, and is evaluated at `t = b/k`, approximately one slow-mode time constant. These are extreme separated timescales, but the expected position is finite and well resolved. A consistent change of time units gives the equivalent call `mju_springDamper(1, 0, 1e18, 1e18, 1)`, which also returns `1.0`.

In single precision, the same loss already occurs with `b = 1e4`, `k = 1`, and `t = 1e4`.

### Steps and code for reproduction

Build the library at the commit above, then compile and link this program against it. No model or assets are needed.

```c
#include <stdio.h>
#include <mujoco/mujoco.h>

int main(void) {
  printf("%.17g\n", (double)mju_springDamper(1, 0, 1, 1e9, 1e9));
  return 0;
}
```

Observed output: `1`. Expected: approximately `0.3678794411714423`.

### Cause

The [overdamped root calculation](https://github.com/google-deepmind/mujoco/blob/bbb71a28a4d16db0d42f98da84e9e56ff663df5f/src/engine/engine_util_misc.c#L1617-L1626) computes

```c
r1 = -b/2 + sqrt(b*b - 4*k)/2;
r2 = -b/2 - sqrt(b*b - 4*k)/2;
```

When `b*b` is much larger than `4*k`, the subtraction in `r1` loses precision and can round the slow root to zero. Computing the root that avoids cancellation first, then recovering the other from `r1*r2 = k`, preserves the slow mode. Which root is computed directly should depend on the sign of `b`.

The existing 66 utility tests pass on the unmodified library, including the #3522 tests. The regime-selection fix from that PR remains valid; this failure is in the subsequent root evaluation.

I checked the current function documentation and searched the repository's issues, PRs, and discussions for related reports. I found no existing report of this root-cancellation case.

Abhayindia:
Fixes #3551.

The overdamped branch of `mju_springDamper` computes the two roots as

```c
r1 = -b/2 + sqrt(b*b - 4*k)/2;
r2 = -b/2 - sqrt(b*b - 4*k)/2;
```

When `b*b` is much larger than `4*k`, `sqrt(b*b - 4*k)/2` is very close to `b/2`, so `r1` is the difference of two nearly-equal numbers and the slow root (`~ -k/b`) is lost - it can round to exactly zero, and then that decay mode disappears from the result.

`mju_springDamper(1, 0, 1, 1e9, 1e9)` returns `1.0`; the analytical answer is `1/e = 0.36787944117144233` (the reporter checked it with 90-digit arithmetic). In single precision it already breaks at `b = 1e4`.

Fix: compute whichever root does not involve a subtraction of near-equal terms (depends on the sign of `b`), then get the other from `r1*r2 = k`. This is the standard stable form for a quadratic. The regime selection from #3522 is untouched - this is only the root evaluation inside the overdamped branch.

Verified against the installed lib:

```
$ mju_springDamper(1, 0, 1,    1e9, 1e9)  ->  0.36787944117144233   (was 1.0)
$ mju_springDamper(1, 0, 1e18, 1e18, 1)   ->  0.36787944117144233   (same system, rescaled time)
$ mju_springDamper(1, 0, 1,    3,   1)    ->  0.78664559930336819   (unchanged)
```

`engine_util_misc_test` passes, including a new `SpringDamperStronglyOverdamped` case.

### DIFF
--- a/src/engine/engine_util_misc.c
+++ b/src/engine/engine_util_misc.c
@@ -1617,9 +1617,14 @@ mjtNum mju_springDamper(mjtNum pos0, mjtNum vel0, mjtNum k, mjtNum b, mjtNum t) 
     // compute w = sqrt(det)/2
     w = mju_sqrt(det)/2;
 
-    // compute r1,r2
-    r1 = -b/2 + w;
-    r2 = -b/2 - w;
+    // take the root without subtractive cancellation, recover the other from r1*r2 = k
+    if (b >= 0) {
+      r2 = -b/2 - w;
+      r1 = k/r2;
+    } else {
+      r1 = -b/2 + w;
+      r2 = k/r1;
+    }
 
     // compute coefficients
     c1 = (pos0*r2-vel0) / (r2-r1);
--- a/test/engine/engine_util_misc_test.cc
+++ b/test/engine/engine_util_misc_test.cc
@@ -118,6 +118,16 @@ TEST_F(UtilMiscTest, SpringDamperInvariantToTimeUnits) {
   }
 }
 
+TEST_F(UtilMiscTest, SpringDamperStronglyOverdamped) {
+  // b*b >> 4*k: the slow root ~ -k/b must not be lost to cancellation in
+  // -b/2 + sqrt(b*b - 4*k)/2. evaluated at t = b/k, one slow-mode time
+  // constant, so the answer is 1/e.
+  constexpr mjtNum inv_e = 0.36787944117144233;
+  EXPECT_NEAR(mju_springDamper(1, 0, 1, 1e9, 1e9), inv_e, MjTol(1e-10, 1e-4));
+  // the same physical system with time rescaled by 1e9
+  EXPECT_NEAR(mju_springDamper(1, 0, 1e18, 1e18, 1), inv_e, MjTol(1e-10, 1e-4));
+}
+
 TEST_F(UtilMiscTest, SphereWrap) {
   static constexpr char xml[] = R"(
   <mujoco>


Vedangalle:
Hi @Abhayindia, I reported #3551 and opened #3552 yesterday with the same root calculation. It also includes regression coverage for velocity-driven decay, zero stiffness, and nonpositive damping, and its CI is passing.

Flagging the overlap so we can avoid duplicate review. The additional time-rescaled regression here looks useful. Could we consolidate the work in #3552?

Abhayindia:
sorry, missed that #3552 was already up - i checked the issue comments but not the linked PRs. yours is more complete, go with it. the time-rescaled case (`mju_springDamper(1, 0, 1e18, 1e18, 1)`) is worth adding since it catches the same bug through a different path.

PR #3556, opened 2026-09-07
Fix "unpinned action reference" alerts from zizmor scan.

mmossg:


yuvaltassa (approved):

yuvaltassa (approved):

mmossg:
landed in 420b461d0946

PR #3557, opened 2026-09-07
Test new Google Chat failure alert

mmossg:
This is only a test.

If you happen to see an alert, please ignore.


mmossg:
abandoned

Issue #3559, opened 2026-09-07
URDF inertial origin rotation is lost during compilation, changing rotational dynamics


sri299792458:
### Intro

Hi! I use MuJoCo for robot simulation and found this while checking the inertial properties of imported URDF models.

### My setup

- MuJoCo **3.12.0**, official Python package; also reproduced on **3.3.7**.
- Python 3.11.15, NumPy 2.2.6.
- Ubuntu 24.04.4 LTS, x86_64.
- Native Python APIs: `MjModel.from_xml_string` and `MjSpec.from_string(...).compile()`.

### What's happening? What did you expect?

A URDF body's `<inertial><origin rpy="..."/></inertial>` rotation is lost during compilation. This changes the reconstructed inertia tensor in the link frame and the resulting rotational dynamics.

The minimal example has one floating body, with its center of mass at the link origin. It specifies inertia `diag(1, 2, 2.5)` in an inertial frame rotated 90 degrees about Z. Therefore its inertia in the link frame should be `diag(2, 1, 2.5)`.

Both URDF import paths instead produce `diag(1, 2, 2.5)`. For an angular acceleration of 1 rad/s² about link X, inverse dynamics consequently returns **1 Nm**, rather than the expected **2 Nm**. An equivalent MJCF with the inertia orientation expressed as a quaternion produces the expected tensor and 2 Nm torque.

| Input / loader | Reconstructed link-frame inertia | X torque for 1 rad/s² |
|---|---|---|
| URDF / `MjModel` | `diag(1, 2, 2.5)` | 1 Nm |
| URDF / `MjSpec` | `diag(1, 2, 2.5)` | 1 Nm |
| Equivalent MJCF | `diag(2, 1, 2.5)` | 2 Nm |

The comparison reconstructs the full tensor from `body_iquat` and `body_inertia`, so it accounts for principal-axis/eigenvalue reordering. `fusestatic` and `alignfree` are explicitly disabled, and the model has no contacts or external assets.

### Steps for reproduction

Install `mujoco==3.12.0` and `numpy==2.2.6`, then run the self-contained script below. It includes the complete URDF, an equivalent MJCF control case, and both tensor and inverse-dynamics checks. The same script produces the same discrepancy on 3.3.7.

### Minimal models and code for reproduction

```python
import platform
import sys

import mujoco
import numpy as np

urdf = """<robot name="inertia_frame">
  <mujoco><compiler fusestatic="false" alignfree="false"/></mujoco>
[...]

PR #3560, opened 2026-09-08
Fix lost URDF inertial origin rotation during compilation

devansh0703:
## What does this PR do?

URDF `<inertial><origin rpy=...>` rotations were being silently dropped during compilation, changing the rotational dynamics (issue #3559).

**Root cause:** the URDF reader stored the inertia tensor in the inertial frame and pre-composed the principal-axis rotation into `iquat` (`src/xml/xml_urdf.cc`). `mjCBody::Compile` then called `mjuu_fullInertia(iquat, inertia, fullinertia)` *again* on the unchanged tensor, which overwrote `iquat` with the principal-axis orientation and discarded the inertial-origin rotation.

**Fix (both changes required together):**

1. `src/user/user_objects.cc` — rotate the `fullinertia` tensor from the inertial frame into the body frame (`mjuu_quat2mat` + `mjuu_mulRMRT`, i.e. `R·M·Rᵀ`) before decomposing with `mjuu_fullInertia`. The compiler is now the single decomposition point.
2. `src/xml/xml_urdf.cc` — removed the `lquat` pre-composition block so `iquat` is the pure inertial-origin rotation; the compiler composes it correctly for both aligned and non-aligned free bodies.

## Why is this needed?

Without it, a URDF like

```xml
<inertial>
  <origin xyz="0 0 0" rpy="0 0 1.570796"/>
  <mass value="1"/>
  <inertia ixx="1" iyy="2" izz="2.5"/>
</inertial>
```

produces the wrong inertia (`diag(1,2,2.5)` instead of `diag(2,1,2.5)`), and the equivalent MJCF gives a different result — so a URDF's dynamics depend on file format rather than physics.

## Tests

Added `MujocoTest.UrdfInertialOriginRotation` to `test/xml/xml_urdf_test.cc` (and wired the previously-orphaned `xml_urdf_test` into `test/xml/CMakeLists.txt`). It verifies the world-frame inertia tensor is `diag(2,1,2.5)` and the inverse-dynamics X torque under **both** `alignfree=true` (URDF default) and `alignfree=false`. Confirmed the new test **fails on unmodified source** and **passes with the fix**.

Also verified existing URDF/MJCF reader, writer, and user-object test suites pass.

### DIFF
--- a/src/user/user_objects.cc
+++ b/src/user/user_objects.cc
@@ -2577,7 +2577,25 @@ void mjCBody::Compile(void) {
 
   // process orientation alternatives for inertia
   if (mjuu_defined(fullinertia[0])) {
-    const char* err = mjuu_fullInertia(iquat, inertia, this->fullinertia);
+    // rotate tensor from inertial frame to body frame
+    double mat[9], full[9], full_body[9], fullinertia_body[6];
+    mjuu_quat2mat(mat, iquat);
+    full[0] = this->fullinertia[0];
+    full[4] = this->fullinertia[1];
+    full[8] = this->fullinertia[2];
+    full[1] = full[3] = this->fullinertia[3];
+    full[2] = full[6] = this->fullinertia[4];
+    full[5] = full[7] = this->fullinertia[5];
+    mjuu_mulRMRT(full_body, mat, full);
+    fullinertia_body[0] = full_body[0];
+    fullinertia_body[1] = full_body[4];
+    fullinertia_body[2] = full_body[8];
+    fullinertia_body[3] = full_body[1];
+    fullinertia_body[4] = full_body[2];
+    fullinertia_body[5] = full_body[5];
+
+    // decompose rotated tensor into principal axes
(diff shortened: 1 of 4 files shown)


Miko997:
The diagonal case does not exercise an oblique rotation of off-diagonal inertia; [this test-only commit](https://github.com/Miko997/mujoco/commit/e7031da8abaef25da42c1b0473d2cfe706d926af) adds that coverage, including three compilations of the same `mjSpec` and full-precision URDF→MJCF round-trips under both `alignfree` settings. Local C++ builds reproduce the failure on base `a7e365917a447dc44e280ff9567064228657d266` (`Ixx ≈ 4` instead of `1004/375`) and pass on head `5ffed86e6884e1b9112cf15367da62efe1ea2453` (maximum world-tensor entry error below `7e-9` against analytically computed `R I Rᵀ`). The test compares physical tensors through `ximat`, accounting for free-body frame alignment; all repeated compilations and round-trips pass at `1e-7`. The commit only adds the regression to `xml_urdf_test.cc` and can be included here; it contains no implementation changes.

PR #3562, opened 2026-09-08
Add deterministic tie-breaking and fix contact compaction in filterFlexContacts

thowell:
When filtering flex contacts in filterFlexContacts, candidate selection previously used strict inequalities for deepest-seed search and farthest-point distance updates. On symmetric meshes (such as a flat cloth sheet on a plane), candidate selection order was therefore non-deterministic and sensitive to broadphase sweep emission order. In addition, the existing in-place contact swap desynchronized candidates from the un-swapped selected and min_dist tracking buffers, which could lead to candidates being skipped or re-evaluated.

This change introduces tieBreakFlexContact, enforcing a strict lexicographic total order on (elem[0], elem[1], candidate index) when distances are exactly equal in both the deepest-seed and farthest-point selection passes. Candidate selection now tracks chosen indices in a temporary selection buffer and compacts the array once upon loop termination, along with an early exit when all remaining candidates are exhausted. A unit test is added in engine_collision_driver_test.cc verifying capping to mjMAXCONPAIR and distance validity under symmetric penetration. A brief comment is also added above mjc_ConvexElem documenting the benchmark evaluation of 2D SAT prefiltering for flex elements.

### DIFF
--- a/src/engine/engine_collision_driver.c
+++ b/src/engine/engine_collision_driver.c
@@ -441,6 +441,22 @@ static inline int contactcompare(const mjContact* c1, const mjContact* c2, void*
 mjSORT(contactSort, mjContact, contactcompare);
 
 
+// Strict total-order tie-breaker for flex contact selection: (elem1, elem2, candidate index)
+static inline int tieBreakFlexContact(const mjContact* c_curr, int idx_curr,
+                                      const mjContact* c_sel, int idx_sel) {
+  if (idx_sel < 0) {
+    return 1;
+  }
+  if (c_curr->elem[0] != c_sel->elem[0]) {
+    return c_curr->elem[0] < c_sel->elem[0];
+  }
+  if (c_curr->elem[1] != c_sel->elem[1]) {
+    return c_curr->elem[1] < c_sel->elem[1];
+  }
+  return idx_curr < idx_sel;
+}
+
+
 // filter flex contacts based on distance
 static void filterFlexContacts(mjData* d, int ncon_before) {
   int n = d->ncon - ncon_before;
@@ -453,6 +469,7 @@ static void filterFlexContacts(mjData* d, int ncon_before) {
   mj_markStack(d);
   mjtByte* selected = mjSTACKALLOC(d, n, mjtByte);
   mjtNum* min_dist = mjSTACKALLOC(d, n, mjtNum);
+  mjContact* temp_contacts = mjSTACKALLOC(d, mjMAXCONPAIR, mjContact);
   memset(selected, 0, n);
 
   for (int i = 0; i < n; i++) {
@@ -464,14 +481,20 @@ static void filterFlexContacts(mjData* d, int ncon_before) {
   int best = 0;
   mjtNum bestdist = -contacts[0].dist;
   for (int i = 1; i < n; i++) {
-    if (-contacts[i].dist > bestdist) {
-      bestdist = -contacts[i].dist;
+    mjtNum dist_i = -contacts[i].dist;
+    if (dist_i > bestdist) {
+      bestdist = dist_i;
+      best = i;
+    } else if (dist_i == bestdist && tieBreakFlexContact(&contacts[i], i, &contacts[best], best)) {
       best = i;
     }
   }
 
+  int sel_idx[mjMAXCONPAIR];
+
   while (nselected < mjMAXCONPAIR && best >= 0) {
     selected[best] = 1;
+    sel_idx[nselected++] = best;
     mjtNum* bestpos = contacts[best].pos;
 
     int nextbest = -1;
@@ -489,22 +512,24 @@ static void filterFlexContacts(mjData* d, int ncon_before) {
       if (min_dist[i] > nextbestdist) {
         nextbestdist = min_dist[i];
         nextbest = i;
+      } else if (min_dist[i] == nextbestdist && nextbest >= 0 &&
+                 tieBreakFlexContact(&contacts[i], i, &contacts[nextbest], nextbest)) {
+        nextbest = i;
       }
     }
 
-    if (nselected < mjMAXCONPAIR - 1) {
-      mjContact temp = contacts[nselected];
-      contacts[nselected] = contacts[best];
-      contacts[best] = temp;
-
-      if (nextbest == nselected) {
-        nextbest = best;
-      }
+    if (nextbest < 0 || nextbestdist <= 0) {
+      break;
     }
 
-    nselected++;
     best = nextbest;
   }
+
+  // compact selected contacts
+  for (int i = 0; i < nselected; i++) {
+    temp_contacts[i] = contacts[sel_idx[i]];
+  }
+  memcpy(contacts, temp_contacts, nselected * sizeof(mjContact));
 
   mj_freeStack(d);
 
@@ -2565,6 +2590,9 @@ void mj_collideElems(const mjModel* m, mjData* d, int f1, int e1, int f2, int e2
 
   // general convex collision
   else {
+    // Note: A 2D SAT prefilter (as in MuJoCo Warp) was evaluated here prior to mjc_ConvexElem.
+    // While conservative (0 false rejects / 15.8M pairs), step time was unchanged (8718 vs 8781 us
+    // on cloth) because GJK rejection costs ~0.16 us/pair, matching SAT evaluation cost.
     num = mjc_ConvexElem(m, d, precon, -1, f1, e1, -1, f2, e2, margin + gap);
   }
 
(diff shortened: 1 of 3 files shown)


src/engine/engine_collision_driver.c
```diff
@@ -441,6 +441,22 @@ static inline int contactcompare(const mjContact* c1, const mjContact* c2, void*
...
+  if (idx_sel < 0) {
+    return 1;
+  }
+  if (c_curr->elem[0] != c_sel->elem[0]) {
+    return c_curr->elem[0] < c_sel->elem[0];
+  }
+  if (c_curr->elem[1] != c_sel->elem[1]) {
+    return c_curr->elem[1] < c_sel->elem[1];
+  }
+  return idx_curr < idx_sel;
```
smallquail, line 458: The index key is only comparable with the mjWarp side because contacts from a single element pair are emitted consecutively in both (`mjraw_CapsuleCapsule` here, the two-contact loop in the warp narrowphase); the first two keys separate everything else. Worth recording, since the warp-side index comes from an atomic counter and carries no ordering of its own.

src/engine/engine_collision_driver.c
```diff
@@ -489,23 +512,25 @@ static void filterFlexContacts(mjData* d, int ncon_before) {
...
 
-    if (nselected < mjMAXCONPAIR - 1) {
-      mjContact temp = contacts[nselected];
-      contacts[nselected] = contacts[best];
-      contacts[best] = temp;
-
-      if (nextbest == nselected) {
-        nextbest = best;
-      }
+    if (nextbest < 0 || nextbestdist <= 0) {
```
smallquail, line 524: `nextbestdist <= 0` is a second stopping rule beyond "no candidates left". `min_dist` holds squared distances, so selection now stops as soon as the farthest remaining candidate is coincident with an already-selected one, where the previous code kept selecting up to `mjMAXCONPAIR`. This matches the warp side's `max_min_d <= 0.0` and looks intended, but it changes `ncon` on scenes with coincident candidates, so it deserves a line in the description alongside the exhaustion exit.

test/engine/engine_collision_driver_test.cc
```diff
@@ -474,5 +474,36 @@ TEST_F(MjCollisionTest, MaxContact) {
...
+  )";
+  char error[1024];
+  MjModelPtr m = LoadModelFromString(xml, error, sizeof(error));
+  ASSERT_THAT(m.get(), NotNull()) << error;
+  MjDataPtr d = MakeData(m);
+  ASSERT_THAT(d, NotNull());
+
+  mj_fwdPosition(m.get(), d.get());
+
+  // A 9x9 flat grid produces 128 elements, all penetrating the plane
```
smallquail: Contacts here come from `mj_collidePlaneFlex`, which iterates flex vertices rather than elements — 81 for a 9×9 grid, not 128. The count still exceeds `mjMAXCONPAIR`, so the assertion holds, but the stated mechanism is wrong.

test/engine/engine_collision_driver_test.cc
```diff
@@ -474,5 +474,36 @@ TEST_F(MjCollisionTest, MaxContact) {
...
+  // identically. Contact filtering should cap the contacts to mjMAXCONPAIR
+  // (50).
+  EXPECT_EQ(d->ncon, mjMAXCONPAIR);
```
smallquail, line 503: This asserts the cap and that the distances are valid, but not *which* contacts are selected, so it passes identically with and without `tieBreakFlexContact`. Either assert the selected `elem` ids against a fixed expected set, or build the candidate array in a permuted order and assert the same set comes back — otherwise the tie-break the test is named for is uncovered.

quagla:
Left some nits, otherwise LGTM, thanks!

src/engine/engine_collision_driver.h
```diff
@@ -59,6 +59,9 @@ void mj_collideElems(const mjModel* m, mjData* d, int f1, int e1, int f2, int e2
 // test element and vertex for collision, add to contact list
 void mj_collideElemVert(const mjModel* m, mjData* d, int f, int e, int v);
 
+// filter flex contacts based on distance
+MJAPI void filterFlexContacts(mjData* d, int ncon_before);
```
smallquail, line 63: Exporting this so the test can call it adds a public symbol named like a static helper. Everything else exported from this header is `mj_`-prefixed (`mj_collideElems`, `mj_collideElemVert`, `mj_isElemActive`), so `mj_filterFlexContacts` would match — if the export is wanted at all. The alternative is keeping it static and driving the selection through `mj_fwdPosition`, as `FlexContactFilterTieBreak` already does.

test/engine/engine_collision_driver_test.cc
```diff
@@ -474,5 +477,124 @@ TEST_F(MjCollisionTest, MaxContact) {
...
+
+  // Generate 60 candidate contacts (> mjMAXCONPAIR) with identical penetration
+  // depth, where symmetric distances lead to ties resolved by (elem[0],
+  // elem[1], candidate index).
+  constexpr int kNumCandidates = 60;
+  static_assert(kNumCandidates > mjMAXCONPAIR,
+                "Candidate count must exceed max pair cap");
+
+  auto build_candidates = [](mjData* data, const std::vector<int>& elem_order) {
+    data->ncon = elem_order.size();
```
smallquail, line 542: This writes candidates straight into the arena without allocating them. `mj_resetData` poisons the whole arena (`engine_io.c:1319`) and only `mj_arenaAlloc*` unpoisons it, and `mjUSEASAN` is defined automatically whenever ADDRESS_SANITIZER is active (`mjsan.h:29-33`), so these are poisoned-memory writes under an ASAN build. The workflows here don't appear to enable one, so CI stays green either way. Allocating the contacts through the arena first would keep it safe.

test/engine/engine_collision_driver_test.cc
```diff
@@ -474,5 +477,124 @@ TEST_F(MjCollisionTest, MaxContact) {
...
+  ASSERT_THAT(d, NotNull());
+
+  mj_fwdPosition(m.get(), d.get());
+
+  // A 9x9 flat grid produces 81 vertices, all penetrating the plane identically
+  // via mj_collidePlaneFlex. Contact filtering should cap contacts to
+  // mjMAXCONPAIR (50).
+  EXPECT_EQ(d->ncon, mjMAXCONPAIR);
+
+  const std::vector<int> expected_verts = {
```
smallquail, line 505: Two notes on this assertion.

`mj_collidePlaneFlex` sets `elem[0] = elem[1] = -1` on every contact it emits, so here `tieBreakFlexContact` always falls through both element comparisons to `idx_curr < idx_sel` — the element keys are never exercised by this test. `FlexContactFilterTieBreakPermutation` covers `elem[0]`; nothing covers `elem[1]`.

Worth one line recording that this pins the selected *set* rather than the selection order: `contactSort` runs after `filterFlexContacts` on this path (the plane is collided inside `mj_collideTree`), which is why the list comes out ascending. Otherwise a future diff to these 50 values is hard to read as regression versus reordering.

src/engine/engine_collision_driver.c
smallquail: The code comment covers the mechanism. The PR description still calls this "an early exit when all remaining candidates are exhausted", which is only the `nextbest < 0` half — the coincident-distance stop is the part that changes `ncon`. The same paragraph also still describes the test as verifying "capping to mjMAXCONPAIR and distance validity", which the new assertions have outgrown.

quagla:
More nits.

thowell:
@quagla

Thanks for the review! I have addressed all feedback in the latest commit:

* **Candidate index tie-breaking rationale**: Clarified above `tieBreakFlexContact` that `(elem[0], elem[1])` partitions contacts across different element pairs, and the candidate index fallback breaks ties between contacts emitted for the same element pair (e.g. `mjraw_CapsuleCapsule`).
* **Coincident contact early exit**: Added an inline comment clarifying that `nextbestdist <= 0` terminates selection when remaining candidates are coincident (`min_dist <= 0`) with already selected contacts.
* **Plane-flex test mechanism**: Corrected the test comment to reflect that `mj_collidePlaneFlex` tests the 81 flex vertices against the plane rather than 128 elements.
* **Tie-break test verification**: Updated `FlexContactFilterTieBreak` to assert the selected vertex indices against a fixed expected set. Added a companion test `FlexContactFilterTieBreakPermutation` that feeds candidate contacts with identical depths and symmetric distances in forward, reversed, and randomly shuffled orderings, asserting that candidate selection is completely invariant to input order.

yuvaltassa:
@kbayes is touching this, he should TAL

kbayes:
Superseded by 0e563a9eadfca189e491da530db8e8f0ec7bcc31

PR #3563, opened 2026-09-08
fix(actuators): prevent cross-type parameter leakage in default classes

### LINKED ISSUE #3561: Loading MJCF in python that contains an actuator subclass default overwrites other subclass params
### Intro

I'm working on an RL environment, and wanted to compare various actuator settings' effects on the learning. I noticed that having defaults of one subclass of actuators in an MJCF  sets the params of other subclasses defined.

I know that the actuator subclasses are just shorthands for parameter settings and that almost everything is implemented as general actuators under the hood (with a bit of special treatment for muscles),  but this is still a surprising outcome for me.

### My setup

Using Linux and the Python API, MuJoCo version 3.12.

I would expect that defining defaults for one actuator subclass would leave other actuators unaffected. i.e., if I have a `position` default defined, and a `position` , a `motor` and a `muscle` actuator in the MJCF, only the `position` actuator should be affected. 

### Steps for reproduction

1. Load the model below in Python.
2. Print biasprm.

<details>

```XML

<mujoco model="mvp">
    <default>
        <joint type="hinge" pos="0 0 0" axis="1 0 0" stiffness="5" damping="0.01"/>
        <position ctrllimited="true" ctrlrange="0 1" kp="2" kv="3"/>
    </default>

    <worldbody>
        <body pos="0 0 0">
            <geom name="arm" type="capsule" size="0.005" fromto="0 0 0  0 0.1 0" rgba=".5 .1 .1 1"/>
            <joint name="shoulder"/>
        </body>
    </worldbody>

    <actuator>
        <motor name="m1" joint="shoulder"/>
    </actuator>

</mujoco>


```
</details>

```python
import mujoco
xml_path="mvp.xml"
print(mujoco.MjModel.from_xml_path(xml_path).actuator_biasprm[0])
```

### Confirmations

anishesg:
When multiple actuator shortcut types (e.g., `<position>`, `<motor>`, `<velocity>`) are defined within the same default class, they all write to the same actuator struct, causing parameters from one type to leak into others.

The fix clears the parameter arrays (`gainprm`, `biasprm`, `dynprm`) in the XML parser before setting type-specific values. This happens in `OneActuator()` right after determining the actuator type, but before reading XML attributes for inheritance.

This approach:
- Prevents cross-type parameter contamination in default classes
- Preserves same-type inheritance (e.g., `<position kp="5"/>` under a `<position kv="3"/>` default correctly inherits kv=3)
- Leaves general and plugin actuators unaffected

The key insight is that parameters must be cleared before reading attributes (which use current values as defaults), not after in the `mjs_setTo*` functions.

Fixes #3561

### DIFF
--- a/src/user/user_api.cc
+++ b/src/user/user_api.cc
@@ -1239,6 +1239,10 @@ const char* mjs_setToMotor(mjsActuator* actuator) {
   actuator->dyntype  = mjDYN_NONE;
   actuator->gaintype = mjGAIN_FIXED;
   actuator->biastype = mjBIAS_NONE;
+
+  // clear bias parameters to prevent inheriting from defaults
+  mjuu_zerovec(actuator->biasprm, mjNBIAS);
+
   return "";
 }
 
@@ -1250,6 +1254,9 @@ const char* mjs_setToPosition(mjsActuator* actuator,
                               double       dampratio[1],
                               double       timeconst[1],
                               double       inheritrange) {
+  // clear bias parameters to prevent inheriting from defaults
+  mjuu_zerovec(actuator->biasprm, mjNBIAS);
+
   actuator->gainprm[0] = kp;
   actuator->biasprm[1] = -kp;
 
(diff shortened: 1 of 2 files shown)


jonathanembleyriches:
Thanks for looking at this, but I don't think this addresses the core requirements of #3561. Every actuator shortcut in a default class writes into the same actuator struct, so they still clobber each other while the default is parsed. Zeroing `biasprm` in the `mjs_setTo*` functions only clears part of the leak. In the issue's example the motor's `biasprm` was never used anyway, and gain and dynamics parameters still leak across types. It also breaks same-type inheritance: `<position kp="5"/>` under a `<position kv="3"/>` default now loses its kv.

anishesg:
good catch, moved the fix to the XML parser. now clears gainprm/biasprm/dynprm before reading attributes for inheritance, so cross-type leakage is prevented but same-type inheritance still works

PR #3571, opened 2026-09-11
Fix dead links in python, mjx, and XML reference docs

rishabhsai:
The dm_control tutorial link in doc/python.rst, both Warp links in doc/mjx.rst, and the Wikipedia link behind "reflected inertia" in doc/XMLreference.rst return 404. The dm_control path dm_control/mjcf/tutorial.ipynb never existed in that repo, Warp moved its docs under /latest/ and split the JAX section into its own page, and no Wikipedia article exists under either "Reflective inertia" or "Reflected inertia". The tutorial link now points at the root tutorial.ipynb, which has the PyMJCF attach example the text describes, the Warp links point at latest/user_guide/interoperability/jax.html and latest/user_guide/differentiability.html, and the Wikipedia link is now plain text. With `curl -sL -o /dev/null -w '%{http_code}' -A Mozilla/5.0 "$url"`, the four old URLs return 404 and the three new ones return 200. I do not care about git authorship, so incorporating this directly is fine if that is faster.


rishabhsai:
@googlebot I signed it!

PR #3605, opened 2026-09-17
Add J2 plasticity plugin

Ei9en:
## Summary

This PR adds a passive plugin implementing history-dependent **perfect J2 plasticity** for 3D tetrahedral flexes.

The implementation uses a von Mises yield criterion with associative flow and radial return. Rather than introducing a separate elastic material model, it reuses MuJoCo's existing tetrahedral flex stiffness and adds a plastic correction to the native elastic force.

The initial scope is deliberately small: perfect plasticity only, with a single additional material parameter (`yield`).

## Motivation

I initially started looking into MuJoCo for a small robotics project involving deformable objects and wanted to experiment with permanent deformation of 3D solids.

My engineering curriculum has a strong mechanics background, including solid and continuum mechanics, so J2 plasticity seemed like a relatively small and well-defined constitutive extension for adding history-dependent material behavior to 3D flexes.

Potential applications include permanent bending, forming, crimping, indentation, and other manipulation tasks where the deformation history of an object matters.

## Implementation

Each tetrahedron stores the six independent components of its plastic Green strain.

The current formulation uses Green-Lagrange strain with the additive decomposition

$$
E^e = E - E^p.
$$

The deviatoric trial stress is evaluated with the shear modulus reconstructed from MuJoCo's native tetrahedral stiffness. If the von Mises equivalent stress exceeds the configured yield stress, a standard perfect-plastic radial return projects the state back onto the J2 yield surface.

The plugin reuses the six squared-edge coordinates already used by MuJoCo's 3D flex elasticity. If `q` is the current edge deformation and `q_p` is the contribution corresponding to the plastic strain, the combined native + plugin response is

$$
-G_x^T K(q-q^p).
$$

This means that the existing tetrahedral stiffness matrix and generalized-force projection remain responsible for the elastic response.

No duplicate Young's modulus or Poisson ratio parameters are introduced. The only plugin-specific material parameter is the von Mises yield stress:

```xml
<plugin plugin="mujoco.plasticity.j2">
  <config key="yield" value="2.5e8"/>
</plugin>
```

### State lifecycle

`Compute()` reads the committed plastic state, performs the trial evaluation and radial return, and stores a candidate updated state. The candidate is used for the current plastic force correction without modifying `mjData::plugin_state`.

`Advance()` commits that candidate after the integration step.

This keeps force evaluation separate from mutation of the persistent material history.

## Validation

The PR includes tests for:

- deformation-gradient and Green-strain kinematics;
- rigid-motion invariance;
- trial deviatoric and von Mises stress;
- radial return and yield-surface projection;
- elastic unloading;
- persistent plugin state and commit behavior;
- plastic edge-force construction;
- integration with native `qfrc_passive`;
- loading and unloading through the actual plugin callbacks;
- a physical single-tetrahedron aluminum regression.

All **7 plasticity test executables** pass locally in both MuJoCo double- and single-precision builds (**14/14**). Precision-dependent tolerances were calibrated using `MjTol` and the `MJTOL_SCALE=0` workflow.

I also used a homogeneous aluminum cube discretized into 48 tetrahedra as a development validation case (`E = 70 GPa`, `nu = 0.33`, `yield = 250 MPa`).

The shear modulus reconstructed from the native stiffness was `26.31578947 GPa`, matching the analytical value. Under homogeneous deviatoric loading and unloading, all 48 tetrahedra yielded consistently and the predicted residual strain matched the analytical perfect-plastic solution.

At the analytical residual configuration, the corrected force norm was approximately `2.0e-7 N`, compared with approximately `2.88 MN` for native elasticity alone, corresponding to a residual-force ratio of approximately `6.95e-14`.

aluminum_cube_benchmark

*Homogeneous loading/unloading validation. The plugin follows the elastic response before yielding, applies the J2 return after yield, and retains the expected permanent deformation after unloading.*

## Related work

#549 and #560 previously explored elastoplastic behavior for the cable elasticity plugin.

This PR targets a different part of MuJoCo's deformable-body system: **volumetric 3D tetrahedral flexes**. It also keeps visualization separate from the constitutive implementation; the viewer setup used during development is not part of this PR.

## Current scope and limitations

This first implementation assumes:

- 3D tetrahedral flexes;
- native MuJoCo flex elasticity;
- von Mises / J2 yielding;
- perfect plasticity;
- no hardening;
- additive Green-strain decomposition;
- six plastic-strain state variables per tetrahedron;
- non-interpolated flexes.

The additive Green-strain formulation is invariant to rigid-body rotations, but it is not intended as a general finite-strain plasticity formulation for large irreversible deformation.

Possible future extensions, if useful for MuJoCo, include isotropic or kinematic hardening, constitutive substepping, finite-strain plasticity, additional yield laws, viscoplasticity, and eventually damage/fracture models.

More details on the formulation and validation are included in `plugin/plasticity/README.md`.

This is my first contribution of this scope to MuJoCo, so feedback on both the constitutive choices and how this functionality fits within the existing flex/plugin architecture would be very welcome. I am happy to adjust the implementation or scope based on maintainer feedback.


sylvesterkaczmarek (commented):
`yield="nan"` is accepted as a valid yield stress. `std::strtod` parses it successfully, and the `*yield_stress <= 0` guard does not reject NaN; once plastic flow is evaluated, NaNs propagate through `yield_function`, `plastic_multiplier`, plastic strain, and the passive-force correction. Please require the parsed yield value to be finite as well as positive and add `nan`/`inf` attribute regressions.

yuvaltassa:
This is interesting.

@quagla PTAL

@sylvesterkaczmarek why should we disallow nan? It's useful for debugging. IIRC there is a place we disallow nans that I actually want to remove. Ah no, maybe we disallow inf? Anyway they should all be allowed IMO

sylvesterkaczmarek:
Fair point. I was treating `yield` as a production-only material parameter, but if MuJoCo deliberately permits NaN/Inf for diagnostics then rejecting them here would be inconsistent. In this plugin, `+inf` effectively disables yielding while `nan` deliberately propagates through the constitutive state, which can be useful for debugging. I withdraw that concern; no finite-value check or NaN/Inf regression needed on my account.

PR #3606, opened 2026-09-18
Index ctrllimited/ctrlrange by control slot in actuator derivatives

### LINKED ISSUE #3598: Discrete actuator derivative indexes control limits by actuator ordinal
### Intro

Hi!

I am testing order invariance for the discrete integrator with variable-width actuator inputs. The effective-metric derivative reads the target control from its control-block address but checks its limits using the actuator ordinal.

### My setup

- MuJoCo source build from `main` at `10124d5d9dca411ec3c8988aa1e3b619103d71bb`
- C API, double precision
- Ubuntu 24.04.5 LTS, Linux 7.0.0-28-generic x86_64
- Normal GCC 11.5.0 build and GCC 13.3.0 AddressSanitizer build

`actuatorInput` correctly reads `d->ctrl[m->actuator_ctrladr[i]]`, but then uses `m->actuator_ctrllimited[i]` and `m->actuator_ctrlrange[2*i]`. Those metadata arrays are indexed by packed control slot. I expected inserting or moving a physically inactive variable-width actuator to leave the target actuator's force derivative and next state unchanged.

The first reproducer builds the same three-input zero-force PID and limited affine actuator in opposite orders. With an in-range target control, both steps agree. With raw control `100`, the forward path clamps both target forces to `0.5`, but only the target-first derivative uses the clamped input:

```text
control=1 target-first adr=0 qvel=0.0049995000499950008 force=0.5
control=1 target-second adr=3 qvel=0.0049995000499950008 force=0.5
control=1 velocity-delta=0
control=100 target-first adr=0 qvel=0.0049995000499950008 force=0.5
control=100 target-second adr=3 qvel=0.0049504950495049506 force=0.5
control=100 velocity-delta=-4.9005000490050243e-05
```

The second reproducer places a zero-input DC motor before the target. A whole-library AddressSanitizer build reports:

```text
ERROR: AddressSanitizer: use-after-poison
READ of size 1
#0 actuatorInput ... src/engine/engine_derivative.c:2415
#1 actuatorLenDeriv ... src/engine/engine_derivative.c:2495
#2 mjd_effActuation ... src/engine/engine_derivative.c:4088
```

Relevant source:

- https://github.com/google-deepmind/mujoco/blob/10124d5d9dca411ec3c8988aa1e3b619103d71bb/src/user/user_model.cc#L3852-L3861
- https://github.com/google-deepmind/mujoco/blob/10124d5d9dca411ec3c8988aa1e3b619103d71bb/src/engine/engine_derivative.c#L2404-L2427
- https://github.com/google-deepmind/mujoco/blob/10124d5d9dca411ec3c8988aa1e3b619103d71bb/src/engine/engine_derivative.c#L4072-L4104

### Steps for reproduction

1. Build the affected revision normally and run `poc/order_invariance.cc` with the commands in `poc/README.md`.
2. Observe equal forces but different velocities only when the target is clamped and follows the three-input actuator.
3. Build MuJoCo and `poc/zero_input.cc` with AddressSanitizer.
4. Observe the invalid metadata read in `actuatorInput`.

Reproduction package: 

[discrete-integrator-control-limit-indexing-poc.zip](https://github.com/user-attachments/files/32260545/discrete-integrator-control-limit-indexing-poc.zip)

Both complete asset-free MJCF models are embedded in the C++ programs. The order test includes a physically zero dummy actuator as the metamorphic control.

See `poc/order_invariance.cc` and `poc/zero_input.cc`. The attached `fix.patch` uses the resolved control address consistently and adds an order-invariance regression. The focused regression and all 142 `engine_forward_test` cases pass; single-precision and repository-wide CI remain pending.

### Confirmations

Mohit-Ak:
Fixes #3598

## What was wrong

`actuatorInput` in `src/engine/engine_derivative.c` is the input the analytic actuator derivatives (`actuatorLenDeriv`, `actuatorVelDeriv`) multiply by the gain for stateless actuators. It reads the control from the actuator's control block, `d->ctrl[m->actuator_ctrladr[i]]`, but then clamps it with `m->actuator_ctrllimited[i]` and `m->actuator_ctrlrange[2*i]`. Those two arrays are sized and indexed by control slot (`nu x 1` / `nu x 2`, per `mjmodel.h`), not by actuator, and the compiler writes them at `actuator_ctrladr[i] + k` in `user_model.cc`.

The two indices only coincide while every actuator has exactly one control. Put a three-input `pid` (or a four-input SO3 servo, or an input-less `dcmotor`) before a limited actuator and the derivative reads the limits of the wrong slot: with a clamped command the forward pass uses the clamped value while the derivative uses the raw one, so `qDeriv` (implicit/implicitfast) and the effective metric (discrete) both go wrong, and the resulting step depends on where the actuator sits in the `<actuator>` block. When the preceding actuator has zero controls the read is past the end of `actuator_ctrllimited`, which is the `use-after-poison` in the issue's ASan trace.

`mj_fwdActuation` already does this correctly (`clampVec(ctrl, m->actuator_ctrlrange, m->actuator_ctrllimited, nu, NULL)` walks control slots), as do the finite-difference path in `engine_derivative_fd.c` (loop over `nu`) and the visualizer (`actuator_ctrllimited[m->actuator_ctrladr[i]]`). This was the only remaining per-actuator use.

## The fix

Look up `actuator_ctrladr[i]` once and use it for both the control read and the limit lookup. One function, three lines.

## Testing

Added `DerivativeTest.ClampedCtrlDerivativeMultiInputOrder`, modelled on the existing `ActuatorOrder` test: the same two actuators — a zero-gain three-input `pid` and a limited affine-gain `general` actuator — loaded in both orders, driven at `ctrl=100` against `ctrlrange="0 0.5"`, stepped ten times under `implicitfast` and `discrete`. The test asserts the force, `qvel` and `qpos` of the limited joint are identical between the orderings.

Before the fix (run from `test/`):

```
engine_derivative_test.cc:590: Failure
  d1->qvel[1]  Which is: -0.77061507012014308
  d2->qvel[1]  Which is: -0.96247911068237624
[  FAILED  ] DerivativeTest.ClampedCtrlDerivativeMultiInputOrder
```

(The implicitfast arm fails the same way: `-0.7788` vs `-0.9625`. The pid-first model is the one that lands on the wrong slot — `actuator_ctrllimited[1]` is `0` there, so the derivative used the unclamped `100`.)

After the fix the values agree to the last bit in both arms, and

```
engine_derivative_test   37 tests   PASSED   (double)
engine_forward_test     142 tests   PASSED   (double)
```

and the full double-precision suite via `ctest`: **2208/2208 passed** (3 skipped by the harness). Both test binaries rebuilt with `-DmjUSESINGLE`: `engine_derivative_test` 37/37, `engine_forward_test` 141 passed / 1 skipped (`NormalizeQuats`, skipped on single precision by design). `clang-format` (the `.pre-commit-config.yaml` pin, 23.1.0) leaves the touched lines unchanged.

Changelog entry added under "Bug fixes".


yuvaltassa:
nice catch. can you do the CLA thing please?

Mohit-Ak:
Signed. Thanks.

